PR #108, opened 2015-03-31
Add syslog support, fix unaligned access

perexg:


catalinii:
Hi,

I will add this .... Thanks for the patch.

perexg:
Thanks. BTW: The HTTP mpegts stream contains extra 0x0d 0x0a bytes before TS data (before first 0x47 sync)..

catalinii:
so you mean there are actually 0d 0a 0d 0a between the last char and the 0x47 ?

Btw, I have merged the pull request in the new commit, but will reject it from here (my version needed some changes to be able to apply this as it is...

perexg:
I see that my code is in commit 1557fb01cef8954406ebaad4f6eb653974826010. It's good habit to keep credits to authors in the open source.

catalinii:
I appologize for that, my intention was not take the credits for your effort, believe me I appreciate it.
I did it because I was too far ahead with the changes for the new commit.

perexg:
Merged to the master, closing.

PR #118, opened 2015-04-13
RTCP: Fix the SES1 block length

perexg:
.. according to spec

### DIFF
--- a/stream.c
+++ b/stream.c
@@ -613,7 +613,7 @@ send_rtcp (int s_id, int ctime)
 	rtcp[51] = 0;
 	rtcp[52] = 0x80;
 	rtcp[53] = 0xCC;
-	copy16( rtcp, 54, (la + 16) / 4);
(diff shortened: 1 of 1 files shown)


catalinii:
Hi,

the specifications are saying:
 length: 16 bits
      The length of this RTCP packet in 32-bit words minus one,
      including the header and any padding.  (The offset of one makes
      zero a valid length and avoids a possible infinite loop in
      scanning a compound RTCP packet, while counting 32-bit words
      avoids a validity check for a multiple of 4.)

relevant lines:
len = la + 16;
if (len % 4 > 0)len = len - (len % 4) +4;  
copy16( rtcp, 54, (la + 16) / 4);

so len  / 4 -1 == (la + 16) / 4  , right ?

When I tested this code, wireshark was happy about the lengths, did u see some issues with this ?

perexg:
TVHeadend is not happy when (la % 4) == 0, the - 1 operation is ommited, so the block length is greater by 4 than the packet size! My change makes all this clear (padding correlation and - 1).

PR #119, opened 2015-04-13
increase kernel UDP buffer size for RTP streaming

perexg:


catalinii:
As the udp sockets respects generally udp_wmem sysctl setting, I was thinking of trying SO_SNDBUFFORCE  first and the fall back to SO_SNDBUF if that fails.

Also make it a configurable variable (opts.udp_send_buffer).... what do you think ?

perexg:
Yes, it might be configurable, I'd like just point to this. The sysctl wmem is only the upper limit which is not the default (I have the default value about 100kB but I can increase the buffer to 1MB or so in my system). SO_SNDBUFFORCE requires CAP_NET_ADMIN so it should be called only when the user has these rights (root).

catalinii:
Yep, this is why I wanted to fall back to SO_SNDBUF if SO_SNDBUFFORCE  fails.
I think multiple systems will not be able to accomodate this with the default settings and on top of that we need to keep in mind the systems with poor resources...

PR #168, opened 2015-09-14
Just some unixish whitspace changes

lucianm:
Hi Catalin,

here are two commits removing the executable bit on source files and converting few of the source files EOL from DOS style to unix style.

Best regards,
Lucian


catalinii:
Thanks for the pull request... by the time I saw it I added multiple changes to the files so I just used dos2unix and configured my editor for UNIX EOLs

lucianm:
That's perfect :+1: 
Just please apply only the commit which removes the executable bit on all the source files, or remove it yourself as well. Could it actually be that you're editing them from a Windows PC over a samba network share? If the editor can take care of EOL like you said you configured it, the executable bit should be fixable by some options in the samba export, I guess.

catalinii:
Hi,

I have cherry picked e37ab24.

Indeed I am using samba and eclipse (when I started using it, EOL changed), but I guess once the permissions are set right, eclipse won't change them.

Thanks for the fixes...

PR #207, opened 2015-12-16
Compatibility fixes

jmendeth:
minisatip currently fails to build for most embedded systems. Even when it builds, it fails to operate for DVBAPI 5.0, 5.1 and 5.2. This pull request addresses [most of] the compatibility problems.

I've tried my best to keep them as small and unobtrusive as possible, and made sure they don't change anything if the code is built for modern systems.

 - Allow the user to supply extra `CFLAGS`, and replace the default `-ggdb -fPIC`.
 - Option to disable use of `execinfo.h`, which isn't present on most embedded systems.
 - Wrap use of recent DVB flags in appropiate `#if` / `#endif`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@ DVBCSA?=yes
 DVBCA?=no
 SATIPCLIENT?=yes
 
-CFLAGS=$(NODVBCSA) -ggdb -fPIC
-LDFLAGS=-lpthread -lrt
+CFLAGS?=$(NODVBCSA) -ggdb -fPIC
+LDFLAGS?=-lpthread -lrt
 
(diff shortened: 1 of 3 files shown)


catalinii:
Thanks for the commit! 
I think it makes sense and is a good idea to have a flag for embedded devices.

BR

jmendeth:
Awesome! I'm making a package for minisatip on [Buildroot](https://buildroot.org), but I need to wait until a stable release. Is there any ETA for 0.6?

catalinii:
Hi,

Probably 0.4 is better than 0.6 at this point. The only improvement that 0.6 has is the multithreading support.

In the last period I did not had too much time to add test more but if you see any issues please feel free to open a new github issue.

Thanks

Issue #205, opened 2015-12-09
RFE: Option for disable SSDP messages in the LOG


lars18th:
Hi,

I suggest to add one option in the command line to silence all SSDP related messages from the log. When I try to debug the connection with a client, several SSDP message are interlaced in the log. These messages are very interesting when you like to debug the entire process, but when you like to see only connection traces, in this case the SSDP are garbagge.

When you work with SAT>IP protocol two different parts exist: DISCOVERY (linked to SSDP) and TRANSPORT (linked to RSTP). So, in order to debug one part, will be useful to 'silence' the other part.

Waht you think?

lars18th:
Hi,

I commented these LOG lines to silence some SSDP and headers in *minisatip.c* :

```
540: 	LOG("read RTSP (from handle %d sock_id %d, len: %d, sid %d):\n%s", s->sock,
966: 	LOG("Received SSDP packet from %s:%d -> handle %d",
1013: 	LOG("ssdp_reply fd: %d -> %s:%d, bootid: %d deviceid: %d http: %s", ssdp,
1214:	LOG("reply -> %d (%s:%d) CL:%d [sock_id %d]:\n%s", s->sock,
```

I suggest to move these lines to another log level. I feel they are too verbose until level 3 or 4.
You agree?

lars18th:
Hi,

I like to suggest also to incorporate another cosmetic change in logs:

In minisatip.c move the print of TCP messages to Log level 3 or 4. Specifcally these lines:

```
552:	LOG("read RTSP (from handle %d sock_id %d, len: %d, sid %d):\n%s", s->sock,
553:			s->id, s->rlen, s->sid, s->buf);

797: 	LOG("read HTTP from %d sid: %d: %s", s->sock, s->sid, s->buf);
```

As you can show the "s->buf" prints several lines. I suggest to print them only in superior log levels (typically I see the logs with level 2). At time I'm commenting out them!

You agree?

lars18th:
Hi @catalinii ,

Simple change for better readability (more easy to isolate relevant info):

Please, at "minisatip.c" replace all "s->buf" logs with "LOGL(3, "%s", s->buf);"

You have already done on some parts, like:
http://github.com/catalinii/minisatip/blob/a5c6903886fc1cdeb08cf5e17ca54a57d7448594/minisatip.c#L993

But not in messages like:
[...]

PR #226, opened 2016-01-11
Logs & Headers cleanups

lars18th:
Hi @catalinii ,

I'm doing some of the changes suggested in #205 .

Also, I try to enable compatibility with systems without Linux-DVB support (in the kernel). As the satip-client mode is starting to work, then I like to enable support for systems without DVB drivers. Now, I'm working on compile without native DVB headers.

Please, can you commit my changes?
Thank you!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,12 @@ DVBCA?=no
 SATIPCLIENT?=yes
 NETCVCLIENT?=no
 STATIC?=no
+LINUXDVB?=yes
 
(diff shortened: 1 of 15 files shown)


catalinii:
I did not accept Option for mute SSDP logs and other cleanups as that is getting more like a general problem... muting a specific component when you are not interested in the logs for it.

And this is not going to get resolved every time by adding a switch.

One of the option would be:

level 1 - http, rtsp
level 2 - dvbapi
level 4 - ssdp
....

Issue #239, opened 2016-02-01
DVBS2 and QPSK not working


REELcoder:
With the following request:

rtsp://10.76.1.3/?freq=12032&src=1&sr=27500&pol=h&fec=910&mtype=qpsk&msys=dvbs2

i get tp->mtype = 9 (i.e. 8psk)

```
[01/02 23:44:01.6306 main]: read RTSP (from handle 22 sock_id 12, len: 0, sid -1):
[01/02 23:44:01.6306 main]: SETUP rtsp://10.76.1.3/?freq=12032&src=1&sr=27500&pol=h&fec=910&mtype=qpsk&msys=dvbs2 RTSP/1.0
CSeq: 2
Transport: RTP/AVP;unicast;client_port=58875-58876
User-Agent: vdr-satip/2.2.3 (device 0)
If-Modified-Since: Thu, 01 Jan 1970 00:00:00 GMT


[01/02 23:44:01.6306 main]: detect_dvb_parameters (S)-> freq=12032&src=1&sr=27500&pol=h&fec=910&mtype=qpsk&msys=dvbs2
[01/02 23:44:01.6306 main]: detect_dvb_parameters (E) -> src=1, fe=0, freq=12032000, fec=11, sr=27500000, pol=2, ro=3, msys=6, mtype=0, plts=2, bw=8000000, inv=2, pids=NULL - apids=NULL - dpids=NULL x_pmt=NULL
[01/02 23:44:01.6306 main]: Setup stream -1 parameters, sock_id 12, handle 22
[01/02 23:44:01.6306 main]: stream.c:201 get_sid returns NULL for s_id = -1
[01/02 23:44:01.6306 main]: Mutex init 0x10036148
[01/02 23:44:01.6306 main]: utils.c:1161 allocation_wrapper malloc allocated 3008 bytes at 0x1041ec50
[01/02 23:44:01.6306 main]: Mutex init 0x1041ec54
[01/02 23:44:01.6306 main]: set_sock_lock: sock_id 12 locks also mutex 0x1041ec54
[01/02 23:44:01.6306 main]: utils.c:1161 allocation_wrapper malloc allocated 144 bytes at 0x1041f818
[01/02 23:44:01.6306 main]: Mutex init 0x1041f81c
[01/02 23:44:01.6306 main]: sockets_add: handle -2 (type 0) returning socket index 13 [0.0.0.0:0] read: 0x100060e0
[...]

PR #240, opened 2016-02-03
Fix for #239

REELcoder:
Hi catalini,

with the change in dvb.c issue #239 is fixed. So far I didn't found any side effects. QPSK and DVB-S2 is rare but Sky has one transponder using this together with FEC 9/10.

Cheers


catalinii:
Hi,

I will add some code to really fix this and keep the original behaviour in place (setting mtype to a proper value if not specified).

PR #251, opened 2016-02-16
fix: add NULL terminator to satipc_sym[]

perexg:


perexg:
I added second commit which uses offsetof() macro and fixes satipc_sym (wrong array).

PR #269, opened 2016-02-29
getTick() - use monotonic clocks

perexg:
I had big issues when NTP updated the clocks after the minisatip was started on an embedded system (but it will apply to all systems where the clocks are changed in a big step). The monotonic clock should be used for internal timing.

### DIFF
--- a/stream.c
+++ b/stream.c
@@ -53,7 +53,7 @@ unsigned init_tick, theTick;
 uint32_t getTick()
 {								 //ms
 	struct timespec ts;
-	clock_gettime(CLOCK_REALTIME, &ts);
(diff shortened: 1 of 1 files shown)


9000h:
RFC 3550, section 5.1 
specifically says that the RTP timestamp "MUST be derived from a clock that 
increased monotonically and linearly in time".

perexg:
I don't think that SAT>IP clients will check the RTP timestamps - there are another timing sources in the MPEG-TS packets. But yes, it's another reason to use the monotonic clocks as the clock source. The timeout socket in the signal thread won't be active when you change time from 1970 to 2016 for example when REALTIME clocks are used. Basically, all time comparisons will hit this issue.

PR #293, opened 2016-03-30
Mutex fixes

perexg:
It's not a good idea to call pthread_mutex_destroy() when mutex can be
used in another thread (deadlocks, crashes).

### DIFF
--- a/adapter.c
+++ b/adapter.c
@@ -58,6 +58,8 @@ adapter *adapter_alloc()
 {
 	adapter *ad = malloc1(sizeof(adapter));
 
+	mutex_init(&ad->mutex);
+
 	/* diseqc setup */
 	ad->diseqc_param.fast = opts.diseqc_fast;
 	ad->diseqc_param.committed_no = opts.diseqc_committed_no;
@@ -171,7 +173,6 @@ int init_hw(int i)
 		return 2;
 
 	ad = a[i];
-	mutex_init(&ad->mutex);
 	mutex_lock(&ad->mutex);
 	if (is_adapter_disabled(i))
 		goto NOK;
@@ -329,7 +330,7 @@ void close_adapter(int na)
 	ad->old_diseqc = -1;
 	ad->old_hiband = -1;
 	ad->old_pol = -1;
-	mutex_destroy(&ad->mutex);
+	mutex_unlock(&ad->mutex);
 	//      if(a[na]->buf)free1(a[na]->buf);a[na]->buf=NULL;
 	LOG("done closing adapter %d", na);
 }
@@ -1092,9 +1093,11 @@ void free_all_adapters()
 		{
 			if (a[i]->buf)
 				free1(a[i]->buf);
+			mutex_destroy(&a[i]->mutex);
 			free(a[i]);
 			a[i] = NULL;
 		}
+	mutex_destroy(&a_mutex);
 
 #ifndef DISABLE_NETCVCLIENT
 	fprintf(stderr, "\n\nREEL: recv_exit\n");
--- a/dvbapi.c
+++ b/dvbapi.c
@@ -975,7 +975,6 @@ int keys_del(int i)
 	enabledKeys = ek;
 	if (!ek && sock > 0)
 		TEST_WRITE(write(sock, buf, sizeof(buf)));
-	mutex_destroy(&k->mutex);
 	return 0;
 }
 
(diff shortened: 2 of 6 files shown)


catalinii:
Hi,

I am not sure about the changes, as I know that sockets_lock are deleting the mutexes associated with a sock when it is disabled (with this change it could still remain attached to it and 2 different sockets locking the same stream after the initial stream was closed).

The other thing is that when I destroy the mutex before unlocking it I set enabled to false which stops another threads from locking it between the unlock and destroy.

There is a small corner case when another thread is already in pthread_mutex_lock between the unlock and destroy and the call could succeed, and pthread_mutex_destroy to fail, but I will have the fix for that soon (basically in mutex_lock, if after pthread_mutex_lock the mutex is disabled, unlock it and exit)


The other choice is to not call  pthread_mutex_destroy, but as you can see in the current implementation, if that fails, the mutex is left enabled....

Let me know your thoughts....

Thanks

perexg:
Hi,

You cannot operate with mutex destroy call to finish lock/unlock sequence. Basically, lock/unlock calls should be balanced and the mutexes should be safely destroyed only when the mutex is not more used (touched) by any thread.

From man page:

`A  destroyed  mutex  object can be reinitialized using pthread_mutex_init(); the results of otherwise referencing the object after it has been destroyed are undefined.`

Imagine:

- thread 1 calls sockets_del - ss->mutex is locked
- thread 2 calls pthread_mutex_lock in the middle of destroy sequence on ss->mutex 
- thread 1 will destroy the mutex, but thread 2 is still trying to lock ss->mutex - undefined behaviour

The protection using enable flag is not sufficient and only hides some issues.

My patch tries to balance lock/unlock calls and destroy mutexes. But, yes, the locking should be reviewed more.

Imagine:

- thread 1 calls sockets_del -  ss->mutex is locked
- thread 2 calls pthread_mutex_lock on ss->mutex in same time
- thread 1 finishes sockets_del, but allocates new socket reusing same structure
- thread 2 will wakeup, but the structure contains data for newly allocated (reused) socket, but thread 2 thinks that the data are for the previous (deleted) socket

One way is to use only reference counter and this counter should be modified when the socket/mutex is in use  - for example in select_and_execute() -  pf[] init - increase refcount for sockets and after poll() - decrease refcount for sockets used in pf[]. The sockets structures with refcount > 0 should not be reused (sockets_add/add_new_lock).

BTW: Looking to the code, s[] used for pf[] initialization (poll) should be locked - 'for (i = 0; i < max_sock; i++)' loop.

BTW2: What about to avoid recursive mutexes ? It also hides the locking problems sometimes.

catalinii:
Hi,

I missed the point about the pthread_mutex_destroy. I will fix it (for now by removing the call from mutex_destroy)

For the scenario that you mention I will still get to have a structure that holds mutex info..

With enabled at least I can check if the object that references it is still active and choose to detach at the sockets from the object that is referenced in ss->lock...(which seems to be used currently just by the streams...)

It is true that I still can do it if I have a reference counter, but I need to mark in a way a mutex that is no longer enabled to avoid another threads locking mutexes that the objects are no longer referencing.

I understand that what I did is not very logical, but my point was to have as less mutex_lock syscalls as possible and to be bothered by mutex_locks as less as possible.

This is one of the reasons I have implemented clean_mutexes (which I agree it can be improved by using counters in the SMutex structure and unlock it if I left in locked).

The another choice would be to pass to mutex_lock struct struct_array instead of SMutex and to check if the object referencing the lock is active, but still I would be able to use it just with struct struct_array and loose the generality a bit.

RE BTW: I use s[x] in a thread only if s[x].tid == thread_id, so I am not using same records in different threads.

RE BTW2: I could but then I would have even more mutex_lock calls as I woudl not be able to double lock from the same thread.

The point was to make sure that at the end of a cycle in select_and_execute there is no lock was acquired and not released.

Over the weekend I have managed to reproduce some of the issues that we've mentioned before:
sockets_del unlocks and another thread locks before desotry. Destroy fails, and the other thread does not release the lock (because is disabled) which ends up in a thread being stuck.

By locking and checking if the mutex is destroy and then unlocking it seems the tests for the past 3 days did not reveal any lock.

perexg:
The reference counting can be really nicely done through the atomic operations - https://gcc.gnu.org/onlinedocs/gcc-4.4.3/gcc/Atomic-Builtins.html .

I looked to your 'fix' but it's only workaround. I think that the whole mutex usage should be reviewed - especially you mentioned the own optimizations - but don't try to do that. The pthread library has already very optimized mutexes which don't try to invoke syscalls until necessary (the mutex functions are using atomic operations and the syscall is really invoked on demand to wait for another thread when the atomic operation signals that the mutex is busy in other thread).

catalinii:
Thanks for the pointer... 

I will watch this, if there will be future issues I will re-implement the whole thing...

catalinii:
HI,

do you have an email / skype / hangouts ?

I would like to integrate your satip-axe code into minisatip if it's fine with you.

If you want to talk, send me an email on catalinii@yahoo.com.

Thanks

catalinii:
Thank you for inputs on this issue.
I believe the implementation is stable for now, even is not at it's best.

I will close this.

PR #305, opened 2016-05-15
fix jess frequency handling

perexg:
Obvious thinko? I've not tested this, but my user reported that Unicable II does not work.


catalinii:
Hi,

You mean unicable ii is not working also except jess?

Thanks

perexg:
Jess is Unicable II, isn't?

catalinii:
Yes it is, I just wanted to make sure we are talking just about the jess code...

Thanks

PR #350, opened 2017-04-28
Changes in the axe version of minisatip

perexg:
The RTP TCP data buffer change is obvious and it reduces the writev() kernel syscalls. I'm not sure if all clients are fine with it, but while RTP TCP data transfer mode is not in the specification, I think that the open source implementation will adopt. Also, users may decrease the defines in stream.h on demand.

EDIT: Added a big patch which sets the RTSP socket to the non-blocking mode when RTP/AVP/TCP mode is activated and fixed bunch of mistakes in the buffer code when the kernel is not able to queue more data for the TCP socket.

### DIFF
--- a/adapter.c
+++ b/adapter.c
@@ -1056,6 +1056,18 @@ get_adapter1(int aid, char *file, int line)
 	return a[aid];
 }
 
+adapter *
+get_configured_adapter1(int aid, char *file, int line)
+{
+	if (aid < 0 || aid >= MAX_ADAPTERS || !a[aid] || disabled[aid])
+	{
+		LOG("%s:%d: get_configured_adapter returns NULL for adapter_id %d",
+						file, line, aid);
+		return NULL;
+	}
+	return a[aid];
+}
+
 char* get_stream_pids(int s_id, char *dest, int max_size);
 char *
 describe_adapter(int sid, int aid, char *dad, int ld)
--- a/adapter.h
+++ b/adapter.h
@@ -139,6 +139,7 @@ int update_pids(int aid);
 int tune(int aid, int sid);
 SPid *find_pid(int aid, int p);
 adapter * get_adapter1(int aid, char *file, int line);
+adapter * get_configured_adapter1(int aid, char *file, int line);
 char *describe_adapter(int sid, int aid, char *dad, int ld);
 void dump_pids(int aid);
 void sort_pids(int aid);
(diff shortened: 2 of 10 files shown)


perexg:
@catalinii: BTW: I think that the flush interval in dmx_read should be configurable (100ms is a bit high value in my eyes).

catalinii:
Hi,

thanks for the PR.

It looks good to me but I have 2 question:
1) 
In https://github.com/catalinii/minisatip/pull/350/commits/a8e4ea0c55e8bce746ff1ef3022545be0044686f#diff-6901e593987d4353aeed5ba2e3b7fae7L696
1
You have removed in socketworks.c: 
ss->type &= ~TYPE_CONNECT;

Is there a reason for this or by mistake.

I am asking because this will break dvbapi connection which uses non-blocking connect.
2)

In the same commit I will remove in minisatip.c:
if (!s->nonblock) {
 +		send(s->sock, resp, strlen(resp), MSG_NOSIGNAL);
 +		if (binary)
 +			send(s->sock, desc, lr, MSG_NOSIGNAL);
 +	} else

In my case there was a partial write on that send (because of the small buffer), so I think is better to just remove that code path.

Thanks

perexg:
1) the ss->type is already set without these flags in sockets_add() and there are many checks in the code like ss->type == TYPE_HTTP, so I expected that the flags are valid only for sockets_add() not later; it means that the 

2) it won't work.. I'm currently testing code which turns on the nonblocking mode for all HTTP / RTSP sockets and I found that there's special ss->timeout_ms == 1 condition in my_writev() which completely breaks everything when you use writev() for standard HTTP requests..

catalinii:
1) Ok then, I will look into it.

2) I can remove:
	if (s->timeout_ms == 1)
		return -1;
from my_writev as it the problem was blocking for too much time, but even if timeout is 100ms (more or less) that code will be ok without it.

perexg:
The last commit https://github.com/catalinii/minisatip/pull/350/commits/e764a6d7dc4e35d7e517fc58643887c4f50d37ab enables the full non-blocking mode for all RTSP/HTTP requests. @catalinii : perhaps, you may consider to not abuse 'timeout_ms' with the 'close : timeout_ms == 1' special condition a create a new variable. The 'timeout_ms_wrwait' may be renamed, too.

perexg:
And the very last commit add -H (--threshold) option to have a chance to modify the packet (data) timing.

catalinii:
I will merge this one.

If you have more feel free to create new ones.

I will also try to create a force_close variable (rename timeout_ms_wrwait).

I might also change minisatip.c: read_http: sockets_timeout(s->id, 1) to something bigger and differentiate between timeout and force_close.

Thank you for the commits, I will test also and let you know if there is something.

catalinii:
Hi,

I have just moved the master to 0.7 now.

There is a 0.5 branch with 0.5 code.

Thanks

PR #351, opened 2017-05-02
obtain the state data through json

perexg:
Please, do not merge @catalinii ! It is a work in progress.

I just like to get ACK or NAK for this. The goal is to move the data interface to something more 'standardized' and it's the JSON data structure for javascript. Also, it's a bad idea to reload the whole page to obtain only data and the $varname$ lookups are really constly when you rewrite the html files (for a lot of variables like the status page is using).

The /status2.html path shows the new page using json, but I have more ideas (extensions and optimizations).


catalinii:
Hi,

I think it is a good idea. I was thinking for a while to do that as the initial model is not that successfull, but ideally I would like the ability of also changing stuff in minisatip. But that maybe later.

To setup a branch and add everything there ?

Thanks

catalinii:
Hi,

I have pushed this commit in 0.7-ca where I am working currently (rewriting some good portions of the CA implementation). I will test there and add new code if needed.

To compile (for now) - basically in the same way as for axe:
make DVBAPI=0 DVBCSA=0 DVBAES=0  DVBCA=0

Please add new commits about this (or another things) in 0.7-ca, unless is really urgent and needs to be in master.

I will close this, but feel free create a new one for 0.7 if u have anything.

Thanks

perexg:
Could you revert? I have more better version in my tree.

perexg:
I pushed all my latest changes to this branch. Just revert (or remove) the initial commit. Also the second (POST) patch should be applied to master (except the status.html change). Otherwise, the wrong HTTP requests hang.

perexg:
Last commit - added 'Stats' button to the status page:

a

perexg:
Rebased against the latest 'master' branch.

catalinii:
Hi,

do you want me to remove f5cde2c519ee7c and apply 816b34e instead ?

And then all other commits ?

Thanks

perexg:
Yep, exactly. Thanks.

catalinii:
Thanks, they are all in 0.7-ca now.

catalinii:
Thank you for submitting the code.

PR #354, opened 2017-05-04
Axe upstream

perexg:
Two patches for master - the complete axe code split and resolve all (at least for my gcc version) warnings when -Wall is used. Also, generate the proper dependencies in makefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /minisatip
 *.o
+*.d
 config.log
 config.status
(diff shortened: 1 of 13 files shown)


catalinii:
Thanks, I've included this in 0.7-ca.

The way it seem now, I might move the Master to 0.7-ca when is ready.

catalinii:
Thank you for submitting the code.

PR #359, opened 2017-05-12
implement diseqc-multi

perexg:
@catalinii : I'm not sure, if you like to merge this one. But it helped on one equipment and it's straight (debugged remotely).

This is really a workaround. I've seen a diseqc hardware which is
not able to switch properly to another input unless the position 0
is activated before any other input.


catalinii:
Thanks, I've merged this. I don't think it is an issue if there is an command line argument enabling it.

BTW, what hardware did u had issues with ? (or the link to the issue).

Thanks

PR #361, opened 2017-05-25
axe: axe is __sh__ platform but it can do 32 pids

perexg:
Uff. It took me 2 hours to find where's the issue (weird no PID data bugs).. Perhaps it should be more visible that this limit is applied. It may really depend on hardware. It should be configurable in my opinion.


catalinii:
Yeah my amiko supports also 32, but did not wanted to get close to it so I've left it at 20.
Maybe I should go higher with it.

Thanks for the PR

PR #366, opened 2017-06-25
adapter and stream info in json format at http://[ip:port]/status.json

squeezeday:
I would like to get info from minisatip into other apps. This is my contribution for getting adapter and stream info in json. Please note that I don't usually code C so some code review may be in order.

### DIFF
--- a/minisatip.c
+++ b/minisatip.c
@@ -51,6 +51,7 @@
 struct struct_opts opts;
 
 #define DESC_XML "desc.xml"
+#define STATUS_JSON "status.json"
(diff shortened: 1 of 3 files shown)


catalinii:
Hi,

Thank you for the commit.
In 0.7-ca there is already implemented this function, as the web page will be based entirely on json.

For example you can get the state by doing: wget -O x.json http://192.168.101.112:8080/state.json

The content is generated based on https://github.com/catalinii/minisatip/blob/0.7-ca/utils.c#L701 .

Please let me know if there is something that you need and is not added.

PR #408, opened 2018-01-11
Fix for DVBLink & PlexDVR to allow using minisatip as SAT->IP Source

Jondalar:
From DVB Logic:

"although sat2ip specification does not talk explicitly about 
separators, it uses comma everywhere in examples, e.g. as on page 69 pf 
the sat2ip 1.2.2 version of the spec:
s=SatIPServer:1 <sat frontends>,<terr frontends>,<cable frontends>"

This does enable DVBLink and PlexDVR compatibility.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -84,7 +84,7 @@ char *describe_streams(sockets *s, char *req, char *sbuf, int size)
 	//	else LOG_AND_RETURN(NULL, "No session associated with sock_id %d", s->sock_id);
 
 	snprintf(sbuf, size - 1,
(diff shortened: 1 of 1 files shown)


9000h:
which server is working that way as it's in contradiction to the spec

9000h:
but it could be there is something we have overlocked
if there is not only dvbs then it may should look like "s=SatIPServer:1 %d,%d,%d\r\n"
https://github.com/DigitalDevices/octonet/blob/dfda1b722bfa99d726e04be86fb3f731f240362d/octoserve/octoserve.c#L866

catalinii:
Merged thank you

PR #445, opened 2018-06-19
Improve handling of opts.drop_encrypted

lars18th:
Hi,

Using the paremeter "-E" in the command line enables (in fact disables) the config option `opts.drop_encrypted`. When using this mode the unencrypted packets are passed without modification to the client. This is good to preserver the PCR. However, when sending the encrypted stream at the start the client can think that the channels are not in clear. So, this patch blocks the stream (like without using the "-E") until the first encrypted packets are descrambled. And only after this point, if some packet can't be descrambled then it is passed transparently.


catalinii:
Hi,

just to understand right, the purpose of the patch is to block the encrypted packets until the first CW, when the encrypted packets are not supposed to be dropped (-E).

If that is the case, you could have multiple streams using the same adapter. In that case you will drop the packets just for the first stream and not for all.

lars18th:
Hi @catalinii ,

I don't have sufficient time to move this patch from the "adapter" level to the "stream" level.
So Please, can you complete it?

Thank you!

lars18th:
Hi @catalinii ,

No more comments about this patch?

catalinii:
I think implementing such patch would lower the performance of minisatip especially on slow cpus. 
This would happen if you move some of the processing from adapter level (once for all streams) to stream level where you need to do it for each stream.

lars18th:
Hi @catalinii ,

Thank you for your comment!

First of all, I don't use a slow CPU so the performance isn't a problem to me. :wink: 

If you are concerned about the performance then I suggest:

- Make the patch "transparent", that's if it's not enabled by a command line parameter, then don't add futher checks. You agree?
- The "if" sentence in `pmt_decrypt_stream()` I feel doesn't have any impact in the performance. I'm right?
- The "if" sentence in `pmt_process_stream()` perhaps needs to be optimized. I'll try to do it.
- And regarding to move the code to the stream level... I don't know how to do it. Please, can you help me?

I'll present an upgraded version soon.

lars18th:
Hi @catalinii ,

Another comment/explain on the purpose of this patch:

- When using a transparent decrypting you can be in trouble with _some_ clients.
- The user case is when the SAT>IP client requests a service and it receives "encrypted" content. Then in this case, some of them refuse to play it.
- The solution it's then to block the stream until you can decrypt it. And that's OK.
- _However_ when you're playing a decrypted stream and **at some point** the decryption fails, for example 100ms of delay in getting the keys, then you're blocking the stream again.
- And in this specific case the client doesn't receive **any** stream. This generates troubles when not receiving the PCR's timestamps, because they are inserted in the encrypted stream with the PCR.
- So the solution is then to "pass" all traffic, even when it' not decoded. And this is not a problem as the behaviour of SAT>IP client is normally to overpass the error: it think the stream has some error, or the data is partially encrypted. But all the time the play continues, and it will be correctly decoded just when new decrypted packets are comming.

So, please help me to apply a workaround to this specific case.
Thank you!

lars18th:
Hi @catalinii ,

Some recent changes make this PR with conflicts.
Please, can you improve it?

The objective is quite simple: block all encrypted packets **until** the _first_ decoding is done; and after that **pass** encrypted packets if the decoding fail.

Can you help on this?

catalinii:
Hi Lars,

Can I ask which client will behave better by doing this ?
And as I suppose you already had the patch working, did that client worked better?

Thanks

lars18th:
> Can I ask which client will behave better by doing this ?

I have two clients that _requires_ this patch. One is a TV, and one is software. Sorry, I can't disclose the names.

> And as I suppose you already had the patch working, did that client worked better?

Yes. I apply the patch to all of my minisatip binaries. Without it the clients _refuse_ to play any encrypted service. The patch works as:

- When the client requests a new service, it doesn't receives the stream until it is decrypted. So it can't detect that isn't in clear. Futhermore, as this happens only at the beginning, the player interprets this as a long-time tunning. So no problem at all.
- And after the first decryption, any error in the decrypt process, can be passed to the client without problems. It interprets the uncrypted data as simple "errors". But the PCR in the packets are required to maintain the reproduction of the stream without troubles.

So, the patch is a must have when using the `-E` parameter.
Regards.

catalinii:
So what you are saying that you need the PCR packets throughout the entire stream. Because this issue is very different from dealing with encrypted/decrypted packets.

Would keep sending PCR packets unaltered work for you?

lars18th:
> So what you are saying that you need the PCR packets throughout the entire stream. Because this issue is very different from dealing with encrypted/decrypted packets.

Sorry?
The question is this (and not other): When minisatip can't decrypt a packet (because it doesn't have the correct key _at this moment_), then what it needs to do? Well, two options are posible: block the packet or pass the packet (unaltered). As each option has benefits and drawbacks, then it's a good idea to leave to the user the option to use one or other (with the parameter `-E`). However, the "pass option" has troubles with the "start of the stream" when the key is not yet available. The trouble is with clients that expect a "clear" channel instead of an "encrypted" one. If these clients detect (even just 1 packet _at start_) then refuse to play the service. So this patch targets this _special_ case.

> Would keep sending PCR packets unaltered work for you?

Obviously! It works perfectly. The `-E` parameter improves a lot the compatibility with the clients. In fact, is the same that comes with hardware decoders. In a STB or TV when the encrypted source stream has some bit errors, the decoding of the packet can fail, but the decoder is still getting the packet. This assures that PCR packets are received.

I repeat: I use `-E` all the time. However, I need to block encrypted packets _only_ at start.
This patch does just that. And I recommend it.

Regards.

catalinii:
Hi Lars,

My understanding is that -E block any encrypted packets: At the start or after start when there isbno key. Without it clients like tvh will stop playing an encrypted channel.

If we block just at start this will affect other clients as well which will stop playing the channel if an encrypted packets get in.

My understanding is that your use case will work if we do not block the PCR packets while using the current implementation of -E. Is that correct?

9000h:
not sure if this is helpful in regards to PCR
https://github.com/videolan/dvblast/commit/f4b1a8e15f0514d80f729c876c6a960639ecf9b9

lars18th:
Hi @catalinii ,

I updated the code. Please review it!

Comments:

- It only changes the behaviour when `-E -t` are used at the same time.
- When enabled, it drops _only_ encrypted packets at start of playing.
- It works at PMT (service) level, so it handles multiple channels from the same adapter.

I hope you agree with it now.
Regards.

lars18th:
And as a bonus update to use `PID_FROM_TS(b)` in all cases.

lars18th:
Hi @catalinii ,

I fixed one error. From my point of view the code of the patch is finished.
So. you can test it and merge when you like.

Regards.

lars18th:
Hi @catalinii ,

Please try this PR!

Let me to show the bahviour with the _bitrate_ output:

* without params (-E -t):
```
             #   ##        #    ##    ##
          ## ## #### #    #########  ####
          ##### #### ###  ######### #####
          ########## ###  ################
          ########## ###  ################
          ########## #### ################
    . . . ########## #### ################
```

* with "-t" only:
```
             #   ##        #    ##    ##
          ## ## #### #    #########  ####
          ##### #### ###  ######### #####
          ########## ###  ################
          ########## ###  ################
          ########## #### ################
    . . . ########## #### ################
```

* with "-E" only:
```
             #   ##        #    ##    ##
      **  ## ## ####*#   *#########  ####
    ******##### ####*###**######### #####
    ******##########*###**################
    ******##########*###**################
    ******##########*####*################
    ******##########*####*################
```

* with "-E" _and_ "-t":
```
             #   ##        #    ##    ##
          ## ## ####*#   *#########  ####
          ##### ####*###**######### #####
          ##########*###**################
          ##########*###**################
          ##########*####*################
    . . . ##########*####*################
```

Without this PR the `-E -t` is equal to `-E`. And as I commented before this change only changes the behaviour when using `-E` _and_ `-t` params. So it can't disturb to users that doesn't use it.

Please, consider to merge it!

catalinii:
Hi @lars18th,

this patch does not work in the following situation:

1) stream 1 is already playing channel A.
2) channel A stops getting a key
3) start stream 2 on channel A

Now the behaviour that you want is for stream 1 to send encrypted packets, while stream 2 to not send encrypted packets.

Because of this scenario, you cannot implement this in the adapter buffer, but to find a way to do filtering on the stream. Also if your issues is with pid 8191, even if you filter it at the stream level it won't work (because the packets that cannot be decrypted are marked as pid 8191).

That's why I suggested dealing correctly with pid 8191.

catalinii:
Basicallly what you need to do is:
1) remove that code that changes the pid from X to 8191 (if the packet should be dropped).
2) have a drop_encrypted per stream
3) if the drop_encrypted is enabled, filter the packet here: https://github.com/lars18th/minisatip/blob/master/src/stream.c#L937
4) if drop_encrypted is enabled, you have a CI enabled, force the packet to be sent to process_packet

lars18th:
I withdraw this PR too as this #708 fixes partially the same thing.

PR #455, opened 2018-09-21
adapter: handle ad->active_pids correctly, fixes VDR issue

perexg:
There's a race for the decrementing ad->active_pids. I saw a problem when TEARDOWN / PLAY operations were processed too quickly (VDR) and mark_pid_add() code changed flags from 3 (delete) to 2 (add) for the changed pid. In this case, the update_pids() incremented ad->active_pids again, which is wrong, because it must be tied to the hw config (ad->set_pid call). This change fixes that.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -878,7 +878,7 @@ int update_pids(int aid)
 			if (dp)
 				dump_pids(aid);
 			dp = 0;
-			if (ad->pids[i].fd <= 0)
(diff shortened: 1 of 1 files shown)


perexg:
The code was tested (see the reference above).

catalinii:
Hi

I wanted to ask if it was tested, if it was I will merge it

Thank you for the PR

Issue #443, opened 2018-06-10
DVB-S2 unicable tuning sequence


auldone:
I have recently bought a Digibit R1 to replace my DD Cine card I am running with TVHeadend, so I setup network access and configured the tuners in TVHeadend.

For testing I just disabled the tuner card and enabled the satip server running on stock firmware and was not able to tune to any of the existing channels. I have tried the stock firmware that came with the box, the latest one from Inverto and several versions of minisatip without success.

I am running a unicable setup with SCR from Kathrein (EXR 1581, EXR 2582) setup for ASTRA 19,2E.

Running a network scan with the stock firmware I noticed that there was always found the same set of channels, several times. TVHeadend was complaning about changed transponder IDs, so there seemed to be an issue with unicable tuning, since there seemed to be no change in the tuned transponder after sending the tune sequence.

So I had a look at the source code of other unicable tuning implementations of VDR and TVHeadend which was working before. I noticed that there was a difference in the tuning sequence used:

VDR uses sequence 0xE0 0x10 0x5A D1 D2
http://www.vdr-wiki.de/wiki/index.php/SCR_Single_Cable_Distribution#ODU_Channel_change_.28mandatory.29

TVheadend uses sequence 0xE0 0x10 0x5A D1 D2
https://github.com/tvheadend/tvheadend/blob/master/src/input/mpegts/linuxdvb/linuxdvb_en50494.c
Comment says:
/* addresses 0x00, 0x10 and 0x11 are possible */
#define LINUXDVB_EN50494_ADDRESS 0x10

So I thought, that might be my problem as minisatip has: 

	struct dvb_diseqc_master_cmd cmd =
		{
{0xe0, 0x11, 0x5a, 0x00, 0x00}, 5};

For a try I replaced the 0x11 with 0x10 in minisatip sources and recompiled the binary and put it on the satip-axe box which made the difference and everything started working as expected.

Currently running minisatip with:
MINISATIP8_OPTS="-9 0 -7 0:1,0:2,0:3 -u 0:4-1748,1:5-1864,2:6-1980,3:7-2096"

As this took me quite some time to figure out, I wanted to report this issue, as someone else might fall over this. Before I had read every unicable/diseqc related information regarding Digibit, minisatip and the axe firmware, tried fiddling with the unicable timings and diseqc related stuff without success.

Would it be possible to change that bit in the sources (which might break existing working setups) or make that configurable by parameter?

EDIT: did a bit more research on the address bit. 
[...]

PR #459, opened 2018-09-25
allow to change the diseqc address through -d option, issue #443

perexg:


catalinii:
@perexg I was wondering if you were able to test this?

perexg:
It's in my queue (testing) for the standard diseqc switches. I cannot test unicable (no hardware here). But I reviewed the code several times - it should not change the default behaviour (except the default unicable device address which is used in another software like tvheadend).

catalinii:
Do you want me to merge it now (low risk) or after testing?

PR #464, opened 2018-10-12
add support for DVB-S2 multistream

perexg:
It was tested in my environment (with satip-axe).


perexg:
Ok, I studied this more (PLS) and here are the improvements. The plsm= argument (PLS mode) is a bit extra, because the ROOT/GOLD PLS codes can be calculated from each other (see pls_scrambling_index() fcn for linux's dvbapi), but for the consistency and the future extensions, I would keep it. The COMBO PLS mode is probably mistake but STV0900+ demodulator has _three_ bits for the PLS mode, so we may find something later.

perexg:
Nice PLS explanation: https://github.com/DigitalDevices/dddvb/blob/master/apps/pls.c

PR #466, opened 2018-10-17
Signal code improvements

perexg:
The signal/snr is reset to prevent to use old values after tuning. Also, improve the signal_thread() function and add fast_status for the axe hardware. The new code was tested on the axe hardware.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -1156,7 +1156,7 @@ int mark_pids_add(int sid, int aid, char *pids)
 	for (i = 0; i < la; i++)
 	{
 		pid = map_intd(arg[i], NULL, -1);
-		if (pid == -1)
+		if (pid < 0 || pid > 8192)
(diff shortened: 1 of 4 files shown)


src/adapter.c
```diff
@@ -2057,24 +2063,28 @@ int delsys_match(adapter *ad, int del_sys)
 
 int signal_thread(sockets *s)
 {
-	int i;
+	int i, status;
 	int64_t ts, ctime;
 	adapter *ad;
-	for (i = 0; i < MAX_ADAPTERS; i++)
-		if ((ad = get_adapter_nw(i)) && ad->get_signal && (ad->fe > 0) && ad->tp.freq && (ad->status_cnt++ > 0) // make sure the kernel has updated the status
```
9000h: where is the (ad->status_cnt++ > 0) gone to?

PR #477, opened 2018-11-08
fix AuroaTV App

9000h:


9000h:
fixes also "TVMosaic Live!"

PR #478, opened 2018-11-13
Signal fix on first tune

9000h:
issue was here on a Panasonic TV


9000h:
I try to solve an issue on a Panasonic TV where after the second manual scan the signal is missing, no rtcp packets from minisatip
please ignore this PR it did not solve it

PR #480, opened 2018-11-15
OPTIONS compare fix

9000h:


9000h:
I'm bit surprised how this could happen

PR #486, opened 2018-11-17
rv uninitialised

9000h:


9000h:
ok but then rv should have a defined value as the ifdef before may disable the fist call

Issue #491, opened 2018-11-18
diseqc issue ?


9000h:
find this one after some switching between dvb-s, dvb-c,dvb-t2
```
[18/11 18:23:59.898 main]: copy_dvb_param start -> src=0, fe=0, freq=466000, fec=9, sr=6900000, pol=0, ro=3, msys=1, mtype=5, plts=2, bw=8000000, inv=2, pids=NULL, apids=101, dpids=116 x_pmt=NULL
[18/11 18:23:59.898 main]: copy_dvb_parameters -> src=1, fe=0, freq=11302000, fec=2 sr=22000000, pol=2, ro=0, msys=6, mtype=9, plts=2, bw=8000000, inv=2, pids=NULL, apids=NULL, dpids=NULL x_pmt=NULL
[18/11 18:23:59.898 main]: send_diseqc fd 19, pos = 0 (c 0 u 0), pol = 1, hiband = 0
[18/11 18:23:59.898 main]: send_diseqc: FE_SET_TONE failed for fd 19: Operation not supported
[18/11 18:23:59.898 main]: send_diseqc: FE_SET_VOLTAGE failed for fd 19: Operation not supported
[18/11 18:23:59.913 main]: send_diseqc: FE_DISEQC_SEND_MASTER_CMD committed failed for fd 19: Operation not supported
[18/11 18:23:59.998 main]: send_diseqc: FE_DISEQC_SEND_BURST failed for fd 19: Operation not supported
[18/11 18:24:00.013 main]: send_diseqc: FE_SET_TONE failed for fd 19: Operation not supported
[18/11 18:24:00.013 main]: tuning to 11302000(1552000) pol: h (2) sr:22000000 fec:23 delsys:dvbs2 mod:8psk rolloff:0.35 pilot: , ts clear=43847, ts pol=43847
[18/11 18:24:00.013 main]: dvb_tune: set property failed 22 Invalid argument
[18/11 18:24:00.013 main]: Adding pid 0 to the list of pids as not explicitly added for adapter 6
[18/11 18:24:00.013 main]: deleting PMT 0, master PMT 0, name 13th Street (VF)
```
http://paste.ubuntu.com/p/r92XFHHy5R/

9000h:
the logging in get_free_adapter seem to also not right as  is 0 all the time, look for sys: and the frequency  which is dvbc 

```
[...]

PR #497, opened 2018-11-19
Restart needed adapters fix

9000h:
in reference to https://github.com/catalinii/minisatip/issues/491
maybe need some additional tests
VDR seems to be happy no "404 Not Found" and  no "dvb_tune: set property failed 22 Invalid argument" ... anymore


9000h:
on shared access with 1 adapter the pid's got deleted :-( so it's probably still something missing

9000h:
probably the wrong way

PR #515, opened 2018-12-14
inconsistent logging for "tuning to"

9000h:


9000h:
it's not irritating when only dvbs tuning get logged to syslog when no logging option is given

PR #521, opened 2018-12-17
fix do_tune

9000h:


9000h:
ad->tp.freq = 0; did confuse me a bit, could it be also dead code?

9000h:
I could not find any negative impact here, but your comment suggest the code is only used on fist tune (a new stream).
if (ad->sid_cnt++ == 0) // tune always first time for a stream
		ad->tp.freq = 0;
If there is a stream with the same freq you invalidate the later compare which is probably not needed/expected and the log shows by the compare  the old freq=0 and all other params still valid, which is really confusing. 
Can you explain why you set the freq=0 here.

catalinii:
A grep 'if.*freq' shows couple of places where freq is compared with 0.
Minisatip evolved and while at some point I was using one way to test something that may have changed in a later version. This is why I was saying that in some places freq is tested against 0.
Overall I do think there is value in having one way of doing things, but my point was that the commit in the way it is right now, it has big chances of breaking existing functionality.

PR #526, opened 2019-01-07
remove unused sym

9000h:
ad_diseqc is not used anywhere, ad_pos has the diseqc


src/adapter.c
```diff
@@ -2082,7 +2082,7 @@ int delsys_match(adapter *ad, int del_sys)
 	return 0;
 }
 
-int signal_thread(sockets *s)
+int signal_thread(void)
```
catalinii: socket_action (to which signal_thread is casted, has a void * parameter, which is actually a sockets. Each time is called socketworks.c will pass that as argument, so this should not be removed.

PR #539, opened 2019-02-14
fix ddci detect

9000h:
if the vendor driver for DDD devices is installed then we need to check for /dev/dvb/adapter0/ci0 the kernel drivers use sec0


9000h:
@catalinii  you may have better way to detect if sec0 or ci0 has to be used

PR #541, opened 2019-02-15
extent max caid

9000h:
as there are cams out which can hold more than 20 caid's like the diablo cam or matrix cam air


9000h:
not sure if it would make sense to do this dynamically

catalinii:
I am not sure if there is a point in bothering with doing it dynamically.

PR #543, opened 2019-02-16
fix detect ddci again

9000h:


catalinii:
The patch has conflicts, can you please re-create it?

catalinii:
fixed manually

Issue #448, opened 2018-07-04
SAT>IP client mode: support for servers without "pids=all"


lars18th:
Hi,

I have different SAT>IP servers. Regarding the "pids=all" support I see three different behaviours:

1. It works! The server sends the full transport stream without problems.
2. It's accepted, but dosen't work. The server not filters the stream, but the bandwith is limited. Then you receive a lot of CC errors, but you can still read the PSI tables.
3. It isn't accepted. The server refuses the request or internally replaces it with "pids=none". So, you can't receive any packet.

In order to improve the support for servers in the third group I suggest to include support for "hacking" the "pids=all" calls. For example, you can provide this command line parameter: `-s --satip-servers [DELSYS:][FE_ID@][source_ip/]host[:port][#no-all]`.

This parameter should work at _server_ level (so you can mix different servers with different support); and it instructs to **replace** the request "pids=all" with "pids=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20" (that's all PSI tables).

With this hack you can use for example:

```
minisatip -s 192.168.1.100:554#no-all -s 192.168.1.101
```

In this example the first server will use the hack and the second no. And the minisatip can serve "pids=all" to any SAT>IP client.

You agree?

catalinii:
Hi Lars,

this is fine to be implemented, if you want you can make a patch for that.

Thank you

lars18th:
Hi @catalinii ,

Now I'm a quite busy. Sorry!

So, I request if you can add it, as you know the code very well.
If not then I'll try in the future.

Regards.

lars18th:
> Hi Lars,
> 
> this is fine to be implemented, if you want you can make a patch for that.
> 
> Thank you

Done #549 !
I hope you like to merge it. :wink:

lars18th:
I close it as it's resolved!

PR #549, opened 2019-02-27
NO-PIDS-ALL support

lars18th:
When your SAT>IP doesn't support "pids=all" or you have troubles with it, then you can prepend the character `~` to the command line (like `-s ~dvbt:192.168.1.3:554`) and then when a client requests all pids, then it receives only pids 0-20.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ Help
 * -R --document-root directory: document root for the minisatip web page and images
 
 * -s --satip-servers [~][DELSYS:][FE_ID@][source_ip/]host[:port] - specify the remote satip host and port with delivery system DELSYS, it is possible to use multiple -s 
-	* ~ When using this simbol at start the "pids=all" call is replaced with "pids=0-20"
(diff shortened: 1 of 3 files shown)


lars18th:
Hi @catalinii ,

This PR is ready to merge it!

catalinii:
Does this patch work on ant device? If yes can you provide a log?

lars18th:
> Does this patch work on ant device?

Sure! It works with _any_ device! You can try it with your current server. Only add "~" at the start.

> If yes can you provide a log?

No. Sorry! However, the patch is quite simple: it only _translates_ all `pids=all` to `pids=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20`.
Nothing more, nothing less.
:wink:

lars18th:
Hi @catalinii ,

Please, re-read https://github.com/catalinii/minisatip/issues/448#issuecomment-403522855

I do the work. So, why not accept this patch?

catalinii:
Hey @lars18th,

Sorry for late reply, but generally I am a bit busy.

Thanks for the patch

Issue #547, opened 2019-02-26
Posible BUG: Threshold disabled?


lars18th:
Hi @catalinii ,

Regarding my glitch problems (described in #544 and not solved) I discovered something that I feel is a bug.
See this block of the code: https://github.com/catalinii/minisatip/blob/7ccaff807c295b40eb64378cfbb4db6feb43eedb/src/stream.c#L1051
- Here is the point of the code where the `threshold` is used.
- This var is initialized as `int threshold = opts.udp_threshold;`.
- And when the threshold is pased then `flush_all = 1;`.
- But **no one** is using this `flush_all` var after setting it!!

So, I try to modify this line:
https://github.com/catalinii/minisatip/blob/7ccaff807c295b40eb64378cfbb4db6feb43eedb/src/stream.c#L1086
with a simple `ad->flush = flush_all ;` and it seems to work.

What you think?

lars18th:
Hi,

I discovered that the glitch described here is produced by the issue #570. And it has a PR pending that solves the problem. So I close this issue.

Issue #544, opened 2019-02-25
DVBAPI: Sync or Async


lars18th:
Hi,

I ask if the current implementation of the DVBAPI is working SYNC or ASYNC.

Why I ask for this?
Because, as I commented in #471 I have a continous glitches in the client... just when the DVBAPI is used.

Anyone with similar problems?

9000h:
not sure what hardware you use for minisatip and what client did show the issues, but I do have an older Panasonic TV (2014) which show some artifacts. I did change with -H3:50 the flush behavior and now it's clean.

lars18th:
> I did change with -H3:50 the flush behavior and now it's clean.

I'll check it! Thank you.

lars18th:
Hi @catalinii ,

Regarding the glitches related to the dvbapi... I see this message a lot of times:
https://github.com/catalinii/minisatip/blob/e4ee5fa587c33605c4baa8eba75d38b15ab5af61/src/dvbapi.c#L216

And I feel the code needs to do a `break` after this block, as the pids doesn't require to update if the same file is requested. So the code can be:
```
				else
				{
					LOG("dvbapi: filter for pid %d and key %d already exists, fid %d", _pid, k->id, new_filter_id);

					if (set_filter_flags(new_filter_id, FILTER_ADD_REMOVE | isEMM))
					{
						k->filter_id[fpos] = -1;
						k->pid[fpos] = -1;
					}
+					break;
				}
```

I'm right?

catalinii:
I do not think is needed, this code is supposed to re-enable the filter for that pid, which could not be in enabled state.

lars18th:
> I do not think is needed, this code is supposed to re-enable the filter for that pid, which could not be in enabled state.

Hi @catalinii ,

I think the opposite because:
1. The pid is active (and it's the ECM of the program).
2. The message indicates that the pid is already active.
3. Futhermore the message appears even if no error is found in the Oscam processing.
4. And finally, if this `break` isn't introduced then what the code execute is doing is **only** a simple "removes and reinsert" of the same pid! Ans this generates innecessary calls to `update_pids()`.

Please, try it! I'm convinced this `break` is required.

lars18th:
When the PR is merged I'll close this issue. The patch resolves the problem!

PR #550, opened 2019-02-27
Fixes some decoding troubles

lars18th:
This patch implements fixes described in:
- https://github.com/catalinii/minisatip/issues/547#issue-414611044
- https://github.com/catalinii/minisatip/issues/544#issuecomment-467194984

So, it improves the decoding.

### DIFF
--- a/src/dvbapi.c
+++ b/src/dvbapi.c
@@ -220,6 +220,7 @@ int dvbapi_reply(sockets *s)
 						k->filter_id[fpos] = -1;
 						k->pid[fpos] = -1;
 					}
+					break;
 				}
 			}
 			if (i >= 0 && fid >= 0 && k)
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -503,6 +503,8 @@ void update_cw(SPMT *pmt)
 		return;
 	}
 	LOGM("%s: pmt %d, parity %d, CW %d", __FUNCTION__, pmt->id, pmt->parity, pmt->cw ? pmt->cw->id : -1);
+	if (pmt->cw)
+		return;
 	pmt->cw = NULL;
 	for (i = 0; i < ncws; i++)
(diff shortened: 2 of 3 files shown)


src/stream.c
```diff
@@ -1083,7 +1083,7 @@ int read_dmx(sockets *s)
 	if (!send)
 		return 0;
 
-	ad->flush = 0;
+	ad->flush = flush_all;
```
catalinii: I think flush_all is not needed at all by looking at the code.

lars18th: Ok. So perhaps it's a good idea to remove all code with `flush_all`.

src/dvbapi.c
```diff
@@ -220,6 +220,7 @@ int dvbapi_reply(sockets *s)
 						k->filter_id[fpos] = -1;
 						k->pid[fpos] = -1;
 					}
+					break;
```
lars18th: Hi @catalinii ,

I'm sure this change is necessary.
It solves problems with my Oscam server.

src/stream.c
catalinii: Yes is better

src/dvbapi.c
lars18th: Hi @catalinii ,

You tested this? It _really_ resolves troubles!

catalinii:
Hi @lars18th,

I can accept this as long as you remove the flush_all code, or do not touch stream.c at all.

lars18th:
> Hi @lars18th,
> 
> I can accept this as long as you remove the flush_all code, or do not touch stream.c at all.

Done! :wink:

src/pmt.c
```diff
@@ -503,6 +503,8 @@ void update_cw(SPMT *pmt)
 		return;
 	}
 	LOGM("%s: pmt %d, parity %d, CW %d", __FUNCTION__, pmt->id, pmt->parity, pmt->cw ? pmt->cw->id : -1);
+	if (pmt->cw)
+		return;
```
lars18th: Hi @catalinii ,

This change fixes the trouble commented at #570

PR #552, opened 2019-03-02
Fix values for snr strength ber 

9000h:
looks like we use signed int but it should be unsigned int
https://www.mail-archive.com/linux-media@vger.kernel.org/msg70287.html


9000h:
I hope I do not miss anything, the sym ber in adapter.c may need to be adjusted also but there is no macro for 32 bit yet

PR #558, opened 2019-03-07
fix continuity count cc

9000h:


src/stream.c
```diff
@@ -837,14 +837,20 @@ int check_cc(adapter *ad)
...
 	if ((p = find_pid(ad->id, 8192)))
 		return 0;
 
 	for (i = 0; i < ad->rlen; i += DVB_FRAME)
 	{
 		b = ad->buf + i;
-		pid = (b[1] & 0x1f) * 256 + b[2];
+		pid = (b[1] & 0x1f) << 8;
+		pid |= b[2];
 		if (pid == 8191)
```
lars18th: Hi @9000h ,

If this code improves how to get the pid number (I assume is more portable between different platforms), then update this MACRO too:
https://github.com/catalinii/minisatip/blob/e4ee5fa587c33605c4baa8eba75d38b15ab5af61/src/pmt.h#L35

9000h: will see what @catalinii has to say about, and there are more places to

src/stream.c
```diff
@@ -837,14 +837,20 @@ int check_cc(adapter *ad)
 	int pid;
 	int packet_no_sid = 0;
 	unsigned char *b;
+        
+	if (ad->buf[0] != 0x47) {
```
catalinii: I am not sure this looks right, you are skipping the entire method for one broken packet.
I think is better if we log it and skip processing it.

9000h: ok did remove that part

9000h:
you are right

PR #563, opened 2019-03-15
fix dvbapi issue

9000h:
the last updates here did break  dvbapi

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -1567,7 +1567,7 @@ int process_pmt(int filter, unsigned char *b, int len, void *opaque)
 
 		if (pmt->stream_pids < MAX_PMT_PIDS - 1)
 		{
-			pmt->stream_pid[pmt->stream_pids].type = stype;
+			pmt->stream_pid[pmt->stream_pids++].type = stype;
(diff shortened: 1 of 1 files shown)


catalinii:
This changes from the diff are final?

9000h:
if there is anything else then I do not see it

catalinii:
pmt->stream_pid[pmt->stream_pids++].type = stype;

This will increment stream_pids so the stype and spid will not use the same stream_pids index.

9000h:
ok then I do not know how to fix it right, do you have proper fix?

catalinii:
The previous code was fine. What issue did you see that you wanted to fix?

9000h:
I need to do more investigation, the issue was with vdr + satip plugin where encrypted channels did not work as before the last changes.

9000h:
the issue did happen with server Ubuntu 18.04 minisatip and oscam, client Ubuntu 18.04 kernel 5.02 firefox with 2 open tabs (oscam log, minisatip status) vdr 2.4.0 + satip plugin. 
I do not have detailed logs but I believe it has something to do with firefox as under the same conditions with chromium there is no issue.

PR #576, opened 2019-04-04
Encapsulate HTTP responses when using persistent requests

lars18th:
When using the HTTP protocol for streaming, it's possible to process multiple requests (aka Persistent Mode). And this is great! However, when doing the streaming the HTTP response is sended in between of the TS packets. Then it's possible to "encapsulate" the response inside a NULL TS packet.

This patch implements this and allows the client to change pids or request new streams without interrupting the streaming.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,23 @@
-dist: xenial
+dist: bionic
 language: c
-install: apt-get install libdvbcsa-dev dvb-apps
-
+before_install:
+- sudo apt-get -qqy update && sudo apt-get install libdvbcsa-dev dvb-apps libssl-dev
+  -qqy
+- echo -n | openssl s_client -connect https://scan.coverity.com:443 | sed -ne '/-BEGIN
+  CERTIFICATE-/,/-END CERTIFICATE-/p' | sudo tee -a /etc/ssl/certs/ca-
+script:
+- "./configure"
+- make -j 4
+- make test
+env:
+  global:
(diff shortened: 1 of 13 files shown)


lars18th:
Sorry, to not squash!
The good patch is here: https://github.com/catalinii/minisatip/pull/576/commits/130602f430873e4d22735efff6ae6f6ab2c20e9c

lars18th:
> I think there are 2 issues with this patch

Perhas even more than 2. :wink: 
However, please note this:

> 1. conceptually, you cannot have multiple requests into http as long as the data is of unspecified length, more specifically when you ask for a TS stream which is not supposed to finish.

The HTTP 1.1 protocol standard **includes** support for Persistent Connections. However, this isn't true when streaming (sending a bitstream without end). Thus you're right and the current behaviour of minisatip is not standard.

**But** this behaviour in fact works. Why? Bacuase the socket is open to receive and process new requests even while streaming. And this is good!

The only drawback is than the HTTP Response from the server is sended between TS packets. Fortunately this is done between two TS packets. But without this patch the client needs to discard the HTTP response, or it losts the synchronization of the stream.

> 2. your code does not handle the case when resp has a siE bigger than 184. In that case you endup accessing negative offsets.

Sorry, this isn't true!
The case of a response with a length > 184 is resolved. See this part of code: https://github.com/catalinii/minisatip/commit/130602f430873e4d22735efff6ae6f6ab2c20e9c#diff-02ae217c34bfd1e1a97e810edf9f4d12R1878
We only copy **up to** 184 chars (and I put the end of string at 189th).
Futhermore, the common response (200 OK) consumes less than 184 bytes. And, any case with length > 184 it's uncommon (in case of error, then the HTTP connection is closed after the response). So the code **is** completly safe.

> My advise is to use RTSP over tcp for this purpose, as a bunch of TS packeta are bundled together and then you can enqueue an answer.

Yes RSTP over TCP can be an alternative for this pourpose. However, the HTTP is still supported here. 

> Http cannot do what you expect it to do.

No, the **current** implementation does the work with Persistent Requests. I'm not incorporating this functionality. I only "fixing" the case of the HTTP response in the middle of the streaming.

> Also please keep in mind that I am not aware of any major http client supporting this behaviour.

False. I'm using it with some HTTP clients for a long time. Why you think I created this patch? Without it the client needs to resync the TS every time I send a new request. And with this patch the stream is kept clean.

Please, rethink on these facts:
- Current HTTP Persistent Requests work (for a long time).
- HTTP Clients can request changes when streaming for changing, for example, the pid list. So the behaviour is useful.
- The patch does not alter anything that is not already implemented.
- Futhermore, some positive side effect is included: printing a debug message when the HTTP streaming starts.

So, please reconsider why not include this patch.

catalinii:
And which clients work with this patch ? Does those clients not support RTSP?


Again I still inclined to not accept the patch as I am implementing non standard functionality that needs to be carried over...

lars18th:
> And which clients work with this patch ? Does those clients not support RTSP?
> 
> Again I still inclined to not accept the patch as I am implementing non standard functionality that needs to be carried over...

When the PR #580 is accepted, I'll rewrite this PR and I prepare another one improved. I'm sure you'll be confortable with the new one.

Stay tuned!

PR #580, opened 2019-04-09
Initialization of start_streaming when using HTTP

lars18th:
Correct use of the start_stream variable when using the HTTP protocol.
It prints too a log message when the streaming starts.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -844,6 +844,9 @@ int check_cc(adapter *ad)
 	for (i = 0; i < ad->rlen; i += DVB_FRAME)
 	{
 		b = ad->buf + i;
+		if (b[1] & 0x80)
+			continue;
+
 		pid = PID_FROM_TS(b);
 		p = find_pid(ad->id, pid);
 
--- a/src/stream.c
+++ b/src/stream.c
@@ -312,7 +312,6 @@ int start_play(streams *sid, sockets *s)
 	if (!sid->tp.apids && sid->tp.pids && (!sid->tp.pids[0] || !strcmp(sid->tp.pids, "0")))
 		s->flush_enqued_data = 1;
 	sid->do_play = 1;
-	sid->start_streaming = 0;
 	if (s->type != TYPE_HTTP)
 		sid->start_streaming = 0;
 	sid->tp.apids = sid->tp.dpids = sid->tp.pids = sid->tp.x_pmt = NULL;
--- a/src/stream.c
+++ b/src/stream.c
@@ -313,6 +313,8 @@ int start_play(streams *sid, sockets *s)
 		s->flush_enqued_data = 1;
 	sid->do_play = 1;
 	sid->start_streaming = 0;
+	if (s->type != TYPE_HTTP)
+		sid->start_streaming = 0;
 	sid->tp.apids = sid->tp.dpids = sid->tp.pids = sid->tp.x_pmt = NULL;
 
 	ad = get_adapter(sid->adapter);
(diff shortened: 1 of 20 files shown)


lars18th:
Hi @catalinii ,

Until I'll review my PR #576 , please merge this patch. I hope you don't see any inconvenience in that. It merely improves the use of the start_streaming member without touching the current behaviour.

Thank you!

src/stream.c
```diff
@@ -313,6 +313,8 @@ int start_play(streams *sid, sockets *s)
 		s->flush_enqued_data = 1;
 	sid->do_play = 1;
 	sid->start_streaming = 0;
+	if (s->type != TYPE_HTTP)
```
catalinii: Probably no point in differentiating between stream types here, so you can delete this.

src/stream.c
```diff
@@ -741,7 +744,14 @@ void flush_streamb(streams *sid, unsigned char *buf, int rlen, int64_t ctime)
 	int i, rv = 0, blen, len;
 
 	if (sid->type == STREAM_HTTP)
+	{
 		rv = sockets_write(sid->rsock_id, buf, rlen);
+		if (sid->start_streaming == 0)
```
catalinii: You should find a new code place where you can set sid->start_streaming to 1 that reaches all code paths and you do not repeat the same code block. (basically set start_streaming to 1 in 1 location instead of 3).

catalinii: Other than those comments everything looks good.

src/stream.c
lars18th: Sorry, in this version one delete is missing.

This check is required, because for HTTP requests this code is executed in every request. So after the first request the next one doesn't needs to reset the `start_streaming`. For HTTP the initialization is done in line 353:
https://github.com/catalinii/minisatip/pull/580/files#diff-b4c34cf7e2b840ebe24729f28a8c5f27R353

src/stream.c
lars18th: > You should find a new code place where you can set sid->start_streaming to 1 that reaches all code paths and you do not repeat the same code block. (basically set start_streaming to 1 in 1 location instead of 3).

No. As I commented, the `start_streaming`to 1 in 1 is a simple error (copy&paste from my other PR). The correct place is here. The start streaming is detected for HTTP here, as it's when the data (of the streaming) is started to be writed in the socket.

lars18th: > Other than those comments everything looks good.

Please, compile with this patch and check the logs with HTTP streaming.
I'm sure you see it good. So, please commit it! :wink:

src/stream.c
lars18th: The missing deletion (which has confused you) is now here:
https://github.com/catalinii/minisatip/pull/580/files#diff-b4c34cf7e2b840ebe24729f28a8c5f27L315

lars18th:
Hi @catalinii ,

You'll merge it?
I feel all is resolved/commented with _this_ PR.

lars18th:
Hi @catalinii ,

Still waiting... note that I resolved/commented _all_ of your requirements.

atarist:
hi @lars18th
do you have time to try add code/pach for dvb-c?
tnx

lars18th:
> hi @lars18th
> do you have time to try add code/pach for dvb-c?

Sorry! Very busy at time. Futhermore, my objective with minisatip it's focused on HTTP support. So I can't work on DVB-C. However, perhaps you can use another good SAT>IP software server: SATPI (https://github.com/Barracuda09/SATPI/commits/master). The main developer uses DVB-C as the primarly source.

I hope it helps you.
Regards.

lars18th:
Thank you @catalinii ! :wink:

PR #581, opened 2019-04-09
small printf format fix

9000h:


catalinii:
I reverted this PR as it breaks stats page on 32 bit systems

PR #603, opened 2019-08-16
netceiver: Enable "API socket" needed for netcvdiag and netcvupdate

M-Reimer:
The sources, libmcli is created from, also contain some useful tools that can be used to maintain and debug the NetCeiver unit.

netcvdiag: Diagnosis tool to query NetCeiver status
netcvupdate: Update and config tool to change NetCeiver firmware and settings
netcvlogview: Live log viewer

Most of the tools require an unix socket to communicate with the NetCeiver platform.

A very small change makes it possible that minisatip creates this socket, so a running minisatip can act as the required "bridge" for the NetCeiver tools.

I've tried this change with minisatip connected to an NetCeiver equipped with 3 DVB-S2 dual tuners. No issues at all. It really helps with setting up if the maintenance tools are working.

### DIFF
--- a/src/netceiver.c
+++ b/src/netceiver.c
@@ -409,6 +409,10 @@ void find_netcv_adapter(adapter **a)
 	if (recv_init(opts.netcv_if, 23000))
 		LOG("Netceiver init failed");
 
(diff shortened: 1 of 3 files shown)


9000h:
makes sense for me

Issue #602, opened 2019-08-10
minisatip does not work with latest oscam


okmikel:
Hi,

With the latest oscam r11537 decryption via dvbapi does not work, last working is r11532. I get the following error in oscam log:

```
Aug 10 08:07:33 han oscam[12357]: 05059956 c   (dvbapi) Client connected: 'minisatip/1.0-unknown' (protocol version = 2)
Aug 10 08:07:40 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:07:44 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:07:48 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:07:52 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:07:55 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:07:59 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:08:03 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:08:06 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
Aug 10 08:08:10 han oscam[12357]: 05059956 c   (dvbapi) ERROR: Received CA PMT list update for unknown program
```
I am using the latest version form git of minisatip.

Mikel

peos3:
I just want to verify something:  

Can you please checkout commit f64d5104c0466a65a6ecb1c8a88156923ac46ec3 and try this again?

okmikel:
I've already on the latest git source and applied this commit already. But it did not work before also.

peos3:
Okay, I was just curious if the little change in commit 6ae72f53dd5a3af25c1627cf37b85c46a4f4561b broke this, because it was the last change to pmt.c. But if it didn't work before it has nothing to do with it.

catalinii:
I saw there are dome updates to the protocol on oscam side

holli73:
hello,

[...]

PR #605, opened 2019-08-25
dvbapi fix for oscam svn

9000h:
change for latest oscam suggested by nautilus7
https://github.com/catalinii/minisatip/issues/602#issuecomment-524543165


9000h:
right

PR #612, opened 2019-10-03
Parse RTCP signal

lars18th:
This fixes two problems with some faulty servers:
 - If the server sends correct info about lock and snr, but not for strength, then rewrite the signal value.
 - Try to parse the RTCP APP packet with the correct "SES1" token to get the status message. This fixes with some servers that write not aligned packets. And when the server sends nothing then it continues to work.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -313,9 +313,11 @@ void set_adapter_signal(adapter *ad, char *b, int rlen)
 	{
 		sscanf(signal + 1, "%d,%d,%d", &strength, &status, &snr);
 		// Workaround for faulty servers (level=0)
+		// Ex. XORO: "SES1...aver=1.0;src=1;tuner=1,0,1,15,10744,h,dvbs,qpsk,off,0.35,22000,56;pids=0,100,200,400,500,600,17,16"
 		if (strength==0 && status>0 && snr>0)
 			 strength = 1;
 		// Workaround for faulty servers (quality=0)
+		// Ex. AVM 6490: "SES1...Qver=1.2;src=1;tuner=1,106,1,0,538.00,8,dvbc,256qam,6900,,,,1;pids=0,118,2351,2352"
 		if (snr==0 && strength>0 && status>0)
 			 snr = 7;
 		if (ad->strength != strength || ad->snr != snr)
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -312,9 +312,12 @@ void set_adapter_signal(adapter *ad, char *b, int rlen)
 	if (signal)
 	{
 		sscanf(signal + 1, "%d,%d,%d", &strength, &status, &snr);
-		// Workaround for faulty servers
+		// Workaround for faulty servers (level=0)
 		if (strength==0 && status>0 && snr>0)
 			 strength = 1;
+		// Workaround for faulty servers (quality=0)
+		if (snr==0 && strength>0 && status>0)
+			 snr = 7;
 		if (ad->strength != strength || ad->snr != snr)
 			LOG(
 				"satipc: Received signal status from the server for adapter %d, stength=%d status=%d snr=%d",
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -312,6 +312,9 @@ void set_adapter_signal(adapter *ad, char *b, int rlen)
 	if (signal)
 	{
 		sscanf(signal + 1, "%d,%d,%d", &strength, &status, &snr);
+		// Workaround for faulty servers
+		if (strength==0 && status>0 && snr>0)
+			 strength = 1;
 		if (ad->strength != strength || ad->snr != snr)
 			LOG(
 				"satipc: Received signal status from the server for adapter %d, stength=%d status=%d snr=%d",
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

Please read the section 3.5.15.2 of the Specifications:
http://www.satip.info/sites/satip/files/resource/satip_specification_version_1_2_2.pdf

9000h:
hi, not sure for which server it will improve somthing, for the AVM 6490 it does not 
grafik

lars18th:
> hi, not sure for which server it will improve somthing, for the AVM 6490 it does not

Hi @9000h ,

In a lot of LOW QUALITY servers, tipically STBs with SAT>IP support.

This patch improves two issues:

1. When the RTCP packet is malformed, or when it sends more info than expected. Now the code search for the correct `SES1` token for reading the tuner status. Belive me, in some cases the new code is the difference between getting or not the tuner status.
2. When the server sends STRENGTH=0 every time, but it send correct SNR and LOCK values. In this case, without this patch the `minisatip` never starts to receive the stream.

And regarding your problem with the AVM 6490, if you send me a capture of the RTCP packets then we can try to fix it.

Regards.

9000h:
the AVM 6490 did only provide SNR=0, there is nothing to fix on the client end

lars18th:
Hi @9000h ,

> the AVM 6490 did only provide SNR=0, there is nothing to fix on the client end

In this case, we can fix it adding a fake SNR value, or using the STRENGTH value to map it

Please, can you print here a capture of the RTCP sended by the AVM?

Note:
You can use my fork of https://github.com/lars18th/sat-ip-proxy with `node ssdp-server.js & ./rtsp-proxy -ddd satip-server-avm-ip` and use any client to receive data from your AVM server throught the proxy. It's a nice solution to capture the RTSP and RTCP traffic of SAT>IP Client-Server traffic.

Regards.

lars18th:
Hi @9000h ,

I'm waiting for your capture of the RTCP data from the AVM 6490 to improve this patch.

Regards.

lars18th:
Hi @9000h ,

I feel @catalinii is waiting to merge this patch until I finish it... Please, can you send me some capture of the RTCP packet from AVM 6490? If you don't have it, then try this tool to capture it: https://github.com/lars18th/sat-ip-proxy

Regards.

9000h:
tcpdump from AVM 6490 Firmware 7.12
https://we.tl/t-LldD3QT0Jw

lars18th:
> tcpdump from AVM 6490 Firmware 7.12
> https://we.tl/t-LldD3QT0Jw

Thank you!
That's all I need (just one RTCP packet):
```
0000   70 85 c2 24 64 77 c8 0e 14 ab 29 d5 08 00 45 80   p..$dw....)...E.
0010   00 b0 c9 9d 00 00 40 11 ca 4b c0 a8 b2 01 c0 a8   ......@..K......
0020   b2 81 13 89 57 ef 00 9c a2 e8 80 c8 00 06 6b 8b   ....W.........k.
0030   45 67 e1 48 17 64 d4 38 f3 66 01 09 41 56 00 00   Eg.H.d.8.f..AV..
0040   2b af 00 e0 70 08 81 ca 00 04 6b 8b 45 67 01 09   +...p.....k.Eg..
0050   41 56 4d 20 44 56 42 2d 43 00 80 cc 00 18 6b 8b   AVM DVB-C.....k.
0060   45 67 53 45 53 31 00 00 00 51 76 65 72 3d 31 2e   EgSES1...Qver=1.
0070   32 3b 73 72 63 3d 31 3b 74 75 6e 65 72 3d 31 2c   2;src=1;tuner=1,
0080   31 30 36 2c 31 2c 30 2c 35 33 38 2e 30 30 2c 38   106,1,0,538.00,8
0090   2c 64 76 62 63 2c 32 35 36 71 61 6d 2c 36 39 30   ,dvbc,256qam,690
00a0   30 2c 2c 2c 2c 31 3b 70 69 64 73 3d 30 2c 31 31   0,,,,1;pids=0,11
00b0   38 2c 32 33 35 31 2c 32 33 35 32 00 00 00         8,2351,2352...
```

lars18th:
Hi @9000h ,

Fixed! Please, try it.
I changed to SNR=7, instead of 15, because the lower value can just indicate an unkown value.
However, if you prefer, then use 15.

Please, @catalinii can you check it too?

Regards.

9000h:
the real fix has to be done in the AVM SAT->IP server software in my opinion

lars18th:
> the real fix has to be done in the AVM SAT->IP server software in my opinion

Sure! However, with this _simple_ change _any_ SAT>IP server that sends invalid (=0) values of SNR or STRENGTH (but not both), can work with the minisatip client functionality.

Please, @catalinii  review it! Without this patch my XORO STB never works, as the the streaming doesn't start because the `strength==0`.

Regards.

catalinii:
From my point og view it looks good. It would be nice to add some type of packet hex in the code as comment that shows what you are fixing with the new added code

lars18th:
> From my point og view it looks good. It would be nice to add some type of packet hex in the code as comment that shows what you are fixing with the new added code

I'm not sure what you're requesting... Add comments with examples of the RTCP problematic packets?

catalinii:
Yes, basically add a comment containing an example of packet that this code fixes and was not working before.

lars18th:
> Yes, basically add a comment containing an example of packet that this code fixes and was not working before.

Done!
Please, merge it. :wink:

lars18th:
Hi @catalinii ,

I'm waiting. Please, provide feedback.

PR #621, opened 2019-10-23
Disable insane log

lars18th:
This log line prints partial content of an entire file. The log is then unreadable. It's preferable to disable this line (only enbale it for deep debug).


catalinii:
You can do DEBUGM instead of commenting

PR #622, opened 2019-10-23
Fix satipc: don't send "plsc" param if no "plsm"

lars18th:
Some SAT>IP servers refuse to serve requests for DVB-S2 tuners when `&plsc=` is used if not used `&plsm=`.
One example is the TVHeadend SAT>IP server.
With this fix the SAT>IP client functionality will be compatible with TVH when requesting DVB-S2 transponders.


lars18th:
Hi @catalinii ,

Please merge it.

PR #623, opened 2019-10-23
Assign adapters to specific sources

lars18th:
This PR is a Work-In-Process...

It adds support for SRC Groups: When using DVB-S/S2 tuners, you can group them using the SRC (sat position) parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,13 @@ Help
 	eg: 2:0-1284[-1111]
 	* When * character is used before frequency, force 13V only for setup
 
+* -U --sources sources_for_adapters: select the adapters for each source
+	* eg: -U 0-2:*:3:1,5,7 (no spaces between parameters)
+	- In this example: for SRC=1 only 0,1,2; for SRC=2 all: for SRC=3 only 3; and for SRC=4 the 1,5,7 adapters.
+	- For each position (separated by : ) you need to declare all the adapters that use this position
+	- The special char * indicates all adapters for this position
+	- If errors or by default all adapters have enabled all positions
+
 * -j --jess jess_string - same format as -u 
 
 * -w --http-host http_server[:port]: specify the host and the port (if not 80) where the xml file can be downloaded from [default: default_local_ip_address:8080] 
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -1572,11 +1572,61 @@ void set_unicable_adapters(char *o, int type)
 
 void set_sources_adapters(char *o)
 {
-	// Hardcoded values
-	source_map[0] = 0 | 0b1001;
-	source_map[1] = 0 | 0b0111;
-	source_map[2] = 0 | 0b1001;
-	source_map[3] = 0 | 0b0110;
+	int i, la, st, end, j, k, adap;
+	int64_t all = 0;
+	char buf[1000], *arg[40], *sep;
+	SAFE_STRCPY(buf, o);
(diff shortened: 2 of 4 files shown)


lars18th:
Ready to merge.
Please check it!

catalinii:
At the high level it looks good. I have a question. How do you deal with polarization and band for a source?

lars18th:
Hi @catalinii ,

> At the high level it looks good.

Great!

> I have a question. How do you deal with polarization and band for a source?

This patch doesn't change anything about polarization or bands. It only targets SOURCES.

A case of use to improve understanding:

- You receive 3 wires. The first is connected to A-B switch with two orbital positiona. The second is connected directly to a LNB with a fixed antenna. And the third is connected too to an individual antenna but with a positioner. So, in this case we assume: LNB1 is SRC=1 and SRC=2, LNB2 is SRC=3, and LNB3 is SRC=4. So, if we have _three_ different tuners (internal or external), then we can use:
`minisatip --sources 0:0:1:2` and then the tuners are restricted to the corresponding source. And when a SETUP command is processed the correct tuner (if it's free) will be selected.

However, the current implementation is the most simple. In the future (after a merge) some improvements can be added. For example:

1. Remove the "SRC" when passing it to the tunner. At time, in the above example, tuners 1 and 2 will receive position=BA (SRC=3) or position=BB (SRC=4). So it will be interesting to overcome this (if the user wants it because the target lnb is connected to some switch). For example with `minisatip --sources 0:0:~1:~2`.
2. Map the sources to different targets and/or limit the bands (High-Vertical, High-Horizontal, Low-Vertical, Low-Horizontal). But in this case I don't know much about any specific syntax to describe it.

So, anyway I hope you will test it and accept to merge it soon.
Regards.

lars18th:
Hi @catalinii ,

Still waiting for this patch. I hope you merge it soon!

lars18th:
Thank you @catalinii ! :wink:

catalinii:
@lars18th I have reverted this commit as it was causing issues

lars18th:
> @lars18th I have reverted this commit as it was causing issues

What happens?
Please, expose more information. I'm using it for weeks and I don't have any problem.

Regards.

9000h:
https://github.com/catalinii/minisatip/issues/678

lars18th:
OK, I'll try to improve it (when I have more time).

In the mean time: The sources value can vary between 1-16 ? In the SAT>IP standard only 1-4 are supported for DVB-S2 sources, and nothing for DVB-T/C.

lars18th:
Hi,

I'll update this patch. However I need to check two things to improve it:
1. The correct value for the `MAX_SOURCES` is 16, it's this value sufficient?
2. The patch requires to support 8 or more tuners, correct?

When these two will be solved, then I'll publish a version 2. You agree?

ctoda:
1) You can have a DISEQC 1.0 (4 sources) and 1.1  (16 sources) together, so 64 sources
2) The patch should not limit in any way the number of tuners (adapters). Use MAX_ADAPTERS to see how many tuners are defined

lars18th:
Hi @ctoda ,

> 1. You can have a DISEQC 1.0 (4 sources) and 1.1  (16 sources) together, so 64 sources
> 2. The patch should not limit in any way the number of tuners (adapters). Use MAX_ADAPTERS to see how many tuners are defined

A new an improved version is publised. See PR #736 .

PR #631, opened 2019-11-05
Change minimum limit of the app adapter buffer

lars18th:
The default value for the app adapter buffer (72192 == ADAPTER_BUFFER) is currently the lower limit.
This limit has been changed (up to 1316 bytes, which is one RTP packet) so the user can use smaller values.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -722,8 +722,8 @@ void set_options(int argc, char *argv[])
 		{
 			sscanf(optarg, "%d:%d", &opts.adapter_buffer, &opts.dvr_buffer);
 			opts.adapter_buffer = (opts.adapter_buffer / 188) * 188;
-			if (opts.adapter_buffer < ADAPTER_BUFFER)
-				opts.adapter_buffer = ADAPTER_BUFFER;
+			if (opts.adapter_buffer < 1316)
+				opts.adapter_buffer = 1316;  // 188 * 7 = 1316
 #ifdef AXE
 			opts.dvr_buffer += 7 * 188 - 1;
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

This change has sense when you want to reduce the jitter and the SAT>IP client has a small buffer. For example, with a Panasonic TV using the value `-b 31584:5775360` you can reduce some glitches that sometimes appear.

I hope you'll merge it soon!
Regards.

9000h:
The only Panasonic series I have seen with visible artifacts was the AX, CX and DX should work with the current defaults in LAN environments. The artifacts on the AX series went away with option -H5:50 also.

lars18th:
Hi @9000h ,

> The only Panasonic series I have seen with visible artifacts was the AX, CX and DX should work with the current defaults in LAN environments. The artifacts on the AX series went away with option -H5:50 also.

First of all, I have now a new GX800 and I'm using the extremly `-H 1:5 -b 31584:5775360` values to overcome the glitches. In my environment the problem is caused because the minisatip is running in SAT>IP _client mode_, so the incoming TS is received over the network with some jitter. So it's required to process it in small chunks to overcome the small receive buffer in the Panasonic TV.

After commenting on this, I want to say that this patch doesn't change _anything_ in the current code (the default app buffer continues to be the value of ADAPTER_BUFFER), except that the user can use _LOWER_ values for the app buffer. Before this patch the user can't use values lower than 72192. And that constraint is too strong. So this patch relaxes it.

I hope @catalinii will merge it soon!
Regards.

PR #638, opened 2019-11-13
Fix incorrect handle of SAT>IP client close

lars18th:
It resolves two problems:

1. Free the adapter pid list when closing the adapter.
2. Don't send incorrect SETUP command when committing DELETE pids before the first connection of the SAT>IP client mode.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -437,13 +437,13 @@ int close_adapter(int na)
 	}
 
 	LOG("closing adapter %d  -> fe:%d dvr:%d, sock:%d, fe_sock:%d", na, ad->fe, ad->dvr, ad->sock, ad->fe_sock);
-	ad->enabled = 0;
-	if (ad->close)
-		ad->close(ad);
 	//close all streams attached to this adapter
 	//	close_streams_for_adapter (na, -1);
 	mark_pids_deleted(na, -1, NULL);
-	update_pids(na);
+	// update_pids(na); // No update pids as the adapter is closing!
+	ad->enabled = 0;
+	if (ad->close)
+		ad->close(ad);
 	//      if(ad->dmx>0)close(ad->dmx);
 	if (ad->fe > 0)
 		close(ad->fe);
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -1040,6 +1040,8 @@ void satipc_commit(adapter *ad)
 	if (sip->lap + sip->ldp == 0)
 		if (!sip->force_commit || !ad->tp.freq)
 			return;
+	if (sip->ldp > 0 && sip->last_connect == 0 && !sip->want_tune)
+		return;
 
 	if (sip->expect_reply)
 	{
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -1040,8 +1040,11 @@ void satipc_commit(adapter *ad)
 	if (sip->lap + sip->ldp == 0)
 		if (!sip->force_commit || !ad->tp.freq)
 			return;
-	if (sip->ldp > 0 && sip->last_connect == 0 && !sip->want_tune)
-		return;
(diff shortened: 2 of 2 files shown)


src/adapter.c
```diff
@@ -437,13 +437,13 @@ int close_adapter(int na)
 	}
 
 	LOG("closing adapter %d  -> fe:%d dvr:%d, sock:%d, fe_sock:%d", na, ad->fe, ad->dvr, ad->sock, ad->fe_sock);
-	ad->enabled = 0;
```
lars18th: Hi @catalinii ,

You can't do it here as then the find_adapter() function used by the mark_pids_deleted() will return FALSE.

src/adapter.c
```diff
@@ -437,13 +437,13 @@ int close_adapter(int na)
...
-	if (ad->close)
-		ad->close(ad);
 	//close all streams attached to this adapter
 	//	close_streams_for_adapter (na, -1);
 	mark_pids_deleted(na, -1, NULL);
-	update_pids(na);
+	// update_pids(na); // No update pids as the adapter is closing!
```
lars18th: If you update the pids then the a new PLAY command is sended... but we're sending a TEARDOWN, so the commit is not required: we only need to free the pid list.

src/satipc.c
```diff
@@ -1040,6 +1040,8 @@ void satipc_commit(adapter *ad)
 	if (sip->lap + sip->ldp == 0)
 		if (!sip->force_commit || !ad->tp.freq)
 			return;
+	if (sip->ldp > 0 && sip->last_connect == 0 && !sip->want_tune)
```
lars18th: This case appears if the session is already open (transponder) but a new adapter is created. In this case, the DELETE PIDS is commited. But in this case, nothing requires to be done. This check resolves the problem.

lars18th: Hi @catalinii ,

Please, don't commit this at time! Perhaps this line is wrong.

lars18th: Now fixed!

src/satipc.c
```diff
@@ -1040,6 +1040,11 @@ void satipc_commit(adapter *ad)
 	if (sip->lap + sip->ldp == 0)
 		if (!sip->force_commit || !ad->tp.freq)
 			return;
+	if (sip->ldp > 0 && (sip->last_cmd == RTSP_TEARDOWN || sip->last_setup < 0))
```
lars18th: This new code fixes the problem _ancient_ pids to remove are requested. In this case (when the session with the target SAT>IP server is empty) the `dpid` list is cleared (these pids are discarded, as they have no sense).

lars18th:
Hi @catalinii ,

Please, review it. I'm sure you can merge it.

lars18th:
Hi @catalinii ,

Any reason to not review this patch?
Please, comment about it.

lars18th:
Hi @catalinii ,

Still waiting... :fearful:

catalinii:
Hey,
Did you see my comments related to update_pids and the changes from adapter.c ?

lars18th:
> Hey,
> Did you see my comments related to update_pids and the changes from adapter.c ?

Sorry, no! Where they are?

lars18th:
Hi @catalinii ,
Please, I need your feedback!

lars18th:
Hi @catalinii ,

I don't want to be a pain, but I don't see any reason to not merge this patch.
Can you explain your reasons for the hold?

Thank you!

catalinii:
Hey @lars18th ,

did u see my comments on the code?
Thank you

catalinii:
Screen Shot 2019-12-07 at 5 59 52 PM

lars18th:
Hi @catalinii ,

I craeted a new version (#730) of this patch, so I close this.

Please, note that your comments in the code doesn't appears in my account. So yours screenshort has been useful.

Issue #635, opened 2019-11-12
SATIP client mode: Trouble with some clients with the pid list


lars18th:
Hi @catalinii ,

Doing some debug I discovered that some clients request a _mixed_ pid list. See this example:
```
PLAY rtsp://192.168.1.27:554/stream=2?addpids=0,17,16&delpids=16,17,0 RTSP/1.0
CSeq: 7
Session: 0288999596
```

I know it, this doesn't have sense. The resulting list is... _the current list_ so, `addpids` and `delpids` requires to be removed. However, this is the request **FROM** the minisatip. So it's the `satipc` module that generates this inconsistent SETUP command.

Please, can you fix this problem? Before passing the pid list to the SAT>IP server perhaps a reduction function can be applied.

Thank you!

lars18th:
Hi @catalinii ,

This problem isn't cosmetic. I detected that some SAT>IP servers have big troubles when some pids are inside both lists: `addpids=` _and_ `delpids=` in the same command. Different behaviours can be shown:

- The stream has glitches (internally the server removes the pid and adds it in a short time).
- The server refuses to continue (I see this when the duplicated pid has the PCR).
- Or the server resets the stream (this is the worst case, when the server doesn't accept duplicates at all).

So, please merge the PR #639. It resolves a lot of problems with the small list of SAT>IP servers that have troubles with the SAT>IP Client implementation in minisatip.

Regards.

lars18th:
I close as the fix is merged!

PR #639, opened 2019-11-14
SATIPC: Fix add/del pid list with duplicates

lars18th:
When a new pid is added/deleted, check first if the same pid is inside the other list.
If this is true, then remove it from the two lists. That's equivalent to not insert in any list and remove it from the other.


lars18th:
Hi @catalinii ,

This resolves the issue #635 .

Please, commit it!

PR #649, opened 2019-11-22
Socket Select: Prioritize DVR sockets

lars18th:
In the reading loop process first DVR sockets, then all others.

This patch improves the performance with low powered servers.

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -696,6 +696,8 @@ void *select_and_execute(void *arg)
 	unsigned char buf[2001], *pos;
 	int err;
 	struct pollfd pf[MAX_SOCKS];
+	int prio;
+	int reorder[MAX_SOCKS];
 	int64_t lt, c_time;
 	int read_ok;
 	char ra[50];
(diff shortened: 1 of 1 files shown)


catalinii:
Hey @lars18th,

Without a clear indication that this fixes the issue I cannot accept the patch, as the downside is that for slow devices this can be a performance problem (this loop is executed many times) on multiple threads.

lars18th:
Hi @catalinii ,

> Without a clear indication that this fixes the issue I cannot accept the patch, as the downside is that for slow devices this can be a performance problem (this loop is executed many times) on multiple threads.

Please, note these two things:

1. The objective of this patch is to READ _first_ from the DVR socket. I'm sure you agree that this will improve the performance with slow devices, as without large incoming buffers (so less consumed memory) is more difficult to lost a packet from the tuner. Right?
2. The loop to reorder the reading streams is quite simple. I'm sure it doesn't consumes a lot of resources. However, if you want it can be optimized. In this case, the problem is that inside the socket processing loop the var `max_sock` is modified. So, is not posible a simple: one-time-computation of the reordering. So, any idea to improve this loop?

Regards.

catalinii:
Hey Lars,

1) You are right, reading faster will optimize the reading from DVR first which could end up saving some packets. If you do not see any E_OVERFLOW (which I did not see just with minisatip by itself), probably is not broken.
2. If we can do some caching that would reduce the overhead of the loop ...

lars18th:
Hi @catalinii ,

> 1. You are right, reading faster will optimize the reading from DVR first which could end up saving some packets. If you do not see any E_OVERFLOW (which I did not see just with minisatip by itself), probably is not broken.

Yes, I don't see any E_OVERFLOW either. However, working with SAT>IP clients with _small_ input buffers I see errors generated by JITTER problems. And the cause of the jitter is related to processing _other_ sockets when DVR data is pending. So this patch targets these problems.

> 1. If we can do some caching that would reduce the overhead of the loop ...

OK. I'll try to reorder the list only at some specific points. In any case, a _perfectly_ orderer list perhaps is not required; and we can just do a reordering when a DVR socket is added (to put it at front). You agree?

catalinii:
You could try to rewrite the whole loop more optimal and account for this case too

lars18th:
Hi @catalinii ,

I withdraw this PR as it will not be necessary.

Issue #461, opened 2018-09-28
Performance issues


hsteinhaus:
I am running minisatip on a rather small embedded board (Dragonboard 410), which is generally working fine. However, minisatip consumes quite a lot of CPU power and also causes a bit more heat buildup than initially expected.

A short glimpse on "top" reveals, that the majority of load is system load, i.e. it is not caused by some difficult calculations. strace reveals, that there is an enormous frequency of system calls in the worker threads, mainly write() and poll(). 

Is this really caused by the way minisatip is working or is there some potential for optimization, e.g. by reading/writing bigger chunks or so?

catalinii:
Hi,

Unless you watch encrypted channels that uses dvbcsa (-o option) to decrypt you should not see high load from minisatip. Basically minisatip does poll, read, write loop for every adapter and the amount of data that is read depends on -b parameter. 

If you do not watch encrypted channels then feel free to upload the log

hsteinhaus:
No encrypted channels involved and no cams/softcams installed. My DVB source for this test is a Netceiver with one double-S2 card installed. I have increased the buffers by a factor of 4 from the numbers mentioned as default in the help message.

I have just recorded two strace dumps of the most active threads - one of the worker threads (strace2.txt) seems to do more than 6000 syscalls per second. The observed load of 30-35% of a core seems to be reasonable for that load pattern, even if the syscalls wouldn't do anything (the SnapDragon 410 cores are not really strong - low-cost smartphone style hardware). The OS is Linaro Buster 18.01 (an arm64 Debian variant with kernel 4.14.0).

For the test, minisatip was launched with this command line:
``` /usr/bin/minisatip -f -b 288768:23101440 -g --netceiver eth0:2```

hsteinhaus:
[strace2.txt](https://github.com/catalinii/minisatip/files/2429355/strace2.txt)
[strace1.txt](https://github.com/catalinii/minisatip/files/2429356/strace1.txt)

perexg:
BTW: I am using this patch for satip-axe (which has also limited CPU power):

<pre>
diff --git a/src/stream.c b/src/stream.c
index bde2992..8443a4f 100644
--- a/src/stream.c
+++ b/src/stream.c
@@ -987,7 +987,7 @@ int process_dmx(sockets *s)
 #endif
 
        rlen = ad->rlen;
-       int packet_no_sid = check_cc(ad);
+       const int packet_no_sid = 0 /* check_cc(ad) */;
 
[...]

PR #651, opened 2019-11-25
WIP: Optimize the reading of the SAT>IP client frontend

lars18th:
When reading from the SAT>IP client socket, try to get all RTP datagrams until the DVR buffer is full.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -498,7 +498,7 @@ void satip_close_device(adapter *ad)
 	sip->enabled = 0;
 }
 
-int satipc_read(int socket, void *buf, int len, sockets *ss, int *rb)
+int satipc_read1(int socket, void *buf, int len, sockets *ss, int *rb)
 {
 	unsigned char buf1[20];
 	uint16_t seq;
@@ -533,6 +533,23 @@ int satipc_read(int socket, void *buf, int len, sockets *ss, int *rb)
 		return 1;
 	}
 	*rb -= 12;
+	return (*rb >= 0);
+}
+
+int satipc_read(int socket, void *buf, int len, sockets *ss, int *rb)
+{
+	int partial = 0;
+
+	satipc_read1(socket, buf, len, ss, rb);
+	partial = *rb;
+	while (partial + 1316 <= len && *rb > 0)
+	{
+		satipc_read1(socket, buf + partial, len - partial, ss, rb);
+		partial += *rb;
+	}
+	*rb = partial;
+	if (*rb > 0 && ss->type == TYPE_DVR && (opts.debug & LOG_DMX))
+		dump_packets("read ->", buf, *rb, ss->rlen);
 	return (*rb >= 0);
 }
 


lars18th:
Hi @catalinii ,

This patch is absolutly required for the SAT>IP client functionality. Without it the minisatip only reads just 1 datagram (from 188 to 1316 bytes) for _EVERY_ `select()` call. And that's very inefficient. Therefore, the patch reads all available datagrams, until the DVR buffer is full or no more data is available.

Please, merge it!

catalinii:
Hey,

If your computer is very fast that it can end up reading just 188 bytes at a timr.
Try doing a ctrl+z for a second and then fg and see how big was your read.

The downside with your patch is that if your buffer is large say 512k and you are listening on a radio station the satip_read will  end up taking seconds without giving the control back to other sockets in the same thread. This could break things such as decrypting, ....and definitely single threaded usages.

Also if satip_read is called and reads just 188 bytes, it process_dmx will not be called right away. So it is just a bunch of selects/reads then a call to process_dmx (see read_dmx)

lars18th:
> The downside with your patch is that if your buffer is large say 512k and you are listening on a radio station the satip_read will end up taking seconds without giving the control back to other sockets in the same thread. This could break things such as decrypting, ....and definitely single threaded usages.

Really?
The code reads more times _ONLY_ if some data is in the socket buffer. If no more data in the socket, then it doesn't waits to full the buffer.

Please, review the data of the patch!

catalinii:
@lars18th , while I am sure you can address some of the comments that I have here, the idea is that you will end up adding most of the logic from select_and_execute.

This will duplicate the logic and make the code more complex as you have 2 portions that can interact in a interesting ways with each other.

Again by doing multiple syscalls for the same socket you will delay the read for other sockets from the same threads which is not right.

For no real benefits, I think we should not merge this patch.

lars18th:
Hi @catalinii ,

Perhaps it's my fault, as I don't explained the reason for this patch. So please let me some minutes to comment about it:

- For _me_ the `satipc` functionality is a must have of the minisatip project. For this reason my top objective is to improve this part as most as we can. So, note that this patch only changes the `satipc.c` source. It doesn't create any interference with the rest of the server.
- The next thing to comment is the objective to target with this patch. With the current code, the value of the DVR_BUFFER is the size of bytes to process for each loop. However, this strategy isn't the best when we're processing packets _from the network_, instead from a local tuner device. Why? Because the jitter problem of the network. Note the current behaviour when **1** RTP packet arrives: If the reading queue has size 0, then the packet waits until the buffer is full or a timeout flushes it. And this current behaviour increases the jitter.
- So a more intelligent strategy it's possible: To process _every_ RTP when it arrives. And only if the buffer has more than one RTP packet waiting in the _socket_ buffer, then read an entire chunk (until the maximum of the DVR_BUFFER).
- This strategy for sure improves the jitter, as each incoming RTP packet is inserted in the internal processing loop, without waiting for future packets. Futhermore, if the incoming stream is quite fast for the internal loop processing, then in this case as the RTP packets are incoming in the socket buffer, then several chunks of packets are processed.

In any case, please note that I'm using this patch for several weeks and it outperforms the current implementation. If you're unsure, then do a test.

The other option to alleviate the problem of jitter is much more complex: use the RTP time stamps to read in a synchronized way or use the PCR. And the first one will only work if the SAT>IP server generates them well and does not have internal synchronization problems. In any case much more expensive and complex, don't you think?

Regards.

9000h:
btw, how did you measure the jitter and on which side of minisatip client/server did you see it.
If the RTP packets are not stamped with the right 90khz time you may try to catch ghosts. 
Most of the SAT>IP server do not have corret timestamps like Panasonic/DD Octonet etc.

catalinii:
@lars18th 

1) I am also using satipc module daily and is really important to work as my satip stream travels around 200ms from end to end.
2) From select man page:
"select() and pselect() allow a program to monitor multiple file
       descriptors, waiting until one or more of the file descriptors become
       "ready" for some class of I/O operation (e.g., input possible).  A
       file descriptor is considered ready if it is possible to perform a
       corresponding I/O operation (e.g., read(2), or a sufficiently small
       write(2)) without blocking."
Basically it can read maximum ADAPTER_BUFFER from the kernel side DVR_BUFFER, however once the smallest amount of data becomes available, it will be read. In case something happens and minisatip waits decrypting packets, then DVR buffer will fill and read more than 1 packet. 
As multiple read_dmx can be called before a process_dmx is starting to process the packet, -H can be used to shorten the time data is actually send over the network to satip clients.

Your patch proposes to perform multiple reads one after another:
- First the DVR buffer may be bigger than the adapter buffer (why ? devices with low amount of memory)
- Adapter buffer is optimized for decryption with dvbcsa. Of course, the adapter buffer can be set for 
 - If you keep reading, you give the opportunity to the scheduler to change the process so the cost of second read is much higher  while you may not actually read much data (second read will be performed right after first read and not give opportunity to get another packet from the network).
- Also the performance of minisatip decreases by increasing the number of syscalls. As a result I am working on reducing it even further on write path. This patch has the potential of at least doubling the number of number of syscalls for DVR socket which by far the most used socket.
About reducing the number of syscalls I was able to increase the speed of minisatip by 20-40% just by reducing the number of write syscalls by 20x (eg switching from rtsp over tcp to http) on arm enigma2 boxes.

3) Processing every packet individually will increase the overhead / packet which decreases the performance. If the goal is to reduce the jitter, as I said, use -H. More details about the code to reduce jitter:
https://github.com/catalinii/minisatip/blob/master/src/stream.c#L1093


If you are still sure that this improves jitter / increases the performance please feel free to post some logs without and with the patch showing what works better. I would be curious how many times in the second read you actually get data and what is the average  number of syscall per select even that you do with this code to actually judge. As you submit the patch, the burden to prove that is better than the current approach is on you.

lars18th:
> If the RTP packets are not stamped with the right 90khz time you may try to catch ghosts.
> Most of the SAT>IP server do not have corret timestamps like Panasonic/DD Octonet etc.

In this case, you have two options:

1. You can detect if the RTP timestamps are wrong, and then use the incoming clock to regenerate them.
2. Or leave to the user the option to remove the original RTP timestamps and recalculate them.

In any case it will be better to have the _incoming_ rate of the stream, than nothing. Do you agree?

catalinii:
@lars18th the issue you're trying to solve is satipc using UDP? Or TCP?

lars18th:
> @lars18th the issue you're trying to solve is satipc using UDP? Or TCP?

Regular SAT>IP with RTP stream over _UDP_.

catalinii:
@lars18th Happy new year.

The solution to this problem is simple, feel free to adapt this patch:

Use recvmmsg syscall instead of readv. You may need a larger buffer where to read the data instead of directly the adapter buffer. See https://github.com/catalinii/minisatip/issues/461 for more details.

recvmmsg will read in one syscall multiple UDP packets and then you can process them individually instead of reading one.

lars18th:
Hi,

I havae a WIP patch for timestamping the satipc inputs. However, I still leave this PR open until I can cleanup with a new PR. Please @catalinii not close it for now.

Regards.

lars18th:
I close this PR as #827 implements a more optimized solution.

PR #656, opened 2019-11-27
use SO_REUSEPORT if available

9000h:
REUSEPORT will work on linux starting with 3.9 kernel version(which was released back in 2013)
https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ/14388707#14388707

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -147,6 +147,15 @@ int udp_bind(char *addr, int port)
 		close(sock);
 		return -1;
 	}
+#ifdef SO_REUSEPORT
+	if (setsockopt(sock, SOL_SOCKET, SO_REUSEPORT, &optval, sizeof(optval)) < 0)
+	{
+		LOG("udp_bind failed: setsockopt(SO_REUSEPORT): %s",
+			strerror(errno));
+		close(sock);
+		return -1;
+	}
+#endif
 
 	if (bind(sock, (struct sockaddr *)&serv, sizeof(serv)) < 0)
 	{


lars18th:
Hi,

This patch doesn't work with Debian 7 x86_64:
```
[02/12 18:01:50.942 main]: minisatip version -3e0dea9, compiled with s2api version: 050A
[02/12 18:01:50.942 main]: Built with dvbcsa
[02/12 18:01:50.942 main]: Built without CI
[02/12 18:01:50.942 main]: Built with dvbapi
[02/12 18:01:50.942 main]: Built without AES (OpenSSL)
[02/12 18:01:50.942 main]: Built with tables processing
[02/12 18:01:50.942 main]: Built with pmt processing
[02/12 18:01:50.942 main]: Built with satip client
[02/12 18:01:50.942 main]: Built with linux dvb client
[02/12 18:01:50.942 main]: Built with backtrace
[02/12 18:01:50.942 main]: Built without netceiver
[02/12 18:01:50.942 main]: Built without ddci
[02/12 18:01:50.942 main]: Built without t2mi
[02/12 18:01:50.943 main]: New TCP listening socket 3 at 0:0:0:0:554, family 2
[02/12 18:01:50.943 main]: New TCP listening socket 4 at 0:0:0:0:8080, family 2
[02/12 18:01:50.943 main]: udp_bind failed: setsockopt(SO_REUSEPORT): Protocol not available
```

And without log the worst is that the error isn't printed. Please, fix this problem!

9000h:
unfortunately there are systems out where it is defined but not working
and it is not covered by POSIX, so I believe it's better to revert the commit
https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_10_16

PR #657, opened 2019-11-29
Fix incorrect Session in DESCRIBE responses

lars18th:
When the DESCRIBE request desn't have an associated Session, then the response should not include any Session.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -1277,6 +1277,8 @@ int read_rtsp(sockets *s)
 		snprintf(buf, sizeof(buf),
 				 "Content-type: application/sdp\r\nContent-Base: rtsp://%s/",
 				 get_sock_shost(s->sock));
+		if (!sess_id)  // Request doesn't have Session !
+			s->sid = -1;
 		http_response(s, 200, buf, sbuf, cseq, 0);
 	}
 	else if (strncmp(arg[0], "OPTIONS", 7) == 0)


lars18th:
You can check it with:

```
(echo 'DESCRIBE rtsp://192.168.178.129/ RTSP/1.0'; echo 'Cseq: 1';echo 'Connection:close'; printf '\r\n\r\n';) | nc 192.168.178.129 554
```
In this case the response contains ALL running streams and without "Session:"

```
(echo 'DESCRIBE rtsp://192.168.178.129/?stream=1 RTSP/1.0'; echo 'Cseq: 1';echo 'Session: 3432424';echo 'Connection:close'; printf '\r\n\r\n';) | nc 192.168.178.129 554
```
In this case the response contains only the stream 1 (if it's running) and with "Session:" (different number if the value of the session doesn't exists).

```
(echo 'DESCRIBE rtsp://192.168.178.129/?stream=1 RTSP/1.0'; echo 'Cseq: 1';echo 'Connection:close'; printf '\r\n\r\n';) | nc 192.168.178.129 554
```
Now as expected (and fixed) the response doesn't contain "Session:"

So, all OK!

lars18th:
Patch updated! Ready to merge it.
Please @catalinii .

lars18th:
Hi @9000h ,

It works for you?

lars18th:
Hi @catalinii ,
Hi @9000h ,

A lot of pending patches. Please, can you confirm that this works for you?

9000h:
s->sid = -1 will invalidate valid sid and struct struct_sockets *s is may used later on

lars18th:
> s->sid = -1 will invalidate valid sid and struct struct_sockets *s is may used later on

No @9000h . You mixed concepts!

Here in the funciton `read_rtsp()` the `s->sid` has the value `-1` when the received command doesn't contains any session identifier. So, the line `s->sid = -1` in this patch only _removes_ the session id readed from the command when no `stream` parameter is indicated. And as the command DESCRIBE can accept the paramter `session` _optionally_ then the change is correct, and doesn't disturb the rest of the code.

As a summary, take note of the _three_ scenarios:
1. DESCRIBE without stream and without session --> OK.
2. DESCRIBE with stream and with session --> OK.
3. DESCRIBE without stream and _with_ session --> NOT PERMITED, so clear the session id.

What does the patch is catching the case three and convert it to case 1.
Nothing more.

So, the patch is safe! :wink:
Regards.

catalinii:
@lars18th , it is unclear here if s->sid is -1. This is persisted across multiple packets. Is possible that describe was run after setup, which means that the connection between the sockets and stream will be broken, which will end up leaking sessions on a buggy client (that may not send session at some point in describe).

In my opinion this is dangerous, and is probably easier to handle this case without touching the s->sid.

lars18th:
Hi @catalinii ,

> In my opinion this is dangerous, and is probably easier to handle this case without touching the s->sid.

OK. What you prefer then? To add a new flag parameter to the current `http_response()` funtion, or to create some workaround to not print the `Session:` value in the response in this case?

catalinii:
Adding a flag sounds good.

9000h:
@lars18th 
the query with rtsp://192.168.178.129/?stream=1 is not valid, there should no ? before stream=

lars18th:
> @lars18th
> the query with rtsp://192.168.178.129/?stream=1 is not valid, there should no ? before stream=

That's right @9000h !
However, minisatip _in this case_ overcomes the additional `?` char, so the result is the same.

lars18th:
Hi @9000h ,

I don't have time to implement the suggestion from @catalinii about using a flag to overcome the  "Session:" in the response. And as I created this patch for you... please, can you finish it using a flag?

Thank you! :wink:

9000h:
i would suggest
```
--- stream.c    2019-12-18 07:06:45.531910928 +0100
+++ stream.c.new        2019-12-18 07:30:28.143482547 +0100
@@ -77,7 +77,7 @@
        int do_all = 1;
        int is_ipv6 = 0;

-       if (s->sid == -1 && strchr(req, '?'))
+       if (s->sid == -1 && strchr(req, '?') && !strstr(req, "?stream="))
                setup_stream(req, s);

        sidf = get_session_id(s->sid);
```

PR #659, opened 2019-11-30
fix for rtsp issue

9000h:
there are wrong 0 length buffer
Most likely not an RTSP packet sock_id: 16 sid: 0 rlen: 0, dropping ....


9000h:
maybe you have a better idea to fix it, but I could not see any side effects in my tests

PR #662, opened 2019-12-02
Fix rtp info seq

9000h:
2019-12-02 16_32_12-Window
for me looks good so far


lars18th:
Ummm....

```
minisatip.c:1269:31: warning: format ‘%jd’ expects argument of type ‘intmax_t’, but argument 5 has type ‘int’ [-Wformat=]
       "RTP-Info: url=%s;seq=%jd;rtptime=%jd\r\nRange: npt=0.000-",
```

PR #664, opened 2019-12-02
Revert "use SO_REUSEPORT if available"

9000h:
Reverts catalinii/minisatip#656
unfortunately there are systems out where it is defined but not working
and it is not covered by POSIX, so I believe it's better to revert the commit
https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_10_16

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -176,15 +176,6 @@ int udp_bind(char *addr, int port, int ipv4_only)
 		close(sock);
 		return -1;
 	}
-#ifdef SO_REUSEPORT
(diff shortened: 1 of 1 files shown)


lars18th:
Why not use some define in the `configure` to enable it instead of full remove it?

9000h:
as it's not covered by POSIX we may should avoid it
It is even more wore if you like to use a pre-complied binary on different systems, then you cannot be sure if it will work or not.
The only way would be to ignore the error if it fails.
I give up here.

lars18th:
> Why not ignoring the error?

Done in PR #665 , so please @catalinii merge it!

PR #665, opened 2019-12-03
SO_REUSEPORT tryfix

9000h:
RedHat defines SO_REUSEPORT with a kernel which does not support
it and returns ENOPROTOOPT so in this case ignore the error.

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -178,10 +178,7 @@ int udp_bind(char *addr, int port, int ipv4_only)
 	}
 #ifdef SO_REUSEPORT
 	if (setsockopt(sock, SOL_SOCKET, SO_REUSEPORT, &optval, sizeof(optval)) < 0)
-	{
-		LOG("udp_bind failed: setsockopt(SO_REUSEPORT): ignored %s",
-			strerror(errno));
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @9000h ,

Last version of the patch runs in Debian 7 x86_64:
```
[04/12 09:08:40.192 main]: New TCP listening socket 3 at 0:0:0:0:554, family 2
[04/12 09:08:40.192 main]: New TCP listening socket 4 at 0:0:0:0:8080, family 2
[04/12 09:08:40.192 main]: udp_bind failed: setsockopt(SO_REUSEPORT): ignored Protocol not available
[04/12 09:08:40.192 main]: New UDP socket 5 bound to 0.0.0.0:1900
[04/12 09:08:40.193 main]: setting multicast for 239.255.255.250
[04/12 09:08:40.193 main]: udp_bind failed: setsockopt(SO_REUSEPORT): ignored Protocol not available
[04/12 09:08:40.193 main]: New UDP socket 6 bound to 239.255.255.250:1900 (mcast:239.255.255.250)
```

So, I suggest to merge it to @catalinii .
Regards.

PR #669, opened 2019-12-05
describe session off by one fix

9000h:
sid starts by 0 but stream= did start by 1

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -80,7 +80,7 @@ char *describe_streams(sockets *s, char *req, char *sbuf, int size)
 	if (s->sid == -1 && strchr(req, '?'))
 		setup_stream(req, s);
 
-	sidf = get_session_id(s->sid);
+	sidf = get_session_id(s->sid + 1);
 	sid = get_sid(s->sid);
 	if (sid)
 		do_play = sid->do_play;


lars18th:
Hi @9000h ,

I don't see the sense of this patch...

Copy&paste from Specs "3.5.7 Listing available media streams (DESCRIBE)":
```
“o=” gives the originator of the session “-“ <session-id> <session-version> IN for internet and the IP address of the
server. The session-id and session-version values shall be present but are not used currently.
```

So these numbers session-id and session-version are irrlevant. And futhermore, the session-id is the number returned by `get_session_id()`.

9000h:
it's may bit more complicated, if you have one rtsp playing (stream=1) and then request in a second session
 (echo 'DESCRIBE rtsp://192.168.178.129/stream=1 RTSP/1.0'; echo 'Cseq: 1';echo 'Connection:close'; printf '\r\n\r\n';) | nc 192.168.178.129 554
RTSP/1.0 200 OK
Date: Thu, Dec 5 16:18:26 2019 GMT
CSeq: 1
Server: minisatip/1.0.2-3a6d9f3
Content-type: application/sdp
Content-Base: rtsp://192.168.178.129/
Content-Length: 249

v=0
o=- 0000000000 0000000000 IN IP4 192.168.178.129
s=SatIPServer:1 4,2,2
t=0 0
m=video 0 RTP/AVP 33
c=IN IP4 0.0.0.0
a=control:stream=1
a=fmtp:33 ver=1.0;src=1;tuner=1,254,1,2,12480,v,,,,dvbs,27500,34;pids=3327,3328
b=AS:5000
a=inactive
^C

it shows inactive for the running stream=1 and  o=- 0000000000 0000000000 
so I think there something mixed up

lars18th:
> it shows inactive for the running stream=1 and o=- 0000000000 0000000000
> so I think there something mixed up

Ummmm... You're right that the stream needs to be `active`. But this is controlled by `do_play` of the sid, and it gets from `get_sid()`.

I feel you need to catch more LOGs with the full RESPONSE description to discover why and when the stream is described as "inactive" (in fact, when the sid is empty... so when the stream exists but it doesn't have a valid session id). Perhaps the problem is a defered creation of the session id.

9000h:
the whole describe_streams function needs a rewrite, my fix is also not working right

9000h:
most of the DESCRIBE works only correct in session, but by a requests outside session it's gives wrong/incomplete info back

PR #670, opened 2019-12-05
wrong do_play

9000h:
use the do_play from the stream

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -82,8 +82,6 @@ char *describe_streams(sockets *s, char *req, char *sbuf, int size)
 
 	sidf = get_session_id(s->sid);
 	sid = get_sid(s->sid);
-	if (sid)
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @9000h ,

You noticed that?
https://github.com/catalinii/minisatip/pull/670/files#diff-b4c34cf7e2b840ebe24729f28a8c5f27L85-L86

Please @catalinii , comment it.

9000h:
I'm unsure which clients do really rely on that describe info especially outside session, but I think it needs a rewrite to cover both situations.

catalinii:
I am not sure is safe to use sid->do_play as sid can get to be NULL, which will crash the app ?

maybe is better to set do_play just before that ?

9000h:
is sid2 not already valid inside the if(sid2

PR #672, opened 2019-12-08
rtp seq rollover fix

9000h:
the seq need a rollover


9000h:
Initial sequence number SHOULD be random, according to RFC 3550
According to RFC 3711, it should be random within 2^15 bit

9000h:
the vdr-satip-plugin will show 1 RTP error if the stream did start with a random seq as it did not take the seq from the RTP-Info

PR #674, opened 2019-12-09
Improve unconditional logs

lars18th:
When running without logs we need to know when the process is started, and when is closed.
Futhermore, a message regading the closing procedure it's added to detect some kill errors.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -134,8 +134,8 @@ int satipc_reply(sockets *s)
 	__attribute__((unused)) int rv;
 	get_ad_and_sipr(s->sid, 1);
 	s->rlen = 0;
-	LOG("satipc_reply (sock %d) handle %d, adapter %d:\n%s", s->id, s->sock,
-		s->sid, s->buf);
+	LOG("satipc_reply (sock %d) handle %d, adapter %d :", s->id, s->sock, s->sid);
+	LOGM("%s", s->buf);
 
 	if ((timeout = strstr((char *)s->buf, "timeout=")))
 	{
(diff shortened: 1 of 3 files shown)


lars18th:
Hi @catalinii ,

With last commits the log prints unconditional errors without `-l general`. With this it's possible to see the cause to not run when some port is already used when log is not enabled (a very common error when port 554 is in use).

Please, check it!
Regards.

src/satipc.c
```diff
@@ -134,8 +134,8 @@ int satipc_reply(sockets *s)
 	__attribute__((unused)) int rv;
 	get_ad_and_sipr(s->sid, 1);
 	s->rlen = 0;
-	LOG("satipc_reply (sock %d) handle %d, adapter %d:\n%s", s->id, s->sock,
-		s->sid, s->buf);
+	LOG("satipc_reply (sock %d) handle %d, adapter %d :", s->id, s->sock, s->sid);
+	LOGM("%s", s->buf);
```
catalinii: Hey, can we leave those things still at LOG ?
Next time we need to investigate I need to enable verbose for this module too...

Same for the next logm.

Otherwise, LGTM.

lars18th: > Hey, can we leave those things still at LOG ?
> Next time we need to investigate I need to enable verbose for this module too...

The idea is just the opposite! Include the full RTSP commands between the minisatip process and the SAT>IP server when the **general** log is enabled is annoying. The objective needs to be to enable them **only** when the debug of this module is enabled. So, when you enable the log of other modules you don't populate the log with irrelevant lines. Note that the line printing the _send_ of the message and the _receive_ of the response is every time included. So _only_ when you require it you can enable it (with a simple `-l satipc`).

I hate extensive logs!

src/utils.h
```diff
@@ -205,7 +205,10 @@ static inline int get_index_hash(void *p, int max, int struct_size, uint32_t key
 
 #define FAIL(a, ...)               \
 	{                              \
-		LOGL(0, a, ##__VA_ARGS__); \
+		if (opts.log) {            \
```
catalinii: If oyu do not enable logging, you will not get even fatal exception which is not right IMO.

lars18th: > If oyu do not enable logging, you will not get even fatal exception which is not right IMO.

Just the opposite, again!
Check the _full_ code:
```
		if (opts.log) {            \
			LOGL(0, a, ##__VA_ARGS__); \
		} else                     \
			LOG0(a, ##__VA_ARGS__);    \
```
When no log is enabled the log is unconditional (`LOG0`) and when it's enabled then it's equal to the general log (`LOGL`).
So the code is correct.

Please, compile with the patch and test it! It works with/without logs.
Regards.

lars18th: Please @catalinii comment about it!

PR #675, opened 2019-12-09
RTP-Info fix Range white space

9000h:
After checking the with ffplay -v trace and Live555 code it seems that's unable to parse correctly the PLAY request header
'Range: npt=0.000-'
because it expects white spaces like this
'Range: npt = 0.000 -'
This then causes "start", "end" and "duration" to be 0 so can explain why it stops after sending one frame of video...


9000h:
looks like a ffplay bug in ffplay version 3.4.6-0ubuntu0.18.04.1

PR #676, opened 2019-12-11
Fix rtp timestamp

9000h:
RFC 2250 / a 90khz clock is needed
The RTP timestamp will not be passed to the MPEG
decoder. This use of the timestamp is somewhat different than
normally is the case in RTP, in that it is not considered to be the
media display or presentation timestamp. The primary purposes of the
RTP timestamp will be to estimate and reduce any network-induced
jitter and to synchronize relative time drift between the transmitter
and receiver.


9000h:
the jitter and clock drift looks better this way

2019-12-11 14_47_51-Wireshark · RTP Stream Analyse · 1234 pcapng

PR #677, opened 2019-12-13
correct DESCRIBE response

9000h:
set o= conform to https://tools.ietf.org/html/rfc4566 and use a timestamp

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -72,7 +72,7 @@ streams *get_sid1(int sid, char *file, int line)
 char *describe_streams(sockets *s, char *req, char *sbuf, int size)
 {
 	char *stream_id, dad[1000], localhost[100];
-	int i, sidf, do_play = 0, streams_enabled = 0;
+	int i, do_play = 0, streams_enabled = 0;
 	streams *sid, *sid2;
 	int do_all = 1;
 	int is_ipv6 = 0;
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @9000h ,

From the RFC:
```
5.2.  Origin ("o=")

      o=<username> <sess-id> <sess-version> <nettype> <addrtype>
        <unicast-address>

   The "o=" field gives the originator of the session (her username and
   the address of the user's host) plus a session identifier and version
   number:

   <username> is the user's login on the originating host, or it is "-"
      if the originating host does not support the concept of user IDs.
      The <username> MUST NOT contain spaces.

   <sess-id> is a numeric string such that the tuple of <username>,
      <sess-id>, <nettype>, <addrtype>, and <unicast-address> forms a
      globally unique identifier for the session.  The method of
      <sess-id> allocation is up to the creating tool, but it has been
      suggested that a Network Time Protocol (NTP) format timestamp be
      used to ensure uniqueness [13].

   <sess-version> is a version number for this session description.  Its
      usage is up to the creating tool, so long as <sess-version> is
      increased when a modification is made to the session data.  Again,
      it is RECOMMENDED that an NTP format timestamp is used.

   <nettype> is a text string giving the type of network.  Initially
      "IN" is defined to have the meaning "Internet", but other values
      MAY be registered in the future (see Section 8).
[...]
```

So, the `o=` parameters `<sess-id>` and `<sess-version>` it's true that they should be in the _format_ of a timestamp. However, their value doens't change for every DESCRIBE command. And they aren't timestamp values.

Then I suggest to add some transform from the `s->sid` (session-ID) to a `<sess-id>` with the timestramp format. For example, something like `sidf = to_timestamp(get_session_id(s->sid));`

Regards.

PR #681, opened 2019-12-18
Fix for rtsp issue

9000h:
fix for "Most likely not an RTSP packet sock_id: 16 sid: 0 rlen: 0, dropping ...."
there are unwanted buffer length 0 given to the action if the client did close the socket normal and read gives back rv 0 and no errno


9000h:
Uhh, there is a side effect with the dvbapi socket, unfortunately I did see this late.

PR #683, opened 2019-12-19
fix last commit

9000h:
there was a side effect on TCP dvbapi in rtsp drop fix
dvbapi did need to get the rlen 0

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -860,7 +860,7 @@ void *select_and_execute(void *arg)
 						read_ok ? "OK" : "NOK", rlen, master->rlen, master->lbuf,
 						ss->sock, ss->id, master->id, pos, master->buf, ss->iteration, master->action);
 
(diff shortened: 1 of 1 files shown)


9000h:
dvbapi did need to get the rlen  0, i hope there is not more like this in axe,ca

catalinii:
@9000h I will revert both those commits, because the logic is not working for the satip client. See the code here: 
https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L541
Basically read_ok is supposed to be set to 0 in case of error. As you can see read_ok = 1 and rlen = 0 is totally fine.

For the issue you have reported, ideally it should be done without a change in socketworks.c
There are couple of corner cases which at this point I am not sure how to handle, except more logging:
eg: if you have already read from the > 0 but you get socket closed: should you ignore the bytes already read (for example the client sends teardown then closes the connection).

9000h:
that's fine for me, the main issue with read on sockets is you will get back len 0 and no error if the socket is normal closed by the other end.

catalinii:
Is easier to just return if the len is 0 directly in read_rtsp

PR #687, opened 2019-12-23
DESCRIBE cosmetic

9000h:
do not process wrong query

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -77,7 +77,7 @@ char *describe_streams(sockets *s, char *req, char *sbuf, int size)
 	int do_all = 1;
 	int is_ipv6 = 0;
 
-	if (s->sid == -1 && strchr(req, '?'))
+	if (s->sid == -1 && strchr(req, '?') && !strstr(req, "?stream="))
 		setup_stream(req, s);
 
 	sidf = get_session_id(s->sid);


catalinii:
Can you upload a log highlighting the issue ?
Did u test the new patch with VLC? as it relies heavily on DESCRIBE

9000h:
it's not that important but if you do the following wrong request on an idle minisatip server, the response is also wrong.
```
(echo 'DESCRIBE rtsp://192.168.178.129/?stream=1 RTSP/1.0'; echo 'Cseq: 1';echo 'Session: 3432424';echo 'Connection:close'; printf '\r\n\r\n';) | nc 192.168.178.129 554
RTSP/1.0 200 OK
Date: Sat, Dec 28 19:13:12 2019 GMT
Session: 1472713773
CSeq: 1
Server: minisatip/1.0.2-6add3c0
Content-type: application/sdp
Content-Base: rtsp://192.168.178.129/
Content-Length: 222

v=0
o=- 1472713773 1472713773 IN IP4 192.168.178.129
s=SatIPServer:1 4,2,2
t=0 0
m=video 0 RTP/AVP 33
c=IN IP4 0.0.0.0
a=control:stream=1
a=fmtp:33 ver=1.0;src=1;tuner=1,0,0,0,0,,,,,,,;pids=
b=AS:5000
a=inactive
````

lars18th:
> it's not that important but if you do the following wrong request on an idle minisatip server, the response is also wrong.

Yes. The problem is that the current code that process the DESCRIBE command doesn't checks the `stream` and/or `Session` parameters. I suggest a ligth refactoring of this part instead of the multiple patches related (#687 and #657).

PR #695, opened 2020-01-13
try fix "no data"

9000h:
use different algo for signal is may not good for some drivers


9000h:
can you please comment on this, the dvr/dmx are probably unreleated

9000h:
will close this as the in session PLAY channel change to a non working freq is may the root cause

PR #699, opened 2020-01-29
Fix tuning to bad channels

9000h:
need to wait a bit to no confuse unicable/jess switches with repeated setup cmds after tuning to non existing/bad channels


9000h:
it's seem not fully fix it for VDR

PR #706, opened 2020-02-06
Access to change the Select timeout

lars18th:
Now the timeout of the Select function is controlled by a thread variable.

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -715,6 +715,7 @@ extern uint32_t writes, failed_writes;
 
 __thread pthread_t tid;
 __thread char thread_name[100];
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @catalinii ,

This PR doesn't change the current behaviour. However, it opens the option to modify the Select timeout. I need this functionality to implement a dejitter buffer in the satipc module. Even this change is not necessry (at time) for other modules, I prefer to merge it. The advantage now is that any working thread can change the timeout of the Select sleeping function.

I hope you agree to merge it.
Regards.

lars18th:
Thank you @catalinii for merging it! :wink:

PR #708, opened 2020-02-06
Increase grace_time

lars18th:
When using HTTP clients this improves the clean detection of decoded programs.
Useful when not using the PR #445 .

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -1836,7 +1836,7 @@ void start_pmt(SPMT *pmt, adapter *ad)
 	pmt->running = 1;
 	pmt->encrypted = 0;
 	// give 1s to initialize decoding or override for each CA
-	pmt->grace_time = getTick() + 2000;
+	pmt->grace_time = getTick() + 4000;
 	set_filter_flags(pmt->filter, FILTER_ADD_REMOVE | FILTER_CRC);
 #ifndef DISABLE_TABLES
 	send_pmt_to_cas(ad, pmt);


catalinii:
@lars18th if this is just for dvbapi feel free to override the value only there. 2s is a good value for CI and CI+ as well.

lars18th:
Hi @catalinii ,

> @lars18th if this is just for dvbapi feel free to override the value only there. 2s is a good value for CI and CI+ as well.

No, the PR isn't for fixing problems with dvbapi. The PR is for fixing problems with HTTP clients that receive "encrypted" packets when using the `-E` parameter. Using this patch then my previous PR #445 is not required. When incresing the grace_time the **initial** undecrypted packets are discarded and this solves the problem with such type of clients. Remember: the problem is with clients that start to receive not decrypted channels and then refuse to play them. To solve the problem we need to discard the _first_ non-decoded packets. Then after the start you can pass undecrypted packets to maintain the PCRs synch without disturbing the client if something fails in the CI/CI+ processing.

Futhermore, I don't think this change will have any other side effect. You see any problem?

Regards.

catalinii:
@lars18th, this feature should not be used to fix another problem, because if we change this code, the original issue for which this code was written will no longer work (or work poorly).
In this case maintaining minisatip will be increasing difficult because we are trying to solve problem X with code written for code Y.

The fact that this solves the problem from #445 this is very coincidentally.
The correct resolution for #445 and the previous similar issues we talked about is to do filtering per stream (not per adapter as it is done now).

While the code to support that is not difficult, that is the correct way of doing it.

As a result I cannot accept the PR in the way it is now as it will break the existing functionality.

lars18th:
Hi @catalinii ,

> The correct resolution for #445 and the previous similar issues we talked about is to do filtering per stream (not per adapter as it is done now).
> 
> While the code to support that is not difficult, that is the correct way of doing it.

To me it's hard to implement it. I don't have the sufficient idea of the threading of minisatip and the modules involved. Every time I work to fix a problem I'm using a lot of hours to understand the code. If you feel it's easy, then please try to modify this code to do the filtering per stream... Please!

Regards.

catalinii:
I will probably get to it at some point.

Thanks

lars18th:
Hi @catalinii ,

> I will probably get to it at some point.

You closed it. So, you'll implement it soon?

PR #715, opened 2020-02-14
remove the not working x_pmt code

9000h:
as it's not needed atm for VDR to function normal with server side decryption


catalinii:
Feel free to remove the code if not needed.

I will accept and Thanks  so you can push it once is fixed.

catalinii:
@9000h can you remove the code so we can merge ?

PR #716, opened 2020-02-15
plp_isi is overwritten intermitted and cause unwanted tune and ad close

9000h:
could not replicate the issue with this changes

### DIFF
--- a/src/dvb.c
+++ b/src/dvb.c
@@ -290,6 +290,9 @@ void init_dvb_parameters(transponder *tp)
 	tp->ds = TP_VALUE_UNSET;
 	tp->plp_isi = TP_VALUE_UNSET;
 	tp->pls_mode = TP_VALUE_UNSET;
+	tp->sys = TP_VALUE_UNSET;
+	tp->fe = TP_VALUE_UNSET;
+	tp->diseqc = TP_VALUE_UNSET;
 }
 
 void copy_dvb_parameters(transponder *s, transponder *d)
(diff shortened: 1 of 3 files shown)


9000h:
need more tests

9000h:
after some excessive testing the plp_isi issue which triggers unwanted tunes seem to be solved, the modification in stream seem to be also beneficial for VDR zapping

src/stream.c
```diff
@@ -144,6 +144,7 @@ void set_stream_parameters(int s_id, transponder *t)
 {
 	streams *sid;
 
+	LOG("%s", __FUNCTION__);
```
catalinii: Is this needed ?

src/dvb.c
```diff
@@ -188,15 +196,8 @@ int detect_dvb_parameters(char *s, transponder *tp)
...
-	while (*s > 0 && *s != '?')
-		s++;
-
-	if (*s == 0)
-		LOG_AND_RETURN(0, "no ? found in URL");
-
-	s++;
-	if (strstr(s, "freq="))
-		init_dvb_parameters(tp);
+//	if (strstr(s, "freq="))
```
catalinii: why do we fail to initialize the dvb_parameters here ?

src/stream.c
9000h: not really

src/dvb.c
9000h: https://github.com/catalinii/minisatip/blob/9af1d6560bc7e248860b13d980cc91dc3bdfccd2/src/dvb.c#L280
this will zero out the most of the tp params that set before to -1

9000h: I did some tests again and it seem to make no difference if it's in or not.

PR #726, opened 2020-04-01
Improve logs of the HTTP module

lars18th:
The HTTP module serves HTTP and RTSP requests. So it's quite complex to isolate RTSP commands only.
With this small change, when using `-l http` only RTSP messages are full shown. But when changing to `-v http` the HTTP messages are also printed.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -1372,7 +1372,7 @@ int read_http(sockets *s)
 	s->rlen = 0;
 
 	LOG("read HTTP from %d sid: %d: ", s->sock, s->sid);
-	LOGM("%s", s->buf);
+	DEBUGM("%s", s->buf);
 
 	split(arg, (char *)s->buf, ARRAY_SIZE(arg), ' ');
 	//      LOG("args: %s -> %s -> %s",arg[0],arg[1],arg[2]);
@@ -1904,7 +1904,14 @@ void http_response(sockets *s, int rc, char *ah, char *desc, int cseq, int lr)
 		(lresp == sizeof(resp) - 1) ? "message truncated" : "", s->sock,
 		get_sockaddr_host(s->sa, ra, sizeof(ra)), get_sockaddr_port(s->sa),
 		lr, s->id);
-	LOGM("%s", resp);
+	if (s->type == TYPE_HTTP)
+	{
+		DEBUGM("%s", resp);
+	}
+	else
+	{
+		LOGM("%s", resp);
+	}
 
 	struct iovec iov[2];
 	iov[0].iov_base = resp;


lars18th:
Hi @catalinii ,

Any problem with this PR?

For a long time I do requests for improving the logs. The UPNP and HTTP messages are a pain when testing the minisatip process. I don't agree with the `grep` filtering, as I need to see all lines, but not the content of the HTTP messages. With this small change, nothing is changed, except that you can "mute" the _content_ of the HTTP and UPNP messages (the line indicating the message continues showing). And you can "unmute" it with the `-v http` parameter. So it can't generate troubles.

Think on this: when you want to debug it's simple to use only RTSP messages. And when using RTSP you don't want to see HTTP/UPNP messages. However, when you try with HTTP clients, then you will don't see RTSP messages as none RTSP client is connected. So with just enabling the full log of the HTTP module you got the expected functionality. Nothing is lost!

Please, consider to merge this patch.
Regards.

ctoda:
This diff will make it very dificult for me to debug issues from the users
There will be plenty of more debug messages, bigger logs, harder to load and so on.

Personally I do not think is a good idea.

catalinii:
This diff will make it very dificult for me to debug issues from the users
There will be plenty of more debug messages, bigger logs, harder to load and so on.

Personally I do not think is a good idea.

lars18th:
Hi @catalinii ,

> This diff will make it very dificult for me to debug issues from the users
> There will be plenty of more debug messages, bigger logs, harder to load and so on.

Please, review the code. I feel you don't check it well...

The code _doesn't change anything_ and it _doesn't increase the debug_! In fact, it _hides_ the _content_ of the HTTP/UPNP messages if the full log (using `-v` instead of `-l`) is not enabled. So you don't lose any relevant information, and you don't see more irrelevant information.

Note this:

- With `-l general` you will see no changes in LOG.
- With `-v http` you will see the _same_ information as now in LOG.
- With `-l http` you will see this _good_ change:
- - Insted of this:
```
[04/04 11:02:25.867 main]: reply -> 7 (192.168.1.41:57480) CL:1207 [sock_id 7]:
[04/04 11:02:25.867 main]: HTTP/1.0 200 OK
Date: Sat, Apr 4 09:02:25 2020 GMT
Server: minisatip/1.0.2-49290af
Cache-Control: no-cache
Content-type: text/xml
X-SATIP-RTSP-Port: 7554
Connection: close
Content-Length: 1207

<?xml version="1.0"?><root xmlns="urn:schemas-upnp-org:device-1-0" configId="1"><specVersion><major>1</major><minor>1</minor></specVersion><device><deviceType>urn:ses-com:device:SatIPServer:1</deviceType><friendlyName>minisatip</friendlyName><manufacturer>cata</manufacturer><manufacturerURL>http://github.com/catalinii/minisatip</manufacturerURL><modelDescription>minisatip for Linux</modelDescription><modelName>minisatip</modelName><modelNumber>1.1</modelNumber><modelURL></modelURL><serialNumber>1</serialNumber><UDN>uuid:11223344-9999-0000-b7ae-000c29364c1b</UDN><iconList><icon><mimetype>image/png</mimetype><width>48</width><height>48</height><depth>24</depth><url>/sm.png</url></icon><icon><mimetype>image/png</mimetype><width>120</width><height>120</height><depth>24</depth><url>/lr.png</url></icon><icon><mimetype>image/jpeg</mimetype><width>48</width><height>48</height><depth>24</depth><url>/sm.jpg</url></icon><icon><mimetype>image/jpeg</mimetype><width>120</width><height>120</height><depth>24</depth><url>/lr.jpg</url></icon></iconList><presentationURL>http://192.168.1.37:9981/</presentationURL>
<satip:X_SATIPCAP xmlns:satip="urn:ses-com:satip">DVBS2-1</satip:X_SATIPCAP></device></root>
[04/04 11:02:25.867 main]: minisatip.c:1916 get_sid returns NULL for s_id = -1
```
- - You will get this:
```
[04/04 11:02:25.867 main]: reply -> 7 (192.168.1.41:57480) CL:1207 [sock_id 7]:
[04/04 11:02:25.867 main]: minisatip.c:1916 get_sid returns NULL for s_id = -1
```
Note that you continue to see the `reply -> 7` line, then you get the info of the message is sendend using the HTTP protocol (however, except if you are trancing SSDP/HTTP/UPNP messages, then the content is completly irrelevant, and in any case debugging such messages you will enable the full log).

> Personally I do not think is a good idea.

Personally, the HTTP messages are completly irrelevant when debugging RTSP. So, I suggest to rethink your decision.

Regards. :wink:

ctoda:
Lars, you are right this affects only HTTP, but this creates inconsistency.
Consistency is more important than which decision you take: (-l http vs -v http).
Because the decision of doing something is no longer logical (why do you get for HTTP with -v but for RTSP with -l).

Generally I am ok with doing some changes, but those changes should be consistent across the entire codebase. That will make the behavior more predictable and easier to argue, this is a bug this is not.

catalinii:
Lars, you are right this affects only HTTP, but this creates inconsistency.
Consistency is more important than which decision you take: (-l http vs -v http).
Because the decision of doing something is no longer logical (why do you get for HTTP with -v but for RTSP with -l).

Generally I am ok with doing some changes, but those changes should be consistent across the entire codebase. That will make the behavior more predictable and easier to argue, this is a bug this is not.

lars18th:
Hi @catalinii ,

- First: I understand your position. And I agree with the idea of global consistency.
- Second: For years (more than two years ago), I'm requesting to improve the LOG. Mainly because it's __VERY__ annoying to see the content of the HTTP UI messages in the middle of the LOG when diagnosing RTSP commands.
- Third: I implemented a _simple_ and not intrusive solution.
- Four: I prefer to have the options `-l http` and `-l rtsp`, but the current implementation sharing the RTSP/HTTP module makes it impossible.

After saying that, please expose how to disable the HTTP message content when debugging RTSP commands with `-l http`.

I hope you can provide a good idea to do it. I'm empty of ideas. :cry: 
But thank you for this good SAT>IP server! :+1: 
Regards.

catalinii:
-l http sounds good vs -l rtsp.
See how -l http is done and add -l rtsp.

lars18th:
Hi @catalinii ,

> See how -l http is done and add -l rtsp.

OK. I'll do it. Now I understand how it works. However, I have one question: I will implement the `-l rtsp` as a subset of the `http` module? The problem is this: for receiving a request the RTSP uses a different function that HTTP. So here no problem. However, for sending a response only one function is used. How you prefer to handle this? One option is to enable some parts with both options, and other with only one. So in this case, for example, the same `LOGM()` can be triggered with `-l http` or `-l rtsp`. You agree with this approach?

catalinii:
You can define a variable holding the log value:
eg in read_rtsp:
int log_level = LOG_RTSP;

if(s->type = TYPE_HTTP)
  log_level = LOG_HTTP;

and then replace LOG(... with LOGL(log_level, ...)

lars18th:
Hi @ca

> You can define a variable holding the log value:
> eg in read_rtsp:
> int log_level = LOG_RTSP;
> 
> if(s->type = TYPE_HTTP)
> log_level = LOG_HTTP;
> 
> and then replace LOG(... with LOGL(log_level, ...)

All is implemented in #733 . So I close this PR.

Issue #655, opened 2019-11-27
BUG satipc frontend never expires streams


lars18th:
Hi @catalinii ,

This is related to #636, however it's a different problem: When a SAT>IP client doesn't sends the TEARDOWN to the minisatip process, and the frontend is a `satipc` tuner, then in this case the opened stream continues open _forever_. The target SAT>IP server continues receiving the keepalive OPTIONS command from minisatip.

So, I suspect that some expire whatdog in the `satipc` isn't working.
Please, can you review it?
Thank you!

catalinii:
Can you reproduce and send me the logs ? Also indicate the time you stop the last channel.

lars18th:
> Can you reproduce and send me the logs ? Also indicate the time you stop the last channel.

Hi @catalinii ,

Please reopen this issue, it is not resolved at all! As indicated in #636 the current _close_ process in the `satipc` module is incorrect. The TEARDOWN is _never_ sended _and_ as pointed in this issue the stream in the server continues open and no timeout closes it.

So, please check it. You can do it without any log, as you can reproduce it:
- Configure one minisatip server as a client of a SAT>IP server (it can be hardware or another minisatip).
- Use a client (RTP or HTTP) to receive any stream from minisatip.
- Enable your debug (tcpdump) for the RTSP stream between the minisatip and the SAT>IP server.
- Close (not channel change, so a true stop) the session from the client...

...And then you will see that the SAT>IP server doesn't receive the corresponding TEARDOWN command. Futhermore the session will continue opened until the SAT>IP _server_ will close it because his internal timeout. As the minisatip `satipc` never expires it.

I hope you'll want to fix this.
Regards.

lars18th:
Hi @catalinii ,

This bug is quite annoying! When the client connected to the minisatip closes the session, the satipc module continues without closing the tuner in the target SAT>IP server. The bug is clear: _the TEARDOWN command is never sended to the SAT>IP server when the adapter is closed_.

Please, can you fix this problem?
Thank you!

catalinii:
Lars, I tried to reproduce this issue:
printf 'PLAY rtsp://192.168.1.1:554/?src=1&fe=2&freq=11671&pol=h&msys=dvbs2&mtype=8psk&sr=23000&pids=0\r\nTransport: RTP/AVP/TCP;interleaved=0-1\r\n\r\n' | nc localhost 554

hit ctrl+c, but on my side I see the adapter is being closed.
Can you provide a repro ?

[...]

PR #728, opened 2020-04-04
Fix no TEARDOWN with satipc module 

lars18th:
This fixes the problem when a TEARDOWN is received from the SAT>IP client, and the tuner is a SAT>IP server. With this patch the SAT>IP server receives a TEARDOWN command when the tuner is released. Fixes the issue #655 .

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -837,7 +837,7 @@ void close_adapter_for_stream(int sid, int aid, int force)
 	mutex_unlock(&ad->mutex);
 	if (((ad->restart_needed == 1) || force) && !is_slave)
 	{
-		LOG("restarting adapter %d as needed", ad->id);
+		LOG("%s adapter %d as needed", force ? "closing" : "restarting", ad->id);
 		//		request_adapter_close (ad);
 		close_adapter(ad->id);
 	}
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -260,6 +260,7 @@ int init_hw(int i)
 	ad->id = i;
 	ad->fe_sock = -1;
 	ad->sock = -1;
+	ad->is_standby = 0;
 	ad->force_close = 0;
 	ad->restart_needed = 0;
 
@@ -789,7 +790,7 @@ int set_adapter_for_stream(int sid, int aid)
 	return 0;
 }
 
-void close_adapter_for_stream(int sid, int aid, int force)
+void close_adapter_for_stream(int sid, int aid, int stby)
 {
 	adapter *ad;
 	streams *s = get_sid(sid);
@@ -811,8 +812,8 @@ void close_adapter_for_stream(int sid, int aid, int force)
 	}
 	if (ad->sid_cnt > 0)
 		ad->sid_cnt--;
-	LOG("closed adapter %d for stream %d m:%d sid_cnt:%d, restart_needed %d, force_close %d", aid, sid, ad->master_sid,
-		ad->sid_cnt, ad->restart_needed, force);
+	LOG("closed adapter %d for stream %d m:%d sid_cnt:%d, restart_needed %d, try_standby %d", aid, sid, ad->master_sid,
+		ad->sid_cnt, ad->restart_needed, stby);
 	// delete the attached PIDs as well
 	if (ad->sid_cnt == 0)
 	{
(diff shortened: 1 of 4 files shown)


lars18th:
Hi @catalinii ,

I forget to mention that when multiple clients are using the same tuner, the behaviour with this patch is correct: the TEARDOWN is only sended to the remote SAT>IP server when the _last_ session using the tuner is closed. So the patch is correct and fully functional.

I hope you want to merge it.
Regards.

lars18th:
Hi @catalinii ,

Sorry for the first version with a typo error: it doesn't compile. Now it's fixed.
Please, review it. Now it's ready to merge it!
Regards.

lars18th:
Hi @catalinii ,

Finally I implemented a _clean_ implementation: actually only sends a TEARDOWN when it's required, and it doesn't really close the adapter. This improves the efficiency when re-starting to use the adapter, as the RTSP session continues open. Futhermore, the mode can be reused in the future with the flag `ad->is_standby` for other devices that can have the functionality of sleeping.

I feel in this last version you will agree with the patch.
Regards.

src/satipc.c
```diff
@@ -496,6 +496,12 @@ void satip_close_device(adapter *ad)
 	satipc *sip = get_satip(ad->id);
 	if (!sip)
 		return;
+	if (ad->is_standby)
+	{
+		LOG("satip device %s:%d going to standby", sip->sip, sip->sport);
+		http_request(ad, NULL, "TEARDOWN");
```
catalinii, line 502: This is in no way different that the previous code: you are just seding the TEARDOWN. Please note that you are not resetting the session id which is not good.

src/adapter.c
```diff
@@ -834,6 +835,13 @@ void close_adapter_for_stream(int sid, int aid)
 		mark_pids_deleted(aid, sid, NULL);
 	update_pids(aid);
 	adapter_update_threshold(ad);
+	if (stby && !is_slave && ad->type == ADAPTER_SATIP && ad->sid_cnt == 0)
+	{
+		LOG("adapter %d closing for standby", ad->id);
+		ad->is_standby = 1;
+		close_adapter(ad->id);
```
catalinii, line 842: Calling close adapter will close the entire adapter and close the socket.

src/adapter.c
```diff
@@ -789,7 +790,7 @@ int set_adapter_for_stream(int sid, int aid)
 	return 0;
 }
 
-void close_adapter_for_stream(int sid, int aid)
+void close_adapter_for_stream(int sid, int aid, int stby)
```
catalinii, line 793: replace stby, with close_stream everywhere

catalinii:
Probably what you want to do is, where there is no stream attached, send teardown but do not close the adapter. You can probably include a new member in https://github.com/catalinii/minisatip/blob/master/src/adapter.h#L109 called "unused"
Set ad->unused to a new function in satipc.c and then call that one from close_adapter if is not NULL

src/adapter.c
lars18th: > replace stby, with close_stream everywhere

OK.

src/satipc.c
lars18th: > This is in no way different that the previous code: you are just seding the TEARDOWN. Please note that you are not resetting the session id which is not good.

The objective is to just send the TEARDOWN. But, yes I don't have reseted the session id. I'll look for this. Any idea?

src/adapter.c
lars18th: > Calling close adapter will close the entire adapter and close the socket.

Here I don't accept it! Please, review your previous comment. If the `ad->is_standby` flag is set, then the call to `close_adapter()` doesn't close the adapter. It only sends the TEARDOWN. Please, note that this code is included in a test with `ad->type == ADAPTER_SATIP`. So only if the adapter can check the `is_standby` flag, then the function `close_adapter()` is called.

I'm sure this part of the code is correct. I tested and debugged it. No problem at all. The TEARDOWN command is sended and the adapter continues open, even the socket.

lars18th:
Hi @catalinii ,

> Probably what you want to do is, where there is no stream attached, send teardown but do not close the adapter.

Yes. That's the objective. And that's what I'm implemented. See the code, please!

> You can probably include a new member in https://github.com/catalinii/minisatip/blob/master/src/adapter.h#L109 called "unused"
> Set ad->unused to a new function in satipc.c and then call that one from close_adapter if is not NULL

My implementation used `ad->is_standby` as a flag for the same pourpose. When calling to `close_adapter()` this flag is checked, and **ONLY** if it's false then the close is done. So when it's true only a TEARDOWN is done. The difference between your suggestion and my suggestion, it's only that `ad->unused()` doesn't require any check (except the NULL function), and my approach requires one `ad->type` tuner type check. But the final behaviour is _identical_.

I can accept that your approach is more elegant. However, no difference after all. Really you want to rewrite it? If you prefer then I'll done it... :disappointed:

catalinii:
It is more abstract and allows other use cases at later time.

lars18th:
Hi @catalinii ,

I do all your suggestions. A new PR is provied: #735 . So I close this.

PR #730, opened 2020-04-07
WIP: SATIPC Fix delpids when closing

lars18th:
It resolves two problems:

1. Don't send irrelevant "delpids" when closing the adapter.
2. Don't send incorrect SETUP command when committing DELETE pids before the first connection of the SAT>IP client mode.

Supersedes PR #638 .

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -261,6 +261,7 @@ int init_hw(int i)
 	ad->fe_sock = -1;
 	ad->sock = -1;
 	ad->force_close = 0;
+	ad->is_closing = 0;
 	ad->restart_needed = 0;
 
 	if (opts.max_pids)
@@ -443,6 +444,7 @@ int close_adapter(int na)
 	//close all streams attached to this adapter
 	//	close_streams_for_adapter (na, -1);
 	mark_pids_deleted(na, -1, NULL);
+	ad->is_closing = 1;
 	update_pids(na);
 	//      if(ad->dmx>0)close(ad->dmx);
 	if (ad->fe > 0)
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -443,8 +443,8 @@ int close_adapter(int na)
 		ad->close(ad);
 	//close all streams attached to this adapter
 	//	close_streams_for_adapter (na, -1);
-	mark_pids_deleted(na, -1, NULL);
 	ad->is_closing = 1;
+	mark_pids_deleted(na, -1, NULL);
 	update_pids(na);
 	//      if(ad->dmx>0)close(ad->dmx);
 	if (ad->fe > 0)
--- a/src/adapter.h
+++ b/src/adapter.h
@@ -54,6 +54,7 @@ typedef struct struct_adapter
 	int sock, fe_sock;
 	int do_tune;
 	int force_close;
+	int is_closing;
 	unsigned char *buf; // 7 rtp packets = MAX_PACK, 7 frames / packet
 	int64_t rtime;
 	int64_t last_sort;
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -828,7 +828,7 @@ int satipc_del_filters(adapter *ad, int fd, int pid)
 	int i;
 	satipc *sip = get_satip(ad->id);
 	fd -= 100;
-	if (ad->is_closing) // Before a TEARDOWN command is redundant to send a "delpids" !
+	if (ad && ad->is_closing) // Before a TEARDOWN command is redundant to send a "delpids" !
 	{
 		LOGM("satipc: del_pid for pid %d not commited before a TEARDOWN", pid);
 		return 0;
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -828,6 +828,11 @@ int satipc_del_filters(adapter *ad, int fd, int pid)
 	int i;
 	satipc *sip = get_satip(ad->id);
 	fd -= 100;
+	if (ad->is_closing) // Before a TEARDOWN command is redundant to send a "delpids" !
+	{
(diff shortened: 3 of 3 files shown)


lars18th:
Hi @catalinii ,

I have implemented all your suggestions. I feel you will be satisfied. So I hope you merge it soon.

Regards.

src/adapter.c
```diff
@@ -261,6 +261,7 @@ int init_hw(int i)
 	ad->fe_sock = -1;
 	ad->sock = -1;
 	ad->force_close = 0;
+	ad->is_closing = 0;
```
ctoda, line 264: Move this to satipc.c

src/adapter.h
```diff
@@ -54,6 +54,7 @@ typedef struct struct_adapter
 	int sock, fe_sock;
 	int do_tune;
 	int force_close;
+	int is_closing;
```
ctoda, line 57: Move this to satipc.c

src/adapter.c
```diff
@@ -442,6 +443,7 @@ int close_adapter(int na)
 		ad->close(ad);
 	//close all streams attached to this adapter
 	//	close_streams_for_adapter (na, -1);
+	ad->is_closing = 1;
```
ctoda, line 446: Move this to satipc_close

ctoda:
As workflow, please do the following when submitting a PR:

0) git checkout -b my_new_branch
1) git commit -a -m "new PR"
2) git push
3) New changes
4) git commit -a --amend
5) git push -f

One this is addressed and the comments I will merge

catalinii:
The point is to have 1 commit per PR

catalinii:
The point is to have 1 commit per PR

src/adapter.c
lars18th: > Move this to satipc.c

Sorry? `is_closing` is an adapter member and requires to be INITIALIZED!! Here is the initialization of the adapter. Why you say to set _only_ in the `satipc` module?

src/adapter.c
lars18th: > Move this to satipc_close

Sorry? The function `close_adapter` is called _before_ any other function inside the satipc module. So it's impossible to set the `is_closing` flag before this point. And it's required to set it here as the next lines change the pids, and inside the satipc handling of the pids I need to know if the adapter is closing.

src/adapter.h
lars18th: > Move this to satipc.c

Again, why that?

catalinii:
@lars18th, my point was that you can handle everything from satipc.c. You do not need to add the member in adapter.h but you can do it in satipc.c.

This change is specific to satipc.c module and should be contained there.

lars18th:
Hi @catalinii ,

> @lars18th, my point was that you can handle everything from satipc.c. You do not need to add the member in adapter.h but you can do it in satipc.c.

How? I don't know how to handle the `ad->is_closing` flag in the `satipc` module if the flag is not set _inside_ the `adapter` module. Any idea?

Regards.

Note: `(@catalinii == @ctoda) ? `

catalinii:
Yes both are my accounts...
You can add a new member to the structure: https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L58
and use it just in satipc.c

lars18th:
> Yes both are my accounts...

:+1: Thanks for letting us know!

> You can add a new member to the structure: https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L58
> and use it just in satipc.c

Sorry? The code works _inside_ the `adapter` module. I commented it before. The function to catch is `close_adapter()`. So I **can't** _just_ use it in `satip.c`.

IMHO from the point of view of a _clean_ code, when two modules require to share information you can use different alternatives. Hovewer, it's more _clean_ to use a flag inside `adapter` that a flag inside `satipc`. Why? Because then the function `close_adapter()` will require a code checking the adapter TYPE. And this code will be ugly, as when adding similar functionality to other adapter types then you require to change the generic code. But when setting this flag all the code is agnostic and only the adapters that require it will check it _inside_ their code.

Sorry, but I don't understand your point of view. So please can you explain in more detail?

catalinii:
Is this still needed after merging https://github.com/catalinii/minisatip/pull/735 ?

lars18th:
Hi @catalinii ,

> Is this still needed after merging #735 ?

Yes and No. Let me to explain it: This PR is for fixing troubles when DELETING PIDS. So it's not related to the TEARDOWN trouble fixed in #735. However, I'm doing more tests with my other already not published patches for improving the deleting of pids (based on a delayed behaviour in different cases: for the client, for the satipc module, etc.). So, I suggest to maintain open this PR and not close it for a while. Perhaps I'll provide another one that will supersedes this. But no sure.

However, in the maintime I suggest to target this trouble before: #731 . As I'm working on fixing other issues, and my understanding of the timers in minisatip is limited... please can you fix that problem? I'm sure your changes can serve as an example for me to fix these other problems. :wink: 

You agree?

lars18th:
Hi @catalinii ,

I'm working on another PR based on this that targets to deferred delete pids. However, it requires a lot of more testing/validation. It's based on two versions (both complementary): SAT>IP client deferred deletes, and SAT>IP server (satipc) deferred pid deletion. More or less the idea is to support something like `minisatip -W *,2`: where `*` indicates "for clients", and the number is the adapter to apply it. However, for adapters it will support only SAT>IP devices. With the functionality enabled the concept is simple: smart deferred delete of a pid. For clients it implies to "block" deletes to the adapter, and for adapters the "block" is when pasing it to the tuner (in this case only for SAT>IP devices).

You agree with that approach?

lars18th:
I'll close this PR (for cleanup), but I need feedback.

Jalle19:

PR #735, opened 2020-04-09
SATIPC/GENERAL: New sleep tuner mode (aka "Fix no TEARDOWN with satipc module v2")

lars18th:
This patch implements a new optional adapter callback function named standby(). This function is called when the adapter doesn't have any stream clients running. This function is previous to the close() adapter. The expected behaviour is to put the tuner in a standby sleep for a fast restart in the future.

This new functionality is used in the satipc module. With the new functionality when the last client sends a TEARDOWN command, the module sends a TEARDOWN to the remote SAT>IP server, but it doesn't close the connection. Then any new tuning command from a SAT>IP client can be served instantly. In any case, when the timeout of the adapter is reached, then the adapter is closed definitively.

This PR supersedes #728.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -68,6 +68,7 @@ typedef struct struct_satipc
 	int sport;
 	char session[18];
 	int stream_id;
+	int sleep;
 	int listen_rtp;
 	int rtcp, rtcp_sock, cseq;
 	int wp, qp;			 // written packet, queued packet
@@ -463,7 +464,7 @@ int satipc_open_device(adapter *ad)
 	sip->expect_reply = 0;
 	sip->last_connect = 0;
 	sip->sent_transport = 0;
-	sip->session[0] = 0;
+	sip->sleep = 0;
 	sip->stream_id = -1;
 	sip->wp = sip->qp = sip->want_commit = 0;
 	sip->rcvp = sip->repno = 0;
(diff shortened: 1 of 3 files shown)


lars18th:
Hi @catalinii ,

This version solves all the problems! So it fixes the problem described in the issue #655 . But it goes further and implements a new _optional_ adapter status: "standby". In this state a fast retune is supported. I feel you can use it for other tuner devices.

Please, check it. I'm sure this time you'll want to merge it.
Regards.

src/satipc.c
```diff
@@ -463,7 +464,7 @@ int satipc_open_device(adapter *ad)
 	sip->expect_reply = 0;
 	sip->last_connect = 0;
 	sip->sent_transport = 0;
-	sip->session[0] = 0;
+	sip->sleep = 0;
```
lars18th, line 467: This is not an error!
The original code is repeated in the same block:
https://github.com/catalinii/minisatip/blob/681f099c8dd6494e24db65129e5ee44d3b6bc96f/src/satipc.c#L457
https://github.com/catalinii/minisatip/blob/681f099c8dd6494e24db65129e5ee44d3b6bc96f/src/satipc.c#L466

catalinii:
This part looks good, but there is nothing in adapter.c calling ad->standby (that part was implemented correctly in the previous patch).

catalinii:
Again, please see a previous PR in which I commented how to add just one commit instead of 2.

lars18th:
Hi @catalinii ,

> This part looks good, but there is nothing in adapter.c calling ad->standby (that part was implemented correctly in the previous patch).

I can't understand this. Please see:
```
		ad->master_sid = -1;
		mark_pids_deleted(aid, -1, NULL);
		init_dvb_parameters(&ad->tp);
+		if (ad->standby)
+			ad->standby(ad);
#ifdef AXE
		free_axe_input(ad);
#endif
```
That's the first change in the PR about the `adapter.c` file.

Are you reading and testing the PRs?

lars18th:
> Again, please see a previous PR in which I commented how to add just one commit instead of 2.

Please, note that I'm testing all my PRs before commiting them. Only when I see a _simple_ and _stupid_ error/bug because a typo or similar I push a second fix. Please, see that the second commit is in this case. This PR is created from my develop tree with around 10 commits. The PR is squished before plublish it. You really want to create a new PR for just a simple typo?

catalinii:
Sorry I missed that.

PR #736, opened 2020-04-10
ADAPTER: Limit adapters to specific sources (aka "Assign adapters to specific sources v3")

lars18th:
This patch adds a new parameter (-U) to limit specific sources to some adapters.

It supersedes previous catalinii#623 with these changes:

- More robust functionality.
- Increased total number of sources to 64.
- The list of the sources parameter can be less than the MAX_SOURCES.
- Improved debug to identify troubles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,14 @@ Help
 
 * -j --jess jess_string - same format as -u 
 
+* -U --sources sources_for_adapters: limit the adapters to specific sources/positions
+	* eg: -U 0-2:*:3:2,6,8 (no spaces between parameters)
+	- In this example: for SRC=1 only 0,1,2; for SRC=2 all; for SRC=3 only 3; and for SRC=4 the 2,6,8 adapters are used.
+	- For each position (separated by : ) you need to declare all the adapters that use this position with no exception.
+	- The special char * indicates all adapters for this position.
+	- The number of sources range from 1 to 64; but the list can include less than 64 (in this case all are enabled for undefined sources). 
+	- By default or in case of errors all adapters have enabled all positions.
+
 * -w --http-host http_server[:port]: specify the host and the port (if not 80) where the xml file can be downloaded from [default: default_local_ip_address:8080] 
 	* eg: -w 192.168.1.1:8080 
 
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -96,6 +96,15 @@ adapter *adapter_alloc()
 	ad->master_source = -1;
 	ad->diseqc_multi = opts.diseqc_multi;
 
+	/* enable all sources */
+	ad->debug_pos[0]='\0';
+	ad->debug_src = 0;
+	int i;
+	for (i = 0; i <= MAX_SOURCES; i++)
+	{
+		ad->sources_pos[i] = 1;
+	}
+
 	/* LOF setup */
 	ad->diseqc_param.lnb_low = opts.lnb_low;
 	ad->diseqc_param.lnb_high = opts.lnb_high;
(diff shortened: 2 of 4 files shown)


lars18th:
Hi @catalinii ,

I prepared an updated version to overcome the rollback of my previous version.
I'm sure this version will work perfectly, however please test it for a while until you merge it.

Some notes about this new version:
- Now when no using the parameter "-U" the users don't be disturbed. I check it a lot to overcome troubles with users that will don't use this functionality.
- You can "full" debug the _parsing_ and the _assignement_ of the sources enabling `-v adapter`. Please, try different configurations and see the values.
- The total number of 64 sources is inase. However, I prefer to not limit to users. Now if you use a short list of 4 sources (ie. `-U 1:2:3:4`) you only modify the first four positions, others will be assumed to set for all. This simplifies the code and doesn't disturb.
- You can check the parsing with examples like this:
- - `minisatip -f -e 0,1,2,3 -s 127.0.0.1:6554 -s 127.0.0.1:7554 -s 127.0.0.1:8554 -s 127.0.0.1:9554 -v adapter -U 0:1:2:3:*:0,1:2,3:2-3:0,2-3:3,0:10:11:12:13:14:15:16:17:18:19:20:21:22:23:24:25:26:27:28:29:30:31:*:1:2:3:4:5:6:7:8:9:1-15,20-30:1,4,8,12,16:10,20,30:1,1,1,1,1,1,1,1,1,1:7:9:13`

I hope you want to merge it soon.
Regards.

catalinii:
Again, please see a previous PR in which I commented how to add just one commit instead of 2.

lars18th:
Hi @catalinii ,

> Again, please see a previous PR in which I commented how to add just one commit instead of 2.

I repeat: A full new PR for a simple typo? Please, check the reason for the second commit.

> Please test the previous case that failed

I tested _all_ cases that I know. The patch works perfectly for me for serveral days testing it. I suggest to test it for other users just in case... but I feel that _all_ possible cases are considered. You have some info about a case failed with _this_ PR?

catalinii:
@lars18th I think you may be the only one testing with the options enabled. Most of the users try to get normal functionality (without the option open).

lars18th:
Hi @catalinii ,

> @lars18th I think you may be the only one testing with the options enabled. Most of the users try to get normal functionality (without the option open).

For this reason I suggest to **try it**.
You tested it?

catalinii:
No I did not yet

catalinii:
Generally the responsibility for testing foe testing is of the submitter

lars18th:
> Generally the responsibility for testing foe testing is of the submitter

Sure! And I'm testing it for almost two weeks without any trouble. However, from the experience with my first version of this patch, some users can run in a very different environment. I'm trying to target any use case, and the code is very robust. But febore a commit (remember the rollback for the previous version) it's preferable a deep test. For this reason I'm requesting this... and not because I'm not doing a full test. In my enviroment this patch is fully tested!

lars18th:
Hi @catalinii ,

Thank you for merging this "large" patch (in the sense of _new_ functionality, and not a patch fix)!
And even I tested it for weeks, I want to receive feedback from users _that are using it_. If the users don't use the `-U` option then all will continue working as before, so no problem. However, when activating it perhaps some cases requires to provide some enhancement.

For example, this use case is not targeted in the current code (as I don't have this environment, and I don't have any clear solution to it):
- Two wires connected to two different Diseqc 4x1 switches, connected to 4 different antennas each with 2 tuners with 1 input each. In this case, each wire uses SRC=1,2,3,4 as the idenfiers. For SAT>IP clients using minisatip actually the sources range from 1 to 8. Then the `-U` parameter can be used to restrict the first tuner to SRC=1-4 and the second to SRC=5-8. But, what I'm not targeting in the source code is any source translation from adapter to tuner. In this case can be expected that the sources of the second tuner will be translated from 5-8 to 1-4, in order to manage the Diseqc switch correctly. But at time this is not the case.

Even I think that some cases like this requires some enhancement, at time I'm empty of ideas about how to implement it. So I request for feedback. Any ideas?

Regards.

PR #737, opened 2020-04-15
SATIPC: Check to don't send a TEARDOWN without stream parameter

lars18th:
In some very specific cases, for example when the SAT>IP server doesn't send any RTP data, the device can be closed without a valid stream id. In this case, it's not required to send the TEARDOWN message. Futhermore, some SAT>IP servers have toubles when receiving a TEARDOWN message without the stream parameter, what actually is a break of the specifications (the TEARDOWN command **requires** the "/stream=" part in the URI).

This patch fixes this problem.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -1040,6 +1040,11 @@ int http_request(adapter *ad, char *url, char *method)
 
 	if (sip->stream_id != -1)
 		sprintf(sid, "stream=%d", sip->stream_id);
+	else if (!strcmp(method, "TEARDOWN"))
+	{
+		LOG("satipc_http_request (adapter %d): impossible to send TEARDOWN without specific stream!", ad->id);
+		return 0;
+	}
 
 	lb = snprintf(buf, sizeof(buf), format, method, sip->sip, sip->sport, sid,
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

I detected this problem after the merge f864e17ee5979f620364922c08a1324c3a0f5e25 . However, they are not related. This PR fixes **all** cases when a TEARDOWN is sended without the "/stream=" parameter. That message in this way is not in compliance with the specifications. As a consequence some faulty SAT>IP servers (e.g. XORO STBs) crash the SAT>IP server process when they receive TEARDOWN messages without indicating the stream.

I hope you merge this patch soon to circunvent this problem with such SAT>IP servers.

lars18th:
Hi @catalinii ,

This patch is quite simple, and fixes a **real** problem. In my tests with a SAT>IP server that closes the RTSP socket (instead of waiting for closing it from the client side), the close of the adapter can be done at very moment. So in this case, the final TEARDOWN can be called _without_ any stream.  This breaks compatibility with the standard, and the command should never be sent.

So, please commit this patch.

lars18th:
Great! :+1:

PR #739, opened 2020-04-16
WEBIF: Provide VLC stream URL as LINK

lars18th:
With this patch you can copy&paste the URL links directly to VLC or any other tool that can handle the MPEG2-TS streaming.


lars18th:
Thank you @catalinii for merging it! :+1:

PR #741, opened 2020-04-21
Discard non-TS payload in RTP packets

lars18th:
Some SAT>IP servers (ex. XORO) generates non-default KEEP-ALIVE RTP packets.
These packets doesn't contain any TS data in the payload, and the size is less than 188bytes.
Therefore, these packages must be discarded to maintain alignment with the actual TS packages.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -566,6 +566,18 @@ int satipc_read(int socket, void *buf, int len, sockets *ss, int *rb)
 	}
 	if (!ad)
 		ad = get_adapter(ss->sid);
+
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

This patch enables compatibility with some SAT>IP servers that send "invalid" RTP packets (in the sense of packets with no payload or with a payload that is not a TS packet). The solution is then discard these packets. This doesn't create any side effect and the new code works perfectly with such servers and regular servers. So it increases the compatibility of minisatip with more SAT>IP servers.

Please, merge it! :wink:

PR #733, opened 2020-04-09
LOG: New RTSP log module (aka "Improve logs of the HTTP module v2")

lars18th:
The RTSP log is separated from the HTTP log.

Supersedes #726 .

PR #744, opened 2020-04-24
LOG improvement v3 (aka "Improve logs of the HTTP module v3")

lars18th:
This patch improves the LOG part in some areas:

- Fixes a bug with some module names ("socketworks" can't be selected because is a substring of "socket").
- Fixes the unkown IP of the SAT>IP servers.
- Prints the name the of the module in each debug line.
- Prints the enabled module debug level of selected modules.
- Disambiguation of HTTP/RTSP protocols.
- Enhacement of some socket messages with incorrect values.
- Other minor changes.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -57,7 +57,7 @@
 
 struct struct_opts opts;
 
-#define DEFAULT_LOG LOG_HTTP
+#define DEFAULT_LOG LOG_GENERAL
 
 #define DESC_XML "desc.xml"
 #define PID_NAME "/var/run/%s.pid"
@@ -1080,6 +1080,9 @@ void set_options(int argc, char *argv[])
 		set_slave_adapters("2-7:0");
 }
 
+#undef DEFAULT_LOG
+#define DEFAULT_LOG LOG_RTSP
+
 #define RBUF 32000
 
 int read_rtsp(sockets *s)
@@ -1088,9 +1091,13 @@ int read_rtsp(sockets *s)
 	int cseq, la, i, rlen;
 	char *transport = NULL, *useragent = NULL;
 	int sess_id = 0;
+	int log_level = LOG_RTSP;
 	char buf[2000];
 	char ra[50];
 	streams *sid = get_sid(s->sid);
+
+	if (s->type == TYPE_HTTP)
+		log_level = LOG_HTTP;
 
 	if (s->rlen > 3 && s->buf[0] == 0x24 && s->buf[1] < 2)
 	{
@@ -1098,9 +1105,8 @@ int read_rtsp(sockets *s)
 			sid->rtime = s->rtime;
 
 		int rtsp_len = s->buf[2] * 256 + s->buf[3];
-		LOG(
-			"Received RTSP over tcp packet (sock_id %d) sid %d, rlen %d, packet len: %d, type %02X %02X discarding %s...",
-			s->id, s->sid, s->rlen, rtsp_len, s->buf[4], s->buf[5],
+		LOGG(log_level, "Received %s over tcp packet (sock_id %d) sid %d, rlen %d, packet len: %d, type %02X %02X discarding %s...",
+			s->type == TYPE_HTTP ? "HTTP" : "RTSP" , s->id, s->sid, s->rlen, rtsp_len, s->buf[4], s->buf[5],
 			(s->rlen == rtsp_len + 4) ? "complete" : "fragment");
 		if (s->rlen >= rtsp_len + 4)
 		{ // we did not receive the entire packet
@@ -1123,13 +1129,11 @@ int read_rtsp(sockets *s)
 	{
 		if (s->rlen > RBUF - 10)
 		{
-			LOG(
-				"Discarding %d bytes from the socket buffer, request > %d, consider increasing  RBUF",
+			LOG("read_rtsp: Discarding %d bytes from the socket buffer, request > %d, consider increasing  RBUF",
 				s->rlen, RBUF);
 			s->rlen = 0;
 		}
-		LOG(
-			"read_rtsp: read %d bytes from handle %d, sock_id %d, flags %d not ending with \\r\\n\\r\\n",
+		LOG("read_rtsp: read %d bytes from handle %d, sock_id %d, flags %d not ending with \\r\\n\\r\\n",
 			s->rlen, s->sock, s->id, s->flags);
 		hexdump("read_rtsp: ", s->buf, s->rlen);
 		if (s->flags & 1)
@@ -1144,10 +1148,10 @@ int read_rtsp(sockets *s)
 	rlen = s->rlen;
 	s->rlen = 0;
 
-	LOG("Read RTSP (handle %d) [%s:%d] sid %d, len: %d, sock %d", s->sid,
-		get_sockaddr_host(s->sa, ra, sizeof(ra)), get_sockaddr_port(s->sa),
-		s->id, rlen, s->sock);
-	LOGM("MSG client >> process :\n%s", s->buf);
+	LOGG(log_level, "Read %s (sid %d) [%s:%d] len: %d sock %d (handle %d)", s->type == TYPE_HTTP ? "HTTP" : "RTSP" ,
+		s->sid, get_sockaddr_host(s->sa, ra, sizeof(ra)), get_sockaddr_port(s->sa),
+		rlen, s->id, s->sock);
+	LOGL(log_level, "MSG client >> process :\n%s", s->buf);
 
 	if ((s->type != TYPE_HTTP) && (strncasecmp((const char *)s->buf, "GET", 3) == 0))
 	{
@@ -1182,7 +1186,7 @@ int read_rtsp(sockets *s)
 	sid = get_sid(s->sid);
 	if (sid)
 	{
-		LOGM("Updating sid %d time to %jd, current time %jd", sid->sid, s->rtime, getTick());
+		LOGL(log_level, "Updating sid %d time to %jd, current time %jd", sid->sid, s->rtime, getTick());
 		sid->rtime = s->rtime;
 	}
 
@@ -1328,6 +1332,9 @@ int read_rtsp(sockets *s)
 	return 0;
 }
 
+#undef DEFAULT_LOG
+#define DEFAULT_LOG LOG_HTTP
+
 #define REPLY_AND_RETURN(c)                    \
 	{                                          \
 		http_response(s, c, NULL, NULL, 0, 0); \
@@ -1346,6 +1353,7 @@ int read_http(sockets *s)
 	char buf[2000]; // the XML should not be larger than 1400 as it will create problems
 	char url[300];
 	char ra[50];
+	int log_level = LOG_RTSP;
 	char *space;
 	int is_head = 0;
 	static char *xml =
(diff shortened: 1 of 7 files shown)


lars18th:
Hi @catalinii ,

This new PR supersedes #733 (and #726 too).
All is in one commit, and the improvements and fixes are relevant.
Please check it and merge when you can.
Regards.

lars18th:
Hi @catalinii ,

Any comment/idea with this PR? From my point of view it's ready to merge it.

src/utils.c
```diff
@@ -351,7 +351,7 @@ int map_intd(char *s, char **v, int dv)
 		return atoi(s);
 	}
 	for (i = 0; v[i]; i++)
-		if (!strncasecmp(s, v[i], strlen(v[i])))
+		if (strlen(s) == strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
```
catalinii, line 354: Was this tested? I am pretty sure this is not right.

src/utils.h
```diff
@@ -177,17 +177,22 @@ static inline int get_index_hash(void *p, int max, int struct_size, uint32_t key
...
 	{                                                   \
 		if ((level)&opts.log)                           \
-			_log(__FILE__, __LINE__, a, ##__VA_ARGS__); \
+			_log(__FILE__, __LINE__, level, a, ##__VA_ARGS__); \
+	}
+#define LOG(a, ...)                                     \
+	{                                                   \
+			_log(__FILE__, __LINE__, DEFAULT_LOG, a, ##__VA_ARGS__); \
+	}
+#define LOGG(level, a, ...)                                     \
```
catalinii, line 186: There is no need for LOGG, there is LOGL doing the same thing

src/minisatip.c
```diff
@@ -1435,10 +1444,10 @@ int read_http(sockets *s)
 	int secs = seconds;
 	sprintf(opts.time_running, "%.0d%s%02d:%02d:%02d", days, days > 0 ? "d " : "", hours, mins, secs);
```
catalinii, line 1446: Use LOGL

src/minisatip.c
```diff
@@ -1798,20 +1828,23 @@ int main(int argc, char *argv[])
...
+		si1 = sockets_add(ssdp1, &sv, -1, TYPE_UDP, (socket_action)ssdp_reply,
 						  (socket_action)ssdp_byebye, (socket_action)ssdp_discovery);
 		if (si < 0 || si1 < 0)
 			FAIL("sockets_add failed for ssdp");
 	}
 
 	sockets_timeout(si, 60 * 1000);
 	set_sockets_rtime(si, -60 * 1000);
-	if (0 > sockets_add(rtsp, NULL, -1, TYPE_SERVER, (socket_action)new_rtsp,
+	fill_sockaddr(&sv, "0.0.0.0", opts.rtsp_port, 1);
```
catalinii, line 1842: This is not logging related and breaks compatibility with IPv6 lets keep it separate from this PR

src/satipc.c
```diff
@@ -401,13 +401,14 @@ int satipc_open_device(adapter *ad)
...
+	USockAddr sv;
 	sip->last_connect = ctime;
-	ad->fe = tcp_connect_src(sip->sip, sip->sport, NULL, 1, sip->source_ip[0] ? sip->source_ip : NULL); // blocking socket
+	ad->fe = tcp_connect_src(sip->sip, sip->sport, &sv, 1, sip->source_ip[0] ? sip->source_ip : NULL); // blocking socket
 	if (ad->fe < 0)
 		LOG_AND_RETURN(2, "could not connect to %s:%d", sip->sip, sip->sport);
 
-	LOG("satipc: connected to SAT>IP server %s port %d %s handle %d %s %s", sip->sip,
-		sip->sport, sip->use_tcp ? "[RTSP OVER TCP]" : "", ad->fe, sip->source_ip[0] ? "from source" : "", sip->source_ip[0] ? sip->source_ip : "");
+	LOG("satipc: connected to SAT>IP server %s port %d%s (socket handle %d) %s %s", sip->sip,
```
catalinii, line 410: Do jot add socket. There is sockets (struct socket in socketworks) and handle (unix handle). This will be confusing

src/satipc.c
```diff
@@ -401,13 +401,14 @@ int satipc_open_device(adapter *ad)
 	if ((sip->last_connect > 0) && (ctime - sip->last_connect < 5000))
 		return 3;
 
+	USockAddr sv;
```
catalinii, line 404: What is the problem you are teying to solve by adding SV?

src/socketworks.c
```diff
@@ -652,8 +655,11 @@ int sockets_del(int sock)
 	ss->is_enabled = 0;
 	so = ss->sock;
 	ss->sock = -1; // avoid infinite loop
-	LOG("sockets_del: sock %d -> handle %d, sid %d, overflow bytes %d, allocated %d, used %d, unsend packets %d",
-		sock, so, ss->sid, ss->overflow, ss->buf_alloc, ss->buf_used, (ss->wmax > 0) ? ((ss->wpos - ss->spos) % ss->wmax) : 0);
+	LOG("sockets_del: deleting socket sock %d (handle %d) [%s:%d], sid %d, overflow bytes %d, allocated %d, used %d, unsend packets %d",
```
catalinii, line 658: Changing wording makes grepping more difficult especially in the beginning. Suggesting to keep the same.

src/satipc.c
```diff
@@ -421,12 +422,12 @@ int satipc_open_device(adapter *ad)
...
 			LOG("Could not listen on port %d: err %d: %s", sip->listen_rtp + 1, errno, strerror(errno));
 
-		ad->fe_sock = sockets_add(ad->fe, NULL, ad->id, TYPE_TCP,
+		ad->fe_sock = sockets_add(ad->fe, &sv, ad->id, TYPE_TCP,
 								  (socket_action)satipc_reply, (socket_action)satipc_close,
 								  (socket_action)satipc_timeout);
 		sip->rtcp_sock = -1;
 		if (sip->rtcp >= 0)
-			sip->rtcp_sock = sockets_add(sip->rtcp, NULL, ad->id, TYPE_TCP,
+			sip->rtcp_sock = sockets_add(sip->rtcp, &sv, ad->id, TYPE_TCP,
```
catalinii, line 430: You cannot add the destination of a socket to another socket

src/satipc.c
```diff
@@ -451,7 +452,7 @@ int satipc_open_device(adapter *ad)
 	{
 		ad->dvr = ad->fe;
 		ad->fe = -1;
-		ad->fe_sock = sockets_add(SOCK_TIMEOUT, NULL, ad->id, TYPE_UDP,
+		ad->fe_sock = sockets_add(SOCK_TIMEOUT, &sv, ad->id, TYPE_UDP,
```
catalinii, line 455: Same here

lars18th:
Hi @catalinii ,

I don't have sufficient time now to implement a new version. However, the _current_ log implementation I feel can be improved. And for this reason I created this patch time ago. Please, review my comments.

Regards.

src/utils.c
lars18th: This is absolutly correct, and it works! I'm using this patch for long time.

Without it, an horrendous bug targets the current implementation: when you want to use `-l socketworks` what is enabled is `socket` because "socketworks" starts with "socket".
Please, try it!

src/utils.h
lars18th: This is necessary because `LOGL(...)` prints only if the module has activated the log functionality, and the `LOGG(...)` prints every time. This modification of two different functions is necessary to support the _inconditional_ log of a module even if the log of the module is not enabled. Take note that `LOG(...)` and `LOG0(...)` do not print the name of the module.

src/minisatip.c
lars18th: This is not my code!

src/minisatip.c
lars18th: Hi,

This is necessary because in the current implementation not every time the full info "address"+"socket" is printed for _any_ socket connection. Then to make the code more readable and standard I use this simple approach: all listening sockets have the address too. And the implementation works, almost for IPv4. Any idea for IPv4/6 support?

src/satipc.c
lars18th: Hi,

To be true, sometimes I don't fully understand the original messages. However, I reviewed all the code and I've used the same names every time:

- `Handle` is the sock number (unix handle).
- `Socket` is the minisatip socket id (struct socket).
- `sid` is the sid of the connection.

I'm using _consistent_ names in the entire patch. However, in this case I'm not sure if `ad->fe` represents "socket" or "handle". What you prefer?

src/satipc.c
lars18th: The problem of the SAT>IP "client" module not printing the address of the server. Please, note that here the `sv` is used in the next `tcp_connect_src()` call. Passing one empty address will disable the option of logging messages with the full addresses info: local+port and remote+port.

src/socketworks.c
lars18th: That's is one of the more frustrating log lines. The current implementation is not clear. We need to now very clearly the ADDRESS:PORT of the connection, plus the full "Socket" and "Handle". So in this case the change of the wording it's necessary to improve the usability of the log.

src/satipc.c
lars18th: Why not? Without this change the ADDRESS of the remote SAT>IP server is not printed in the log. With this simple change the address is printed, that's more easy to understand.

src/satipc.c
lars18th: Yes, same here. Required to print the address of the remote SAT>IP server.

lars18th:
Hi @catalinii ,

After including the answers to your comments, I feel that you don't have tested the patch. Please, try it! I dedicated much time to clean/improve the log. This patch (the third iteration) is very complete. I'm sure is not perfect, but it represents a huge improvement in usability. Please, compile and test it. :wink: 

Regards.

src/utils.h
catalinii: But that means that you are printing the message regardless if http was enabled or no, right ?  If you want to print it in the http log level (or rtsp) feel free to use LOGL.

src/socketworks.c
catalinii: Adding more info is always helpful, but the wording is important for checking issues. I really doubt the users read those message and most likely will open an issue in the forum/github. that's why is better to make the job of finding issues a higher priority.

src/satipc.c
catalinii: I understand you are adding it because of logging, but basically you are binding multiple sockets to the same destination, which makes no sense and could lead to hidden bugs.

src/satipc.c
catalinii: Please see the next message why adding one destination to multiple sockets is a bad idea.

src/minisatip.c
catalinii: Can you point out where the printing happens is not right (eg NULL is printed) ?
You can use NULL instead of "0.0.0.0"

src/satipc.c
catalinii: ad->fe is none of them, we can leave it as it is... or add ad->fe %d

catalinii:
Hey @lars18th , added some more comments. Can you please answer them ?

Jalle19:
Seems like parts of this have already been merged in other PRs, what should we do about this?

lars18th:
> Seems like parts of this have already been merged in other PRs, what should we do about this?

Well, some are merged and some aren't. So then it will be reasonable to generate a v4 of the PR.

Jalle19:

PR #745, opened 2020-04-27
Filter out only invalid TS packets with payload

lars18th:
When the size of the RTP payload is zero, then pass it as regular packet.


lars18th:
Hi @catalinii ,

This is only to reduce the possibility of a problem if a zero size packet is received. In this case (when RTP payload is zero) instead of execute the new checking, the control is passed to the upper layer. Then the regular code will be executed.

I hope you want to merge it.
Regards.

PR #746, opened 2020-04-27
Fix erroneous source_map index

lars18th:
The j index starts from 0, but the SRC 0 can't be changed. Futhermore, the source_map[-1] doesn't exist.


lars18th:
Hi @catalinii ,

Please, merge this asap. The original code has a small bug!

Issue #636, opened 2019-11-12
SATIP client mode: No TEARDOWN because the socket is closed before


lars18th:
Hi @catalinii ,

I found another trouble with the SAT>IP client mode: When some SAT>IP client request a `pids=none` for one stream, then the `satipc` module never closes the remote session when the client doesn't sends a TEARDOWN command. So the session with the target SAT>IP server continue open forever.

Please, can you enforce to close (send a TEARDOWN message) if the client session is expired?
And, as a tip, I suggest to close too the session if `pids=none` and the client socket (from the SAT>IP client to minisatip) is closed. In this specific case it will be desirable to not wait until the watchdog to close the session.

I hope you want to consider it.
Thank you!

lars18th:
Hi @catalinii ,

I discovered the cause of this problem: The `satipc` module when the adapter is closed, then it closes the socket _before_ the TEARDOWN command is acknowledged.

See this part of the log:
```
[12/11 18:02:08.701 main]: closing adapter 2  -> fe:14 dvr:15, sock:17, fe_sock:15
[12/11 18:02:08.701 main]: satip device 192.168.1.27:554 is closing
[12/11 18:02:08.701 main]: satipc_http_request (ad 2): sending to sock 15:
TEARDOWN rtsp://192.168.1.27:554/stream=2 RTSP/1.0
CSeq: 12
Session: 0272793942


[12/11 18:02:08.701 main]: satip_close called for adapter 2, socket_id 15, handle 14 timeout 30000
[12/11 18:02:08.701 main]: sockets_del: sock 15 -> handle 14, sid 2, overflow bytes 0, allocated 0, used 0, unsend packets 0
[12/11 18:02:08.701 main]: sockets_del: 15 Last open socket is at index 17 current_handle 14
[12/11 18:02:08.701 main]: Destroying mutex 0x7fbd49ea0268
[12/11 18:02:08.701 main]: satip_close called for adapter 2, socket_id 16, handle 16 timeout 0
[12/11 18:02:08.701 main]: sockets_del: sock 16 -> handle 16, sid 2, overflow bytes 0, allocated 0, used 0, unsend packets 0
[12/11 18:02:08.701 main]: sockets_del: 16 Last open socket is at index 17 current_handle 16
[12/11 18:02:08.701 main]: Destroying mutex 0x7fbd49ea03c8
[12/11 18:02:08.701 main]: adapter.c:1105: get_adapter returns NULL for adapter_id 2
[...]

PR #749, opened 2020-04-29
Support for reporting the closing of the socket

lars18th:
In the loop of sockets activity a new check is added for reporting to the adapter that the socket is closed.

This new functionality is based on this:
- The report is done calling to the function `action()` with a negative `rlen` size.
- This call is only done if the adapter understand this new semantic.
- To enable this new semantic for a socket it's required to use the new function `set_socket_disconnect_flag()`.
- When reporting the close when returning from `action()` the return value indicates: >=0 --> OK; <0 --> ERROR.
- When responding with error, the general behaviour is executed. When responding with ok, the loop continues.

The new behavioiur can be used to trigger some action in the adapter when this case is detected. In general, if the adapter catch this action, then the adapter will not be closed when the sockets is closed.

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -605,6 +605,7 @@ int sockets_add(int sock, USockAddr *sa, int sid, int type,
 	ss->lbuf = 0;
 	ss->timeout_ms = 0;
 	ss->id = i;
+	ss->flags = 0;
 	ss->sock_err = 0;
 	ss->overflow = 0;
 	ss->buf_alloc = ss->buf_used = 0;
@@ -923,6 +924,25 @@ void *select_and_execute(void *arg)
 							get_sockaddr_port(ss->sa), types[ss->type], err);
 						if (err == EOVERFLOW || err == EWOULDBLOCK)
 							continue;
+						if ((ss->flags & 2) && rlen == 0)
+						{
+							LOGM("select_and_execute[%d]: reporting that the remote client has closed the socket (rlen: %d, flags:%d)",
+							     i, ss->rlen, ss->flags);
+							ss->rlen = -1;  // This indicates that the remote end has closed the socket
+							errno = ss->action(ss);
+							ss->rlen = 0;   // Restore the value previoius to the call
+
+							if (errno < 0)
+							{
+								LOG("closing report of socket_id %d failed sid %d return err %d: %s",
+								    i, ss->sid, errno, strerror(errno));
+							}
+							else
+							{
+								LOGM("closing report of socket_id %d OK", i, ss->sid);
+								continue;
+							}
+						}
 						if (err == EAGAIN)
 						{
 							ss->err++;
@@ -1203,6 +1223,14 @@ void set_socket_receive_buffer(int sock, int len)
 	sl = sizeof(int);
 	if (!getsockopt(sock, SOL_SOCKET, SO_RCVBUF, &len, &sl))
 		LOG("receive socket buffer size is %d bytes", len);
+}
+
+void set_socket_disconnect_flag(int sock)
+{
+	sockets *ss = get_sockets(sock);
+	if (!ss)
+		return;
+	ss->flags |= 2;
 }
 
 void set_socket_pos(int sock, int pos)
--- a/src/socketworks.h
+++ b/src/socketworks.h
@@ -106,6 +106,7 @@ void free_pack(SNPacket *p);
 void sockets_setread(int i, void *r);
 void set_socket_send_buffer(int sock, int len);
 void set_socket_receive_buffer(int sock, int len);
+void set_socket_disconnect_flag(int sock);
 void set_socket_pos(int sock, int pos);
 void set_sock_lock(int i, SMutex *m);
 void set_socket_thread(int s_id, pthread_t tid);


lars18th:
Hi @catalinii ,

I'm working on fixing the issue commented here #636 . Futhermore it will fix other problems with some buggy SAT>IP servers.

But, before do this I need to add this support. This patch is fully transparent to current users, as the code is only executed if the flag is set. So it's safe to merge it.

Futhermore, this patch fixes the initialization of the `flag` member of the sockets structure. Until now is not initialized.

I hope you want to merge it. It will be useful in the future, not only for the `satipc` module, but perhaps for others too.

Regards.

lars18th:
Hi @catalinii ,

I really need this functionality to fix some troubles. Please, check it. If you have some disconformity with this patch, then comment it here. I'll try to change it if you want. However, in my case it works without any trouble.

I'm waiting for your feedback.
Regards.

catalinii:
I am trying to understand why would you want to keep a socket open if the remote closed it already. 
The general idea is that you may want to make the adapter survive a close of the socket from the satip server. For example, for the satip client I implemented restarting the connection if the server dies for whatever reason.

As the patch is implemented is hard for me to accept it as it is an antipattern to keep a socket open once the other party closed the socket.

lars18th:
Hi @catalinii ,

Sorry for a late response. I'm very busy these days. :fearful: 

I'll try to explain this patch, and the need for it.

- First, I want to comment that this is part of my implementation for supporting the Xoro STB devices. As you can see in several previous patches from me, the SAT>IP implementation on these devices is very weak. However, finally the only part pending for a full support in minisatip is one issue related to this patch. This part, that I already solved in the satipc module (not included in this patch), is the handle of the RTSP disconnections. These boxes have a rather strange implementation that closes the _listening_ RTSP if no activity is detected. This is true whether the RTP streaming is working or not. With some other SAT>IP clients this is not a problem, as the specification imposes a decoupling of the RTSP socket state with the RTP stream. In fact, there is nothing to prevent the server from closing the RTSP socket at some point. And this patch focuses on this.

- Second, please note the SAT>IP specification regarding the CLIENT point of view. The CONTROL plane is done over a TCP socket initiated from the client to the server. This connection has the objective to send/receive commands to/from the server to manipulate one/more RTP stream/s from the server to a multicast/unicast address. So, from a pattern point-of-view here the minisatip is the CLIENT. And the client initiates the connection, and it _reconnects_ when necessary. The decoupling of the RTSP protocol from the RTP streaming, in the sense of the socket state has no indication about the connection status, it's good to handle network errors and overcome for glitches. Then please, note of this: A SAT>IP client *cannot* interpret anything about a disconnection of the RTSP socket. From the point-of-view of the client the server continues in the same state. So if it (the client) wants to stop the streaming, then it needs to reopen the socket and send a TEARDOWN. And if it wants to continue the streaming, then it needs to reopen the socket and send the OPTIONS command to maintain the stream before the timeout. That's the *expected* client behaviour from specs if the RTSP socket is closed.

- Third, to hanle the above behaviour it's mandatory to inform to the module (the one that uses a TCP connection) about this event. I have implemented this improvement in this patch in an agnostic way. And I've tried to touch the code as few times as possible. I don't think that this implementation is an antipattern. Your implementation is correct, but oriented to _listening_ sockets (aka server sockets). When working as a network client, it's the responsability of the client to handle for any reconnection. And to add support for it, it's required to receive the event. That's what implements this patch.

At this point, I want to comment that if you have another more useful and/or elegant solution to solve this problem (handle the TCP close event), then please provide your implementation. I've detached this part from the satipc patch because just this reason. However, I feel my proposal is simple and clean. Futhermore, please note that perhaps other modules will require to know if a TCP _client_ socket is closed. So this small and non-intrusive patch could be useful in those cases as well.

Please, let me know if you need more information about this patch.
I'll stay waiting for the required support to publish the satipc part to handle the reconnections.
Regards.

lars18th:
Hi @catalinii ,

Please, can you give me some feedback? I really need a solution to receive and process TCP disconnections of the RTSP channel. Without this functionality I can't continue fixing the satipc module. I'm completely blocked by this.

Also, I need to comment that using my patch for weeks I haven't seen any trouble or side effects. So please consider it or provide an alternative.

Regards.

lars18th:
Hi @catalinii ,

Please, give me some feedback. It's absolutly necessary to have a method to handle the RTSP socket disconnection. The current assumption that the closing of the RTSP TCP socket from the server implies a server error or anything other is false.

I repeat another time what I've implemented:
- If a TCP socket disconnection is detected then a new flag is checked.
- The flag indicates if the TCP module that handles the connection wants to receive a signal when the TCP socket is closed from the other side.
- Then if the TCP socket is closed and the flag is on, then the module is informed using a simple trick: the `action()` function is called using a read length of `-1`.

And as you can see if the flag is not set, then nothing changes.

If you can provide another different implementation, please share it. In any case, the `satipc` module **requires** a method to be informed about the RSTP disconnection. Closing the device when this event occurs without further checks is an error.

I wait for your response.
Regards.

catalinii:
@lars18th my point was to keep the disconnection of the rtsp socket in satipc module.

I have done this for RTSP over TCP ... in some cases allow the rtsp tcp connection to be disconnected while the adapter is still open. I did not test UDP but probably is not complicated to change that:
https://github.com/catalinii/minisatip/commit/1deff1fda62fdf1a31ec93fb017d2896822765f5

Probably testing for UDP and added small fixes should make it work also when the server disconnects a RTSP over UDP connection.

lars18th:
Hi @catalinii ,

I can't found the way you resolved the problem in the `satipc` module. Please, remember the question: Some SAT>IP servers (i.e. XORO devices) close the RTSP socket if not receive a new command in less than 2-3 seconds. So, for practically _every_ command you need to re-open the RTSP socket (including the TEARDOWN, that the devices requires to free the tuner... In fact, if you forget to send the TEARDOWN the device will be locked for several minutes and it requires a manual reboot). So please, if you can help me to incorporate the code for re-opening the RTSP socket _without touching the internal state machine_ (as the streaming of the RTP packets continues, and the server doesn't changes the internal state). From my point of view this is impossible without touching the `socketworks` module. Perhaps you can provide one example implementation or you can provide a direct fix for this problem.

So please, comment it.
Thank you!

catalinii:
What if the timeout is every second ? Basically keep sending options very often, does that work ?

lars18th:
Hi @catalinii ,

> What if the timeout is every second ? Basically keep sending options very often, does that work ?

Yes. However, in fact this is not the solution to the case when the server closes the RTSP TCP socket but continues working, so the session continues active (in PLAY state) and the RTP streaming is working.

After more testing with such XORO devices, I found that the best strategy is to work with such devices is using this technique:
- Run as normal if you send commands with less than 1 second in between. This is true for alls commands, with the exception of PLAY, and partially with OPTIONS and TEARDOWN. For example, if you start a session, then you send OPTIONS, DESCRIBE, SETUP and PLAY. Then the TCP socket continues open until the last PLAY, as around 1-2 seconds after the server closes the socket (but the session continues open).
- To overcome the close of the TCP socket, the best strategy is to _close_ the socket (from the client) after sending the command that you expect that will not have more commands after. For example, after every PLAY; and also after one keep-alive OPTIONS and the TEARDOWN. This guarantees that the RTSP will be closed when you send the next command. If you don't work in this way, then the socket can be in the state closed or open when you want to send the next command. So it's preferable to be sure about the socket state.
- Then, the regular behaviour is to open the socket after sending the command if it's not open. That's all! Simple and efficient. And it really works with these stupid boxes.

If you agree with this technique, then my suggestion is this:
- First, add the option to identify such boxes in the `sastipc` module (I prefer using a command-line optional parameter).
- If the previous flag is set, then close the RTSP socket after _some_ commands (PLAY, keep-alive OPTIONS and TEARDOWN), but don't close the session or change anything else in the adapter structure (the adapter needs to think that the RTSP port is open).
- Before sending any new command, if the flag is set, then open the RTSP socket.

With these changes in the `satipc` module, the loop of the socketwork doesn't require to be modified. And therefore any closing action of the TCP socket from the _server_ can continue to be handled as it's now implemented. And you can continue assuming that a close of the RTSP socket will implies the closing of the session.

What you think now with this alternative?

catalinii:
I am fine with this implementation as long as 1) separate this mode with command line argument - as you mentioned
2) don't touch socketworks
3) leverage existing code for tcp mode (rtsp_socket_open, .... )

For tcp mode close of tcp connection means close of session. This is not true for udp mode (according to the specs).

lars18th:
Hi @catalinii ,

> For tcp mode close of tcp connection means close of session. This is not true for udp mode (according to the specs).

That's the key question! The current implementation of the `satipc` module implies that if the _server_ closes the TCP socket of the RTSP session, then minisatip understands that the session is closed. That's what the _current_ code is doing. And we need to fix this, as obviously this is false according to the specifications.

Regards.

Jalle19:

PR #780, opened 2020-11-26
New values in the Web UI

lars18th:
Now these values are included in the STATUS webpage:
- Decoding errors of the adapter (when decoding).
- BOOT-ID and DEVICE-ID values used in the configuration.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -2345,6 +2345,19 @@ int get_adapter_ccerrs(int aid)
 	return cc;
 }
 
+int get_adapter_decerrs(int aid)
+{
+	int i, dec = 0;
+	adapter *ad = get_adapter_nw(aid);
+	if (!ad)
+		return 0;
+
+	for (i = 0; i < 2; i++)
+		if (ad->pids[i].flags == 1)
+			dec += ad->pids[i].dec_err;
+	return dec;
+}
+
 _symbols adapters_sym[] =
 	{
 		{"ad_enabled", VAR_AARRAY_INT8, a, 1, MAX_ADAPTERS, offsetof(adapter, enabled)},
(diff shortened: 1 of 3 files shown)


lars18th:
Hi @catalinii ,

If you prefer, replace the text "DEC-ERRORS:" with "Undecoded packets:". Maybe that sounds better.

And one question: It will be posible to reset (set 0) of these errors until the decoding process starts? For example, when no key is found (at start), then could you not increase the dec_err counter?

Regards.

src/adapter.c
```diff
@@ -2367,6 +2380,7 @@ _symbols adapters_sym[] =
 		{"ad_allsys", VAR_FUNCTION_STRING, (void *)&get_all_delsys, 0, MAX_ADAPTERS, 0},
 		{"ad_pids", VAR_FUNCTION_STRING, (void *)&get_adapter_pids, 0, MAX_ADAPTERS, 0},
 		{"ad_ccerr", VAR_FUNCTION_INT, (void *)&get_adapter_ccerrs, 0, MAX_ADAPTERS, 0},
+		{"ad_decerr", VAR_FUNCTION_INT, (void *)&get_adapter_decerrs, 0, MAX_ADAPTERS, 0},
```
catalinii, line 2383: Probably is better to use ad->dec_err for the number of decryption errors instead of per pid counter. ad->dec_err is incremented once the decryption process ends. Feel free to reset ad->dec_err after tune.

lars18th: @catalinii ,

Please, note this: First I implemented it with `ad->dec_err` but I found that this is not useful. Why? Two reasons:

1. This references the TOTAL number of decoding errors of the adapter. So when you change the service, the number is not reseted (obviously). So what in fact we want is the number of decoding errors when _"deconding a service"_. That's what we want and what we need. And for this reason the print is done inside the colum of decoding, and not in the signal one.
2. Futhermore, when using the _pass unencrypted packets_ option, the number of `ad->dec_err` is not increased. So, I made a modification to increase the counter when this option is in use. But I don't add to this patch as it's useless for it. You suggested to reset `ad->dec_err` after a new tune, and I agree with you. Perhaps I can provide _another_ PR with these two enhancements: reset `ad->dec_err` and increase this counter when using the _pass unencrypted packets_ option. You agree with this?

So, a simple conclusion now: this patch looks good to me as it.
I hope you want to merge it.
Regards.

catalinii: Thanks @lars18th for the testing. Sounds good, will merge it.

PR #781, opened 2020-11-26
Emulated pids all: Include a fixed pid list

lars18th:
When using the emulated-pids-all mode, include a list of fixed missing pids related to standard DVB tables. Without these pids the output is not DVB conformant and some pids aren't included in the output stream. The list doesn't include the NULL pid (8191).


lars18th:
Thank you for merging it! 👍

PR #782, opened 2020-11-26
Debug: Minimal fixes

lars18th:
Some small changes in the log:
- Fix an error when enabling the log of a module that the name has collision with other module names.
- Missing CSA default module.
- Improve the message of the Mutex initialization.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -353,7 +353,7 @@ int map_intd(char *s, char **v, int dv)
 		return atoi(s);
 	}
 	for (i = 0; v[i]; i++)
-		if (!strncasecmp(s, v[i], strlen(v[i])))
+		if (strlen(s) == strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
 			n = i;
 	return n;
 }
--- a/src/utils.c
+++ b/src/utils.c
@@ -1266,7 +1266,7 @@ int mutex_init(SMutex *mutex)
 	mutex->create_time = getTick();
 	mutex->enabled = 1;
 	mutex->state = 0;
-	LOG("Mutex init %p", mutex);
+	LOG("Mutex init OK %p", mutex);
 	return 0;
 }
 
--- a/src/utils.c
+++ b/src/utils.c
@@ -353,7 +353,7 @@ int map_intd(char *s, char **v, int dv)
 		return atoi(s);
 	}
 	for (i = 0; v[i]; i++)
-		if (strlen(s) <= strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
+		if (!strncasecmp(s, v[i], strlen(v[i])))
 			n = i;
 	return n;
 }
--- a/src/utils.c
+++ b/src/utils.c
@@ -353,7 +353,7 @@ int map_intd(char *s, char **v, int dv)
 		return atoi(s);
 	}
 	for (i = 0; v[i]; i++)
-		if (strlen(s) == strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
+		if (strlen(s) <= strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
 			n = i;
 	return n;
 }
--- a/src/csa.c
+++ b/src/csa.c
@@ -47,6 +47,8 @@
 #include "pmt.h"
 #include <dvbcsa/dvbcsa.h>
 
+#define DEFAULT_LOG LOG_DVBCA
+
 void dvbcsa_create_key(SCW *cw)
 {
 	cw->key = dvbcsa_bs_key_alloc();


lars18th:
Hi @catalinii ,

I hope you agree with these small improvements. I'll try to extract some of the changes from #744 in small chunks.

Regads.

src/utils.c
```diff
@@ -353,7 +353,7 @@ int map_intd(char *s, char **v, int dv)
 		return atoi(s);
 	}
 	for (i = 0; v[i]; i++)
-		if (!strncasecmp(s, v[i], strlen(v[i])))
+		if (strlen(s) == strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
```
catalinii, line 356: Not sure this is a good change. The whole point of mapintd is to check the first part of both strings. There are plenty of places that can happen. Of course the new line is equivalent of strcasecmp.

Where is this check going wrong currently?

lars18th: Hi @catalinii ,

> Where is this check going wrong currently?

I commented previously. Even though, I repeat it: 😉 
- When the name of two _different_ modules have a collission, the incorrect module is selected. You can test it enabling only the debug of the module `socketworks`. So, as the module `socket` is a substring of the `socketworks`, and it's listed before, the code selects the module "socket" when you request to use "socketworks". So instead of changing the name of a module (as I feel you don't want to do it), I provide this definitive patch.

I'm surprissed you never see this before!
And, why you want to use "substrings" and not the full name?

lars18th: Hi @catalinii ,

No feedback about my last comment? Why you want to support substrings? And what you prefer to fix the current bug: change the name of the "socketworks" module or disable the substrings support?

Please, note that bug exists and we need to resolve it!

catalinii: I think the best way is to create a new function that uses strcasecmp and use that function instead of map_intd.

lars18th: Hi @catalinii ,

> I think the best way is to create a new function that uses strcasecmp and use that function instead of map_intd.

Ok. Now I see the point: the `map_intd()` function is used for other uses too. So, _only_ for the detection of the modules name here: https://github.com/catalinii/minisatip/blob/da53e5a9b26da9bd28fc71aa63958e5635fa1cc1/src/minisatip.c#L681 it will be necessary to use another different (aka. new) function.

That's your opinion?

catalinii: Yeah, create a new function: string_to_index

lars18th (commented):
Hi @catalinii ,

I hope you accept my last modification. I prefer to not generate a new function for the same functionality. And with this simple check of not be longer than the target string to compare the bug is resolved.

Please, check it! :wink:

src/utils.c
catalinii: @lars18th the previous yeah was not applying to your last comment.

Basically a new function is a better option as loose matching is not a good idea in general as it can create such issues you are pointing out. I would rather migrate some map_intd use cases to the new function. The other problem with this function is that it acts also as atoi.

I think creating a function that just gets the index of a string in an array makes more sense.

Once is done you can try to add some unit tests in test_utils.c to make sure it does what you want.

lars18th: Hi @catalinii ,

Thank you for the response. So the `map_intd()` requires to be reimplemented with a new name. OK. However, I don't have sufficient time now to do all tasks that you request. So, please, as the `<=` check doesn't create new side effects, and solves the bug of "socketworks" logging, then you will be so kind as to accept this PR?

Regards.

lars18th: Hi @catalinii ,

(ping) Are you confortable with the current (the `<=` check) workaround that resolves the _current_ bug? In the meantime I feel is preferable to use it.

Please, comment it.
Thank you!

lars18th:
Hi @catalinii ,

I'm rewriting my full PR #744 (the third iteration, so a large work) in small chunks. However, I'm not sure if you agree with my suggestions. Please, can you recheck this _simple_ PR? It targets very single troubles.

I want to continue with other issues, but I prefer first to close this (merging it, of course!).

src/utils.c
catalinii: The use cases I mention that the function will break is exactly when strlen(s) > strlen(s[i]).

With the current method you cannot fix both cases: the bug in LOG and the issues I mention about. You need 2 methods to fix both of them.
map_intd is used intensively to parse the stream parameters (like msys, mod, ... ). Breaking that part will make minisatip  unusable that's why I am reluctant to accept the patch as it is.

lars18th: > The use cases I mention that the function will break is exactly when strlen(s) > strlen(s[i]).
> 
> With the current method you cannot fix both cases: the bug in LOG and the issues I mention about. You need 2 methods to fix both of them.
> map_intd is used intensively to parse the stream parameters (like msys, mod, ... ). Breaking that part will make minisatip unusable that's why I am reluctant to accept the patch as it is.

OK, now I understand the reason. So, the only option is to create or use another different function when checking the LOG. 👍

lars18th:
Hi @catalinii ,

I removed the map_intd() change. I'll provide another PR for it. In the meantime, please merge this super-simple PR.

Thank you!

Issue #795, opened 2021-01-15
sendmmsg(): errno 38: Function not implemented


Jalle19:
Reported here: https://github.com/Jalle19/satip-axe/issues/3

satip-axe runs Linux 2.6, while `sendmmsg` was introduced in Linux 3.0 (https://man7.org/linux/man-pages/man2/sendmmsg.2.html)

https://github.com/catalinii/minisatip/issues/346 seems to be related - apparently the Android NDK doesn't expose `sendmmsg` either.

@catalinii what do you suggest? The ifdef should probably be extended somehow.

catalinii:
Thanks for reporting this, was not aware.
I think the only thing needed is to test for AXE here:
https://github.com/catalinii/minisatip/blob/31c8d4b99d226568a9322bc4eaa6c3606786088b/src/socketworks.c#L1220

Jalle19:
Yeah that should work, I'll try to get some actual runtime testing done during the weekend.

Jalle19:
@catalinii is it really worth keeping macOS compilation support around?

catalinii:
I use it for local development.... if the delta is small then why not.

Jalle19:
Ran into this again when attempting to run this on an ancient DM800 HD. I modified it so that I could pass `CFLAGS="-DNEEDS_SENDMMSG_SHIM"` in order to force the compiler to include the custom version of it. Do you think that's worth making a pull request for @catalinii ? I also tried using AC_CHECK_FUNCS but it does the detection on the host system so it's not very useful.

catalinii:
Yes, I think is a good idea

PR #796, opened 2021-01-15
Ensure _GNU_SOURCE is always defined

Jalle19:
Fixes https://github.com/catalinii/minisatip/issues/795


Jalle19:
Actually, this probably won't fix anything, except make it a bit cleaner perhaps. socket.h for satip-axe has this:

```
#ifdef __USE_GNU
/* Send a VLEN messages as described by VMESSAGES to socket FD.
   Returns the number of datagrams successfully written or -1 for errors.

   This function is a cancellation point and therefore not marked with
   __THROW.  */
extern int sendmmsg (int __fd, struct mmsghdr *__vmessages,
                     unsigned int __vlen, int __flags);
#endif
```

So I guess that since it compiles and doesn't crash, it means the implementation is missing from the kernel, hence glibc gives the caller `errno 38`.

Jalle19:
@catalinii do you want to merge this or should I close it? I guess it's a matter of personal preference how these various macros are defined

catalinii:
I prefer keeping it as it is but if you identifoed something missing in the current approach feel free to let ne know

Jalle19:
Nah, not really. I'll close this, maybe if in the future there are more of these it could be worth defining them all in one place.

catalinii:
Thanks

PR #797, opened 2021-01-15
Use writev on axe devices

### LINKED ISSUE #795: sendmmsg(): errno 38: Function not implemented

melias122:
Fixes #795
Fixes https://github.com/Jalle19/satip-axe/issues/3

### DIFF
--- a/src/socketworks.c
+++ b/src/socketworks.c
@@ -1222,13 +1222,16 @@ struct mmsghdr {
     struct msghdr msg_hdr; /* Message header */
     unsigned int msg_len;  /* Number of bytes transmitted */
 };
+#endif
 
-int sendmmsg(int rsock, struct mmsghdr *msg, int len, int t) {
+#if defined(__APPLE__) || defined(AXE)
+int sendmmsg0(int rsock, struct mmsghdr *msg, int len, int t) {
     int i;
     for (i = 0; i < len; i++)
         writev(rsock, msg[i].msg_hdr.msg_iov, msg[i].msg_hdr.msg_iovlen);
     return len;
 }
+#define sendmmsg(a,b,c,d) sendmmsg0(a,b,c,d)
 #endif
 
 int writev_udp(int rsock, struct iovec *iov, int iiov) {


melias122:
Tried `if defined(__APPLE__) || defined(AXE)`, but that lead to
```
socketworks.c:1227:5: error: conflicting types for 'sendmmsg'
 int sendmmsg(int rsock, struct mmsghdr *msg, int len, int t) {
     ^
In file included from /opt/STM/STLinux-2.4/devkit/sh4/target/usr/include/sys/socket.h:40:0,
                 from /opt/STM/STLinux-2.4/devkit/sh4/target/usr/include/netinet/in.h:25,
                 from /opt/STM/STLinux-2.4/devkit/sh4/target/usr/include/arpa/inet.h:23,
                 from socketworks.c:20:
/opt/STM/STLinux-2.4/devkit/sh4/target/usr/include/bits/socket.h:437:12: note: previous declaration of 'sendmmsg' was here
 extern int sendmmsg (int __fd, struct mmsghdr *__vmessages,
            ^

```

catalinii:
Where do u have axe sdk from?

Jalle19:
I'm personally using https://github.com/Jalle19/docker-centos7-stlinux24/blob/master/Dockerfile

Jalle19:
It's a bit tricky because sendmmsg() is defined in glibc, it's just not implemented in the kernel.

Maybe the only solution is to simply not use sendmmsg() anywhere?

melias122:
I am using satip-axe. 

Not using sendmmsg will be simplest, but I am not sure about performace of writev vs sendmmsg.

Martin Elias

catalinii:
@Jalle19 probably is better if you define this method as:
sendmmsg0(int rsock, struct mmsghdr *msg, int len, int t)

and for axe and MAc os just do:
#define sendmmsg(a,b,c,d) sendmmsg0(a,b,c,d)

melias122:
@catalinii, updated PR, but it is not tested yet

catalinii:
Looks good from my point of view. If you guys don't manage to test it on axe as I do not have one will try on MAC.

melias122:
I will test it during the day, when nobody is watching tv in the house and let you know.

Martin Elias

Jalle19:
I use TCP so I'm not affected by this :D

Jalle19:
@melias122 did you get around to testing this?

melias122:
Yes, it works fine on digibit r1.

PR #800, opened 2021-01-29
WiP: first version NationalChip GX support

danielkucera:
Added support for NationalChip GXAPI (GX6605S) SoC, for `koqit k1 mini` device, taken from https://github.com/McMCCRU/minisatip-gx

Please review and comment if this has a chance to be merged.


catalinii:
Hey @danielkucera ,

Thanks for submitting the PR.

The only major comment I have so far is to separate the code out of adapter.c and dvb.c and create a separate (gxapi.c or use avapi.c) to implement device functions in the same way that axe.c does it. If the functions does not change feel free to use the method from dvb.c.

src/dvb.c
```diff
@@ -412,38 +436,117 @@ void dvb_set_demux_source(adapter *ad)
...
+		return;
+	}
+	ad->demux_lock = 1;
 #endif
 }
 
 int dvb_open_device(adapter *ad)
 {
 	char buf[100];
+	int fn = ad->fn;
```
danielkucera, line 493: @McMCCRU , why have you defined a new variable `fn` here? Is there any reason to not use `ad->fn` ?

McMCCRU: Hi! 
This was done to support combo devices with using an optional external demodulator and tuner(DVB-C/T/T2) on the gx6605s. The DVB GXAPI is based on the old Linux DVB API v.3.x and does not support the operation of a multi-frontend with one demultiplexer, so I made it so that would not write set values fn to ad->fn, which is further used in original minisatip.
P.S. If I'm not mistaken, then the minisatip itself does not support multi frontend on one adapter, only one adapter and one frontend...

lars18th:
Hi @danielkucera ,

This PR seems very interesting. Please, can you split it in several parts to maintain it? The idea is to separate: general code (parts independent of the GX support), headers (the added part of the `src/gxapi/` directory), GX adaptation (the new code added), GX compilation (the changes to enable and complie it).

I hope this support will be included at some point, as perhaps it will enable a future support for other low-end platforms.
Regards.

danielkucera:
Hi @lars18th , I put this a little bit on hold because I have found out there is a possibility to compile mainline kernel for this device so it only needs dvb device drivers to be ported and standard minisatip would work.
I plan to get back to this in some time, not sure when exactly.

lars18th:
Hi @danielkucera ,

Any improvement?

danielkucera:
Hi @lars18th ,
I currently don't have to satellite dish so I cannot test so I cannot develop.
I plan to work on this second half of this year.
If this PR annoys you, feel free to close. I will open a new one when I make some progress.

Jalle19:

Issue #577, opened 2019-04-05
BUG: Can't stream two programs when using HTTP


lars18th:
Hi,

I feel this is a bug:

- I do this test: two SAT>IP clients and one HTTP client (VLC with a full `http://...:8080/?src=...&pids=...`), so a total of 3 clients. I request from the first SAT>IP a program in a transponder. Then in the second SAT>IP client I request another program in the _same_ transponder. Finally, from the HTTP client I request another program in the _same_ transponder.
- The result is: the HTTP client can't receive the stream!

And this is true every time an HTTP client requests some program from an active adapter, even if the same source is used. So the same result when just one SAT>IP client and one HTTP client. Futhermore, if I do the opposite (first tune with the HTTP client and _after_ with the SAT>IP client) then the connection is accepted.

So the problem is only when an HTTP tunning is requested over an already running transponder.

Please, review the code. And check it by yourself! No logs are required. It's trivial to reproduce the behaviour!

9000h:
card/driver pid limit ?

lars18th:
> card/driver pid limit ?

No! I'm using the SAT>IP client mode. And **indentical** configuration of client programs work when _all_ request are using SAT>IP. The problem is just when connecting using the HTTP protocol.

I feel something is incorrected checked when requesting using the HTTP protocol.

catalinii:
@lars18th, can you try a different client than vlc ?

I tried as you suggested but used ffmpeg as http client and worked good.

lars18th:
Hi @catalinii ,

I discovered the cause of this issue. You can reproduce if you like, and I feel is simple to fix it:

- The problem is caused when the HTTP client requests _less_ parameters than the current tuned parameters. For example, if the `mtype` is missing (because the client assumes to use _auto_ mode) but the previous request specifies a `mtype` then the HTTP connection is refused. At time, when the same HTTP requests without the `mtype` and no session exists, then isn't any problem, as the tuner uses _auto_.

So, the solution I recommend is this: when a client (HTTP _or_ RTSP, don't care about this) requests a new TUNE command and the parameters are a _subset_ of the current parameteres in the same frequency and source, then use the current parameters and accept the command (instead of returning an error).

You agree?

9000h:
Hi,
HTTP clients/servers are not specified in the SAT>IP spec, it's a freestyle extensions, and more or less only implemented to make tests easier.
CU
9000h

lars18th:
Hi @9000h ,

[...]

PR #801, opened 2021-01-29
Fixes concurrent clients sharing the frequency

lars18th:
This fixes some cases when two (or more) clients using RTSP and HTTP want services in the same transponder.
Futhermore, if some adapter is disabled at the start of the free list, then this is not selected.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -688,7 +688,12 @@ int get_free_adapter(transponder *tp) {
     int msys = tp->sys;
     int fe = tp->fe;
 
-    adapter *ad = a[0];
+    adapter *ad = a[0]; // If none is found then the first is the default!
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

This solves (some part) of the issue #577, as when streaming to one HTTP client, the second RTSP client can't connect without this patch. So it complements your commit https://github.com/catalinii/minisatip/commit/80239d45cbd582f8cf79539d07f89bede1392125 time ago.

Futhermore, I detected why you can't reproduce my troubles: I'm disabling the internal tuner 0 (with `-e 1-2`) and using only my external SAT>IP tuners. So, this patch includes the part that fixes my use case (that in fact is related to this). The problem is when you try to iterate over the list of free adapters, but the first is disabled.

Please, test it and merge when you can.
Regards.

PR #806, opened 2021-02-04
fix return code

9000h:
fix return code


9000h:
you did commit the void to int but the return 0 is still missing

PR #815, opened 2021-03-07
FIX of debug -l socketworks

lars18th:
When enabling the debug of the "socketworks" module the enabled module is the "socket" one, becase the name "socket" is contained inside "socketworks".
This fixes the problem enabling only a module when the exact name matches. To do this is used a new function maps_intd() that is used only here.


lars18th:
Hi @catalinii ,

This resolves the bug using your comments from previous discussions (see for example https://github.com/catalinii/minisatip/pull/744#discussion_r419045585). You suggested at some point to implement a new reading function for the log. So I've implemented that. You agree with this PR?

lars18th:
Hi @catalinii ,

Any problem with this PR?

src/utils.c
```diff
@@ -367,6 +367,31 @@ int map_intd(char *s, char **v, int dv) {
...
+        LOG_AND_RETURN(dv, "maps_intd: s is empty");
+
+    if (v == NULL) {
+        if (s[0] != '+' && s[0] != '-' && (s[0] < '0' || s[0] > '9'))
+            LOG_AND_RETURN(dv, "maps_intd: s not a number: %s, v=%p, %s %s", s,
+                           v, v ? v[0] : "NULL", v ? v[1] : "NULL");
+        return atoi(s);
+    }
+    for (i = 0; v[i]; i++)
+        if (strlen(s) <= strlen(v[i]) && !strncasecmp(s, v[i], strlen(v[i])))
```
catalinii, line 390: This is equivalent with strcasecmp

src/utils.c
```diff
@@ -367,6 +367,31 @@ int map_intd(char *s, char **v, int dv) {
     return n;
 }
 
+int maps_intd(char *s, char **v, int dv) {
```
catalinii, line 370: Please find a more suggestive name for the function

lars18th: Any suggestion? I selected this as `map_intd()` is transformed to `map_special_intd()` and contracted to `maps_intd()`. What you want? `map_with_strcasecmp_intd()` or something similar?

src/utils.c
lars18th: > This is equivalent with strcasecmp

Are you sure?

Check these four cases:
1. Check "socket" with "socketworks".
2. Check "socketworks" with "socketworks".
3. Check "sock" with "socketworks".
4. Check "socketworks" with "socket".

If we don't do the check `(strlen(s) <= strlen(v[i])` the case 4 is incorrectly handed. And this is the case that this patch is targeting. I accept that a "substring" can be used in the parameter list; but this "substring" match requires to be exact in case of the target is smaller than the passed name. So in any case "socketworks" can be processed as a "socket" module name.

catalinii: To perform exact match (case 2), strcasecmp can be used and checked for 0

https://man7.org/linux/man-pages/man3/strcasecmp.3.html

 strcasecmp(const char *s1, const char *s2);
       
DESCRIPTION         top
       The strcasecmp() function performs a byte-by-byte comparison of
       the strings s1 and s2, ignoring the case of the characters.  It
       returns an integer less than, equal to, or greater than zero if
       s1 is found, respectively, to be less than, to match, or be
       greater than s2.

src/utils.c
catalinii: search_string_in_array?

lars18th: Hi @catalinii ,

Before change it, I want to be sure: You agree with the name `check_strs()` ?

lars18th (commented):
Hi @catalinii ,

> Before change it, I want to be sure: You agree with the name check_strs() ?

Just waiting for the final name of the function to generate the final version of the PR.
Please, comment if you agree with the name `check_strs()`.

Note: The code now works as expected!
Note 2: It includes now a method to print in the LOG wich modules are enabled... `minisatip ... -l sock,dm -v pmt` prints this:
```
[08/04 09:56:23.530 main]: Built without t2mi
[08/04 09:56:23.530 main]:  LOG  of the module enabled: general
[08/04 09:56:23.530 main]:  LOG  of the module enabled: pmt
[08/04 09:56:23.530 main]: DEBUG of the module enabled: pmt
[08/04 09:56:23.530 main]:  LOG  of the module enabled: socket
[08/04 09:56:23.530 main]:  LOG  of the module enabled: dmx
```
So a relevant improvement.
Please, comment to finish this PR.

src/utils.c
lars18th: Hi @catalinii ,

The code is updated and cleaned. I hope you agree with the final implementation.

lars18th:
Hi @catalinii ,

From my point of view the PR is completed. Please, test it and merge (after)! :wink:

Issue #810, opened 2021-02-26
Suggestion: Fast HTTP reconnections


lars18th:
Hi @catalinii ,

I have an idea about a new extesion for minisatip, but before describe it I need to explain how to reproduce a test that illustrates the use case:

1. Create a M3U file with an HTTP link of one program of your minisatip server, for example:
```
#EXTM3U
#EXTINF:0,KiKA HD
http://192.168.1.35:8080/?src=1&freq=11347&pol=V&msys=DVBS2&mtype=8PSK&sr=22000&pids=0,6610,6620,6600,6610,6630
```
2. Open one VLC instance and open the M3U file, you'll see the service.
3. Open _another_ VLC instance and open the same M3U file, you'll see two times the same service.
4. Now open the PlayList (CTRL + L) of one of the two VLC and _click on the service_. The result is that the HTTP stream is closed, and a new session is opened. And what you see in this instance is a small glitch in the playing... around 300-500ms.
5. But, now, close the second VLC instance, and execute the same procedure. What you observe? Around a cut of 2-3 seconds in the best case.

So, what's the problem that I want to illustrate? Using the HTTP streaming, the close of the socket implies _every time_ the close of the session. And this generates some troubles with HTTP only clients, when the socket is closed by whatever reason and it wants to reconnect.

Then my suggestion is a workaround to overcome this case. The concept is based on a _delayed HTTP session close_. Mainly is quite simple: when one HTTP client socket is closed, the session is not closed directly, but queued. Then a timeout is started and only when it expires the session will be really closed. The timeout can be set around 5 seconds. This will enable the option for a fast HTTP reconnection. However, one case requires to be addressed. It's when the client reconnects to make a request of _another_ different service. In this case the queued session requires to be closed inmediatly. Otherwise, the new tunning can't be served (even if it's in the same transponder). So to handle this case, you can store the IP of the client with the session in the queue, and only if this client reconnects (assuming that that the addess is not shared) and the request is different, then you stop the queued session before serve the new session.

What you think about this enhancement? My concern is with HTTP clients that doesn't support RTSP, for example Android devices (STBs or Smartphones/tablets) that work as regular IPTV clients.

catalinii:
Let me understand the requirement:
1) client A connects to minisatip and plays the stream using http
2) client A disconnects from minisatip for whatever reason
3) client A reconnects to minisatip (let's say after 2s)

What is the expectation related to stream playing after the reconnection? To play exactly where it left off? 

[...]

PR #817, opened 2021-03-09
satipc: Manage delayed TEARDOWN commands

lars18th:
When a satipc adapter is closed, in some circunstances the TEARDOWN command can't be sended. That's because the client module is still waiting for the previous response and it doesn't have the sesssion/stream identifiers. In this case, it's necessary to store the identifiers when the response is received. And if a new SETUP command is required to be committed after that, then the session and stream ids will be reused and the protocol continues as if the TEARDOWN never exists (so it sends a PLAY instead a SETUP).


lars18th:
@catalinii ,
This PR fixes the trouble described in https://github.com/catalinii/minisatip/issues/810#issuecomment-793744630.

I hope you want to merge it soon.

src/satipc.c
```diff
@@ -1216,11 +1237,11 @@ int satipc_commit(adapter *ad) {
     send_apids = send_apids && sip->lap > 0;
     send_dpids = send_dpids && sip->ldp > 0;
 
-    if (getTick() - sip->last_setup < 1000 && !sip->sent_transport) {
-        LOG("satipc: last setup less than 10 seconds ago for server %s, maybe "
+    if (getTick() - sip->last_setup < 5000 && !sip->sent_transport && sip->last_cmd != RTSP_TEARDOWN) {
```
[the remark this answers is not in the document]
lars18th: Hi @9000h ,

Original code was wrong! 1 second are 1000 miliseconds. And the `getTick()` function returns values in miliseconds. So before the PR it only waits for 1 second (even the comment say 10sec). So I extended the time to 5 seconds. What is safe, as the code is only triggered if the last command **is not** a TEARDOWN.

Futhermore, I've cheked this code for some time. And it works nice, and it has resolved some troubles. So, it's necessary.
I hope @catalinii will merge soon.
Regards.

src/satipc.c
```diff
@@ -1110,6 +1119,7 @@ int http_request(adapter *ad, char *url, char *method, int force) {
     if (sip->stream_id != -1)
         sprintf(sid, "stream=%d", sip->stream_id);
     else if (!strcmp(method, "TEARDOWN")) {
+        sip->uncommitted_sid = 0;
```
catalinii, line 1122: Here the value is set to 0, everywhere else is -1? Is there a need for 2 values or 1 value (-1) is enough?

lars18th: Hi @catalinii ,

The use of the member `uncommitted_sid` is this:
- `-1` : init value, no TEARDOWN send, so regular behaviour.
- `0`  : TEARDOWN command delayed, pending to read the "stream_id".
- `>0` : Value of the "stream_id" active because a TEARDOWN command is not sended.

So, the short response to your question is: NO.

Long answer: The TEARDOWN command can be delayed, not because the module wants to delay it, but because the RESPONSE of the previous command has not yet been received. Then you can't close the session as it's possible you don't know the "stream_id". So the code waits until the pending command is answered and then it copy the "stream_id" in the `uncommitted_sid` variable. Then the execution continues reusing this value in the next command (instead of sending a new SETUP it send a PLAY with the same "stream_id").

The code is then correct. I'm using it for several days and the behavious has improved a lot. So, please commit it.
Regards.

catalinii: @lars18th I do see the issue you are mentioning happening and I agree it is a problem. 

And thinking about it is that I am using an async way of communicating with the server by caching future commands instead of waiting for the answer.
Eg. Send tune, wait for answer then respond to the satip client.

But the issue with this patch is more that feels like a band aid solution instead of fixing the scenario: PLAY after a cached TEARDOWN. 
I also agree that handling of stream_id and session is unfortunate. It should be cleared only when TEARDOWN is sent.

I would prefer fixing this scenario:
- fix habdling of session, stream_id with no new variables
- if a teardown is cached and a play comes, delete the cached teardown.


Otherwise by accepting the patch It would be very hard to fix if a new issue ie discovered in the previous code.

lars18th: Hi @catalinii ,

Sorry, but I feel I can't solve the issue using what you propose. Let me explain why:

- The current code is only detecting an edge case: when the module is instructed to send a TEADOWN message and it can't be sended as the stream_id is unkown, because the precedent command is pending and the response has not yet been received.
- So the code only targets this specific case with minimal changes and without adding innecessary variables.

Then my questions are:
- How I can handle this case without new variables? I need a flag to know when it appears, and I only added one int var (`uncommitted_sid`) and a cache of the session id string (`uncommitted_session`).
- I don't know how to "delete the cached TEARDOWN command". In fact, as the command is not sended we don't need to cache anything. When the next command is processed we only need to know if we start from a "stop state" or from "resume state" (aka delayed teardown).

Regards.

lars18th:
Hi @catalinii ,

This PR is open for some time. I'm using it for a lot of time. Without it I can't use minisatip without glitches with the `satipc` module. And it only solves one particular case: when queueing a TEARDOWN command before receiving the previous response of a SETUP or PLAY that contains the SID value. So in any other case, if doesn't changes anything. It's a code for a particular case not handled at time. And from my point of view and IMHO I don't see any problem with the current async implementation. This handles this edge case without troubles and the code is clear.

However, if you have more questions to comment, please let me know.
I hope we can fix this problem (with this patch, or with another solution).
Regards.

lars18th:
Hi @catalinii ,

Please, can help with some comments/feedback?

src/satipc.c
catalinii: This is an example of deleting cached OPTIONS:
https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L276

lars18th:
ping to @catalinii ...

> Please, can help with some comments/feedback?

I'm running in my E2 boxes and server with this patch. And I feel it will be more easy/fast if you provide some implementation. Even if it's partial. The delayed handle is quite complex. Can you help please? I'm very busy at time.

Regards.

Jalle19:

lars18th:
Hi @Jalle19 ,

This PR requires to be open _until_ @catalinii will want to reimplement the `satipc` module. Without this patch, the module fails in the edge case described in the description. I'm using this patch in _all_ of my minisatip servers. And I feel other users are using it. The patch is valid and necessary. The only reason to not merge it, is because @catalinii thinks that it's necessary to implement it in a different way. 

So, please re-open it.
Thank you.

Jalle19:
@catalinii can we merge this then?

lars18th:
Hi @Jalle19 , My recommendation is to merge it, because it solves a real problem.

Hi @catalinii , If you don't want to merge it, then please reopen the PR. This will remember us that this edge case requires to be considered in your future reimplementation of the `satipc` module.

Regards.

Jalle19:
There's no point in keeping pull requests open forever, we can make an issue instead if we don't want to forget.

Jalle19:
I don't understand the issue here good enough to decide whether to merge or not.

catalinii:
I will re-write the satipc logic to not buffer responses which will fix this too.

lars18th:
> I will re-write the satipc logic to not buffer responses which will fix this too.

Great! 😄

lars18th:
Hi @catalinii ,

> I will re-write the satipc logic to not buffer responses which will fix this too.

You have completed this rewrite? I suffer a lot of troubles with linked minisatip instances (one or more servers driving the Tuner, and one as client with satipc... serving to other SAT>IP clients). And without this _ancient_ patch a lot of requests fail, and the client needs to retune with a channel change to overcome tha failed tunning.

Even this PR seems old, it solves a real problem. So if you agree, I can prepare and updated version and merge it. Because it's a simple change.
What you think?

Regards.

catalinii:
Yes the satipc in master does not buffer the responses anymore

PR #819, opened 2021-03-19
Enhanced drop of non-decrypted packets

lars18th:
When a packet can't be decrypted (and the drop encrypted packets is enabled) instead of filter out all encrypted packets do a more smart strategy. The implemented strategy consists of passing only packets with PCR time stamps. The passed PCR packets will don't have any payload data, so only the TS header is preserved. This will be more friendly for decoders that require the PCR marks. Futhermore the client will continue receiving packets, and then the stream can be considered active. Without this patch when a decoding disruption is a bit long several clients will close the session if it is not repaired in a short time.

In addition a second fix limits the wait time in the output buffer to resolve troubles when packets with PCR timestamps are not sended. This is necessary to flush the PCR packets when a temporary interruption of decoding happens.

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -104,7 +104,10 @@ static inline void mark_pid_null(uint8_t *b) {
 
 static inline void mark_pcr_only(uint8_t *b) { // Generate a clean packet with only the Adaptation field header and no payload
     if ((b[3] & 0x10) == 0) // No payload
-        return; 
+    {
+       mark_pid_null(b); 
+       return; 
+    }
 
     // Convert the Payload Data in Adaptation Stuffing
     int i;
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -88,9 +88,42 @@ static inline int get_adaptation_len(uint8_t *b) {
     return adapt_len;
 }
 
-void mark_pid_null(uint8_t *b) {
+static inline int get_has_pcr(uint8_t *b) {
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @catalinii ,

I hope you want to merge this patch. It solves the troubles with some clients that require the `-E` parameter. With this new implementation several of them no longer need to receive all packages (so we can discard the unencrypted). Therefore, the client can now continue without any problems, even with the missing packets. This new strategy maintains a minimum flush of packets and the synchronization of the clock. And after several tests almost all clients can work now without the `-E` parameter.

Please, try it and merge if you accept it!

lars18th:
Hi @catalinii ,

No comments about this patch? It resolves the troubles with a lot of clients that requires the `-E` parameter. It's clear, simple and documented. Why you don't check it?

catalinii:
Hey did not have much time free this week. The patch looks good.

Not sure why the content of the packet is marked as adaptation stuffing, but if it works is good.

lars18th:
> Hey did not have much time free this week. The patch looks good.

It works great for several weeks. Thank you for merging it! :wink: 

> Not sure why the content of the packet is marked as adaptation stuffing, but if it works is good.

The function removes ALL payload data inside the TS packet with the PCR timestamp. This is to not disturbing the decoder. So the only way to remove the payload data is to generate HEADER PADDING (it replaces the payload space). And this enforces to mark the header with Adaptation Stuffing.

Regards.

PR #820, opened 2021-03-20
Pass NULL packets when pids=all is emulated

lars18th:
When the pids=all is emulated it will be useful to receive the NULL stuff in some circunstances.
Instead of pass all NULLs by default, this is only activated when -E (pass unencrypted) is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,8 @@ Help
 	- note: * as adapter means apply to all adapters
 
 * -E Allows encrypted stream to be sent to the client even if the decrypting is unsuccessful
- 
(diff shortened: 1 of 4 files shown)


lars18th:
Hi @catalinii ,

This PR completes my previous (already merged) #816. Now, if you use one Enigma 2 box with minisatip and use it with `-E` (even without any decoding enabled), then you'll receive the full TS with a "pids=all" request (if the number of active pids is not longer than the hard limit of the hardware pid filtering). So with this PR you can convert any E2 box in a full TS streamer.

PR #821, opened 2021-03-20
Fix pids=all filtering encypted packets

lars18th:
When "pids=all" is running, then instead of filter out ALL encrypted packets you need to pass them. Without this patch, when not using the `-E` parameter (disabled by default) you receive all encrypted packets converted to NULL stuff. This modification solves this problem, and when you request the full transport stream (pids=all) then encrypted packets will pass untouched.

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -914,10 +914,17 @@ int pmt_decrypt_stream(adapter *ad) {
 
 void mark_pids_null(adapter *ad) {
     int i;
+    int pids_all = 0;
+
+    if (ad->drop_encrypted) { // Check (one time) if pids=all is activated
+        SPid *p = find_pid(ad->id, 8192);
+        if (p)
+            pids_all = 1;
+    }
     for (i = 0; i < ad->rlen; i += DVB_FRAME) {
         uint8_t *b = ad->buf + i;
         int pid = PID_FROM_TS(b);
-        if (pid == 0x1FFF)
+        if (pid == 0x1FFF || pids_all) // When pids=all don't drop encrypted packets
             continue;
         if ((b[3] & 0x80) == 0x80) {
             if (opts.debug & (DEFAULT_LOG | LOG_DMX))


lars18th:
Hi @catalinii ,

This patch has high priority! It fixes the use case when you request the full transport stream and the `-E` parameter is not set. Until now I use it every time, however after the PR #819 I don't need every time. So now I've discovered this bug: the client receives all encrypted packets converted as NULL when requesting for "pids=all". That's not the expected behaviour, as only when playing you need to filter out encrypted packets. With "pids=all" what you want is to receive the entire stream untouched.

So please, merge it soon.
Regards.

catalinii:
I am not sure this is the right way to deal with this issue.
For pids=all minisatip should attempt to deal with decrypting the packets. This PR prevents that.
I think the patch should be done on the cofe that changes the pid null

lars18th:
Hi @catalinii 

> I am not sure this is the right way to deal with this issue.

Why not? First this is a BUG. Until now every time I've used the `-E` parameter. And after changing to not use it appears the BUG: __ALL__ encrypted packets are converted to NULL packets. So, event if minisatip discards NULLs (in the input) in fact it is generating some of them. And for sure doesn't have any sense to transform encrypted packets in NULL packets without filtering them.

> For pids=all minisatip should attempt to deal with decrypting the packets. This PR prevents that.

Sorry? With "pids=all" you don't want at all to decrypt all services inside the TS (that's nearly impossible). What you want is to receive the full transport stream **untouched**. This PR si not preventing any action. What it's doing is _not clearing_ (translate to NULL) the encrypted packets that will never be decrypted as you have requested "pids=all". It's fixing an incorrect handle of packets. Nothing more! 

> I think the patch should be done on the cofe that changes the pid null

Why? I don't see any difference. When "pids=all" is set then the translation doesn't need to be applied. That's all.
The patch is absolutly necessary. I've not detected this before as for years I never have used minisatip without `-E`. But the bug is here now!
Futhermore, the code is optimized to only do the check one time outside the loop. So it's run fine.

Please try it! I feel you never use analytical tools with minisatip. And for this reason you don't use the "pids=all" function. However, it's very necessary to garantee a correct functionality. And at time minisatip is reporting a wrong behaviour with "pids=all" and the default configuration (that's without the `-E` parameter).

Sorry if I sound a bit rude, but this patch fixes a serious bug!
If your idea is to provide a method to decrypt all services inside a TS. Then I suggest to implement it after merge this PR. This will be _another_ different functionality. Perhaps it should be enabled optionally with a new command line parameter. In fact I think it could be interesting to see it.

Regards.

catalinii:
@lars18th there are 2 different use cases:
- decrypt all TS
- pass thru the entire TS without decryption

If you specify the -o option or own a CAM, you expect to decrypt everything.
Regardless of how those situations are "detected" 
the changes added should be mindful of all use cases.

We both agree that this is a bug, just my suggestion was to add it to the right place to prevent deleting in the future, which would make your work wasted.

IMHO I know if can be frustrating from your point of view, but if we do the things right there are few benefits:
- your code will stay unchanged
- you will learn better the code base and need less  comments on PRs in the future

lars18th:
Hi @catalinii ,

I've implemented the option `-o [~]host:port` in #928. So I close this PR. Please, merge the new version.

Regards.

PR #827, opened 2021-04-07
Update satipc_read() with recvmmsg()

lars18th:
When using the satipc module the reading of the RTP packets requires to be as fast as you can (without overloading the process).
Until now the UDP reading thread was reading only one packet at a time. So when receiving a chunk of packets it will be preferable to read as many as can be put into the read buffer. This new implementation uses the function `recvmmsg()` to just use one syscall to read more than one packet in one shot. The maximum number of packets to read is defined by `MAX_RTP_MSG`. The current value is sufficient to not generate oversubscription and to avoid delaying the packets too much when the bitrate rises a lot at some point in time.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -46,8 +46,20 @@
 
 #define TCP_DATA_SIZE ((ADAPTER_BUFFER / 1316) * (1316 + 16) * 3)
 #define TCP_DATA_MAX (TCP_DATA_SIZE * 8)
-#define MAX_RTP_MSG 25  // max number of UDP datagrams (with 1316 bytes of payload) to be read at one time
 #define MAKE_ITEM(a, b) ((a << 16) | (b))
+
+#if defined(__APPLE__) || defined(AXE)
+#define MAX_RTP_MSG 1   // read 1 UDP datagrams at one time (no recvmmsg() support!)
+int recvmmsg0(int sockfd, struct mmsghdr *msgvec, unsigned int vlen) {
+    if (vlen < 1)
+        return 0;
+    msgvec->msg_len = readv(sockfd, msgvec->msg_hdr.msg_iov, msgvec->msg_hdr.msg_iovlen);
+    return 1;
+}
+#define recvmmsg(a,b,c,d,e) recvmmsg0(a,b,c)
+#else
+#define MAX_RTP_MSG 25  // max number of UDP datagrams (with 1316 bytes of payload) to be read at one time
+#endif
 
 extern char *fe_delsys[];
 int satip_post_init(adapter *ad);
@@ -623,6 +635,8 @@ int satipc_read(int socket, void *buf, int len, sockets *ss, int *rb) {
     uint8_t buf1[20 * MAX_RTP_MSG];
     int i, rr, num_msg = 0, size_msg = 0;
     int force_ok = 0;
+    int check_holes = 0;
+    void *holes[MAX_RTP_MSG];
     struct mmsghdr messages[MAX_RTP_MSG] = {0};
     struct iovec iovs[(MAX_RTP_MSG * 2) + 1] = {0};  // RTP Header + Payload (up to 1316 bytes) for each UDP datagram
     uint16_t seq;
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

This PR supersedes #651. It includes all your suggestions, in particular: https://github.com/catalinii/minisatip/pull/651#issuecomment-570094778 .

Please, check it. :wink:

src/satipc.c
```diff
@@ -617,51 +618,120 @@ int satip_standby_device(adapter *ad) {
...
     }
-    *rb -= 12;
-    return (*rb >= 0);
+
+    size_msg -= (num_msg - rr) * 1316;  // update the theoretical total size of the read buffer
+
+    // Loop over all datagrams received
+    ad = get_adapter(ss->sid);
+    uint8_t *bf1 = buf1;
+    for (i = rr - 1; i >= 0; i--) {  // loop in reverse order (i = 0; i < rr; i++)
```
catalinii, line 674: Why looping in reverse order?
Will the rtp sequence number checking work in this case?

src/satipc.c
```diff
@@ -617,51 +618,120 @@ int satip_standby_device(adapter *ad) {
...
+            if (ad && sip->ignore_packets) {
+                force_ok = 1;
+                dlen = 0;
+            }
+
+        }
+
+        // Clear empty unused buffer space
+        if (dlen < iov[1].iov_len) {
+            int ss = (188 * (dlen / 188));
```
catalinii, line 719: If you read a packet of 12 bytes header + 188 bytes payload, this check will be false, which means the buffer will still contain holes. I would suggest doing a memset over the *rb as the cost is not big.

And of course fix the if condition.

catalinii:
@lars18th overall looks good with some comments.

What is not included in this PR is the emulation of recvmmsg on platforms that does not have it (AXE, MAC).
Please see implement it similarly to sendmmsg

src/satipc.c
lars18th: > Why looping in reverse order?

To be more efficient. When reading multiple RTP packets and pushing the PAYLOAD directly in the READ BUFFER, you can obtain HOLES in the buffer. So, if you're reading in reverse order then you only move valid data, and never any hole. The final result is identical, but in reverse order you copy less data if the number of holes is greater than one.

As a note, please take in consideration that a hole is when a RTP packet has a payload of less than 1316 bytes (that's less than 7 TS packets).

> Will the rtp sequence number checking work in this case?

Opps! In this point you're right. I need to revise this. Thank you for the tip.

src/satipc.c
lars18th: Sorry? This condition is right!
- `dlen` is equal to the PAYLOAD data, so the 12 bytes of the RTP header are out! See the line `dlen = msg->msg_len - 12;`
- `iov[1].iov_len` is equal to the space reserved for the PAYLOAD in the read buffer (in fact 1316 bytes).

The objective of this code is to remove the end holes of the slice if the input packets contain between 1-6 TS packets (less than the total of 7 TS packets). The strategy is then to remove the hole copying (aka moving) the data of the next slices over the last valid packet. My first implementation has to convert the free space into NULL packets, but this will interfere with the stream. So even it works I prefer to not use this approach. Now your suggestion is to memset this free space to `0x0` values? In this case the read buffer will be wasted with invalid data. Are you sure this will not generate troubles in upper layers? Futhermore, the wasted space can be important. For example, if you read 25 consecutive RTP packets with just 1 TS packet, then you're occuping 1316\*25=32900 (~32KB) for only 188\*25=4700 (~4.5KB) of valid data. You want this?

lars18th: Hi @catalinii ,

Using the technique of `memset()` the holes (fast and simple; but wasted space) the upper layer requires a simple modification. This line: https://github.com/catalinii/minisatip/blob/7ba80e07884466dfa63d97b50396b336eb9372af/src/stream.c#L909
requires a small change: For example to print only if the value is different from one indicating "internal NULL data".

So you agree with the idea of memsetting to 0xFF to indicate the "internal NULL stuff"? Then the previous line in the `stream.c` will continue discarding the TS packet but it doesn't print the error (as it's not an error).

You agree?

lars18th: Hi @catalinii ,

I'll implement this strategy:
- Process in forward order.
- Mark holes internally.
- After the end of the function I'll remove the holes (if exist) before returning.

I feel that's the best solution.
You agree?

lars18th:
> @lars18th overall looks good with some comments.

Great!

> What is not included in this PR is the emulation of recvmmsg on platforms that does not have it (AXE, MAC).
> Please see implement it similarly to sendmmsg

In this case, I think I'll not implement like with sendmmsg... Why? Because this is for receiving. So a simple loop will block. And I don't want to check the number of pending UDP datagrams as this is a kernel optional function that perhaps is not available for the platform. So if you agree, then I'll maintain the original function, and based on the platform one or other will be used. You agree?

src/satipc.c
lars18th: Hi @catalinii ,

This is solved. Now processed in forward way.

src/satipc.c
lars18th: Hi @catalinii ,

The new implementation uses the three steps commented: process in forward order, mark holes, post-process holes.

lars18th:
Hi @catalinii ,

All is fixed and included: forward processing and emulation of recvmmsg().
So the PR is ready to merge... after you test it! :wink: 

Regards.

catalinii:
Thanks for the patch. As long as it works on those platforms I am fine

PR #828, opened 2021-04-08
New PIDS_ALL emulation with NULLs

lars18th:
Instead of using two list, this code now adds the NULL (8191) pid when it's necessary.

This partially rollbacks the PR #820. The problem is created with some compilers that doesn't calculate the sizeof of the pointer when assigned an array. This new implementation is more simple.

### DIFF
--- a/src/minisatip.h
+++ b/src/minisatip.h
@@ -12,7 +12,6 @@
 #endif
 
 #define EMU_PIDS_ALL_ENFORCED_PIDS_LIST 1, 16, 17, 18, 20, 21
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @catalinii ,

Please, merge this patch as it fixes the changes introduced in https://github.com/catalinii/minisatip/commit/c750f4c696208c6b7bb82d4b9553ade7a377c5e0 .
The change compiled in Enigma2 doesn't work. So this new implementation solves the problem.

Regards.

Issue #811, opened 2021-02-27
Help: How to configure two tuners in master-slave


lars18th:
Hi,

Until now I'm using minisatip to drive one or two _independent_ SAT>IP servers. However, now I need to configure two tuners in a master-slave mode, and I need some help.

That's the configuration:
- 4 orbital positions with one DiSeqc 1.1 swith.
- 1 coax cable.
- 1 passthrought 2x1 derivator.
- 1 SAT>IP server (1 tuner) connected as regular 13-18v with 22KHz (master).
- 1 SAT>IP server (1 tuner) connected without any output current (slave).

Then the slave tuner can only work if the first is running and only consuming the same band in the same satellite position. In this scenario: What's the correct configurition for the minisatip parameteres?

I understand that the minisatip will use every time the master tuner, and only if it's in use and a second request is in the same band/position, then it starts the slave tuner. However, in case of running both tuners, if the session that uses the master tuner closes... what does the minisatip logic? It continues using the master tuner (perhaps with a fake `pids=0`) until the second closes?

Please, help me with this configuration.
Thank you!

catalinii:
In case the stream on the master iadapter is closed, minisatip will keep the frontend enabled with no pid filters to allow the slave to continue.
You can use https://github.com/catalinii/minisatip/blob/master/src/minisatip.c#L452 as an example

lars18th:
Hi @catalinii ,

> In case the stream on the master iadapter is closed, minisatip will keep the frontend enabled with no pid filters to allow the slave to continue.

OK, so then it will work as expected. However, I suggest to change the keeping of the frontend enabled with no pid filters to something different... when using the `satipc` module. What you want to say with "no pid filters"? This implies `pids=all` or `pids=none`? From my point of view the most safe way is to use `pids=0` without changing the original parameters of the original session in the master tuner. The reason is because if the tuners are SAT>IP servers and we're using the `satipc` module, then is a waste of resources to use `pids=all`; and `pids=none` will generates troubles with some servers.

So, in short: This master-slave mode works with the `satipc` module without troubles?

catalinii:
Generally you need to specifically add the pid filters. In this case no pid filter is added.

lars18th:
> Generally you need to specifically add the pid filters. In this case no pid filter is added.

Then, the master tuner will run with `pids=all` until the slave closes?

catalinii:
Will run with no pid filter (not receive aby data)

lars18th:
[...]

PR #830, opened 2021-04-09
Search for slave adapters

lars18th:
This fixes the problem when searching for free adapters. Before this patch no adapters are found when all master tuners are in use and slave adapters are free. This patch solves this problem searching over the slave adapters after the initial search. Now the slave adapters will be used when needed.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -712,6 +712,8 @@ int get_free_adapter(transponder *tp) {
         LOG("get free adapter %d msys %s requested %s", fe, get_delsys(fe),
             get_delsys(msys));
 
+    dump_adapters();
+
     if (fe >= 0) {
         if (ad && delsys_match(ad, msys)) {
             match = 0;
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

This PR solves the first problem described in #811 .
Please, try it and merge.

Note: It not solves the secondary problem (more complex to fix). This other problem is when the slave tuner has _no power_ and requires that the master tuner be active to inject it. As you can imagine this is more hard to solve. However, I'm sure this PR will solve the problem for users that have master-slave tuners but the slave tuner provides power to the LNB. In this case this patch solves all problems.

lars18th:
Hi @catalinii ,

Please, can you check and merge this patch? To resolve the secondary problem I've an idea. But, I need to discuss it with you, but first it's a requirement that the selection of slave tuners is resolved (that is the target of this patch).

So, please check it!

catalinii:
Merged thanks

PR #835, opened 2021-04-20
Flush output in some necessary cases

lars18th:
This patch incorporates these to enhancements:

- SATIPC: When the reading of the incoming RTP packets is full, then it will be a high probability that there are still pending packets in the buffer. To improve the efficiency, flush the output buffer. This increases the performance. So not needed an aggressive `-H 1:1` command line option.
- PMT: When converting undecrypted packets to PCR only packets, it's mandatory to flush the output to send the corresponding TS packet with the PCR mark at time, without waiting to pack them with next packets. Without this, the player can receive inconsistent PCR timestamps joined.

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -967,7 +967,7 @@ void mark_pids_null(adapter *ad) {
                 if (get_has_pcr(b)) // It has PCR
                 {
                     mark_pcr_only(b);
-                    ad->flush = 1;
+                    // ad->flush = 1; // Not necessary as process_packets_for_stream() flush all packets.
                 }
                 else
                     mark_pid_null(b);
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -965,7 +965,10 @@ void mark_pids_null(adapter *ad) {
             {
                 // Instead of remove ALL packets, when the packet has a PCR remove all payload and pass it
                 if (get_has_pcr(b)) // It has PCR
+                {
                     mark_pcr_only(b);
+                    ad->flush = 1;
+                }
                 else
                     mark_pid_null(b);
             }
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -684,6 +684,14 @@ int satipc_read(int socket, void *buf, int len, sockets *ss, int *rb) {
 
     size_msg -= (num_msg - rr) * 1316;  // update the theoretical total size of the read buffer
 
+    // It checks if the read is full, because in that case there is a high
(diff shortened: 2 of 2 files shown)


lars18th:
Hi @catalinii ,

Please, check it! I'm using it. So I suggest to merge it. It improves the stability of the player (and performance) when using the `satipc` module.

Regards.

src/pmt.c
```diff
@@ -965,7 +965,10 @@ void mark_pids_null(adapter *ad) {
             {
                 // Instead of remove ALL packets, when the packet has a PCR remove all payload and pass it
                 if (get_has_pcr(b)) // It has PCR
+                {
                     mark_pcr_only(b);
+                    ad->flush = 1;
```
catalinii, line 970: This is not really needed as at the end of stream processing the stream will be sent including the pcr pids.

lars18th: Hi @catalinii ,

Sorry, I don't agree. I've done a lot of testing to complete this patch. The problem is this: when the non-decrypted packets are removed, if you create the PCR only packet, it will wait in the output queue until some more packets come to the output buffer. But then, if the PCR packet waits in the queue then it doesn't resolve anything. So the PCR packet requires to be sended _in time_, or from the client point of view you will see one UDP packet with 7 PCR packets. Therefore, in the time interval inside _all_ packets are removed, except the PCR one, every packet (aka PCR) requires to be flushed.

Note this: When the `mark_pcr_only()` function is called, only 1 packet (the PCR) will output. And it requires to be flushed. This is mandatory in order for the client to remain synchronised.

So, please merge the patch.
Regards.

catalinii: mark_pids_null is called by the pmt_process_stream which is called by process_dmx.
In process_dmx, after pmt_process_stream, process_packets_for_stream is called which flushes all the packets from adapter buffer to the network.

lars18th: Hi @catalinii ,

Thank you! This time you pointed very well the question.
My concern has with this line inside `read_dmx()` (what's the function that calls to `process_dmx()`:
https://github.com/catalinii/minisatip/blob/261069a4bc7daaaac768a66926bf75ac87bf77c8/src/stream.c#L1083

However, as you pointed, it's _after_ that the pending packets are flushed.
Then I remove this line as it's not necessary. Even I leave a comment. :wink:

lars18th:
Hi @catalinii ,

Ready to merge?

catalinii:
Yep

PR #838, opened 2021-04-21
Added new "Service Name" option

lars18th:
Added a new "name_app" option to change the announced name of the server (aka Service Name).

By default "minisatip" is the name of the server, but with this new option it could be set to any name. This is useful to use more than one servers in the same LAN and distinguish them. The optional name is used only for advertaising, so the device continues to be "minisatip". Any client that uses the name of the server to identify it in the LAN will use the Service Name, even if the IP address changes. So use it carefully. However, any tool that requires to know if the server is "minisatip" will continue working as expected as only the advertised name is changed.

Useful for example when using several instances with clients like VideoLAN with the UPnP autodiscovering option.


lars18th:
:+1:

PR #858, opened 2021-06-20
add support for binding to an address and/or an interface

ddimension:
-V --bind address: address for listening (all services)
-J --bind-dev device: device name for binding (all services)
 beware that only works with 1 device. loopback may not work!

Please notice that binding the interface is especially useful in multihomed setups.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -30,6 +30,7 @@
 #include <getopt.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <net/if.h>
 #include <signal.h>
 #include <stdarg.h>
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


ddimension:
Switched from -V to -J option. There was double usage.

lars18th:
Hi @ddimension ,

I suggest (if you want to do it) to add these new values to the UI.
If you want to do it, then you only need to change two files:
- To export the values here: https://github.com/catalinii/minisatip/blob/3f2f0425970823039ab0620d2399d1faf8521019/src/minisatip.c#L2001
- And to print the values here: https://github.com/catalinii/minisatip/blob/3f2f0425970823039ab0620d2399d1faf8521019/html/status.html#L319

Thank you for your contribution.

catalinii:
We can address this as part of different PR. I will merge this one.

PR #860, opened 2021-06-22
Coverity changes

9000h:
some small fixes


9000h:
look like the §TOKEN is empty for the scan

catalinii:
Yeah I think the same

PR #867, opened 2021-06-28
Check previous key even it's expired

lars18th:
When searching for a new key, it's a good idea to check the previous one even if it's expired. It has a high probability of still being valid.

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -686,8 +686,8 @@ void update_cw(SPMT *pmt) {
     for (i = 0; i < ncws; i++)
         if (cws[i] && cws[i]->enabled && pmt->parity == cws[i]->parity &&
             cws[i]->pmt == pmt->id) {
-            if (ctime > cws[i]->expiry)
-                continue;
+            if (ctime > cws[i]->expiry && (!old_cw || (old_cw && old_cw->id != cws[i]->id)))
+                continue;  // Don't check old keys except if it's the previous one.
             if (len && !test_decrypt_packet(cws[i], start, len)) {
                 LOGM("correct CW found (len %d): %s", len,
                      cw_to_string(cws[i], buf));


catalinii:
@lars18th not sure if this is the right fix.
Basically there is odd and even key. The old_cw for a even_key is the odd key not the previous even key.

In your message:
found CW: -1 for TEST-SVR PMT 1, old cw 3

Basically you expect to find CW 2 not CW 3.

Of course the simplest fix here is to just increase the expiry of the key. Let me know if that fixes your issue.

lars18th:
Hi @catalinii ,

> @lars18th not sure if this is the right fix.
> Basically there is odd and even key. The old_cw for a even_key is the odd key not the previous even key.

Yes, I know it!
 
> In your message:
> found CW: -1 for TEST-SVR PMT 1, old cw 3
> 
> Basically you expect to find CW 2 not CW 3.

Yes, because the CW 2 is the EVEN (the active key until now), the CW 3 is the "new" ODD and the CW 1 is the "old" ODD.
Please, take note that the use case that fails is (in reference to the _parity_) is this:

- `00000000111111111000000000101111111100000000011111111`

You can see that here the current implementation fails with the change `...000_1_0_111...` when the EVEN key ("0") is experired.

> Of course the simplest fix here is to just increase the expiry of the key. Let me know if that fixes your issue.

To be clear: I do a lot of changes to enlarge the expirity value (you know it for a long time), and any of my fixes related to this has approved. Futhermore, they in fact not resolve the root cause target here. The root problem is this:

- When a key change is detected you _automatically_ change the key for the entire service. However, it's possible that at the point of the key change _some_ pids will use different parity. From my point of view the best solution has been to maintain a pair of active keys: even and odd, and not only one. Then it will be possible the use both at the same time. However, your current implementation works; so it's best to not change it. But it's necessary that you _NEVER_ invalidate the checking of the `old_cw`; bacause it could be already in use by a different pid. Therefore, this PR fixes the problem by ensuring that the `old_cw` will be included in the check list (this key is the previous active with the different parity).

Please accept this PR now and if you want to improve the code in the future, please do so.
Regards.

catalinii:
@lars18th the only invalidation currently is the one caused by expiry.

Generally the decryption is done for a set of packets, where for each set of the same parity the key is checked against the not expired CWs.

Basically a CW that did not expire can be used only if it is proven to decrypt correctly the  stream.

There is some caching as well (if nothing changes, a new CW will not be searched) but that does not come to play.

In your case is likely that the encrypted packet cannot be verified (no PUSI present)  so is better to enable the debug logs for pmt and try to reproduce.

lars18th:
Hi @catalinii ,

Please, belive me: I've checked and tested searching for this error for a long time! But until now I don't see that sometimes the key change is NOT done at the same time in all pids. Futhermore, it's true that not every time the key change is done when a PUSI packet appears. This differs for some operators to other based on the different CA technologies. So in fact the problem is this: when you change to a new key, because _in some pid_ of the service the parity is changed, you assume that the entire key changes. Then if another pid continues using the old parity is then when the problem appears. That's because in this case you're assuming these two postulates:
- The key with the old parity is not expired.
- When the parity will change in other pids it will pass only in a PUSI packet.

However, are you sure this is true every time?
Regarding the expiration please remember my suggestion about increase the value years ago.
And with regard to the second, I'm not sure that some CA system with not do the parity change before/after a PUSI packet.

So, IMHO the current code requires a refactoring. But instead of a full change of the current implementation, perhaps the most simple (if you agree) is this addressing only the edge case:
- When there is a change of parity and the old key of the previous parity is still in use (in the same or in another pid), in this case not do any key change or testing, but use directly the old key (because in fact it will be the correct one).
 
What you think? I feel this could be a simple and straightforward solution.
Regards.

catalinii:
I do not make the assumption that PUSI arrives when the parity changes. But if it does not arrive (which is the case most of the time), is impossible for your case to check which key is correct. So at that point is an assumption that the last CW will work, which may not be true for your use case.

if you want we can discuss this over slack.

lars18th:
Hi @catalinii ,

Sorry, but I feel you don't have understand the described USE CASE. So, I repeat the key change with a more detailed diagram:

```
Video PID: 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 ...
Audio PID: 0 0 0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 ...
Aud-2 PID: 0 0 0 0 0 0 0 0 1 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 ...
```

In this case with the current code, if the key with even parity (0) has expired, the decoding fails. The problem appears with the audio packets with the even parity after the parity change to odd in the video pid. The reason is becuase the function `update_cw()` doesn't found any key. And it doesn't found it because the code automatically discards the key 0 only for the time passed. Futhermore, this function could also fail because it calls to `test_decrypt_packet()` that it will fails too if no PUSI packet are found in the pending buffer.

You agree with that, or I'm missing something?

So if you agree, then a very simple solution is that: After the change 0->1 detected in the video pid, the new key is the odd key (1). Until here nothing to change in the current implementation. But, for any packet with even parity (0) between the transition period, instead of searching and testing, it's sufficient to directly use the `old_cw`. It's not necessary to check anything, this is the key. It's impossible that this key (the `old_cw`) will not be the correct key. And regarding the transition period, it's not necessary to make any assumption. The period will cover from the key change until the next key with different parity will arrive. So in fact, you can decode with the `old_cw` (for the oposite parity of the current key) until a new key arrives with the same parity of the `old_cw`. Here the expiration time has no sense, as the transition period could be defined in a deterministic way.

Therefore I'm sure this can solve the problem. And I hope you agree with this approach. It's more clear now?
Regards.

catalinii:
@lars18th I agree with the problem statement.

There 2 concerns I have:
- the fix may or may not be right 
- in the future the code around decryption may change so this case may be broken again.

My suggestion is to have unit testing using the case you are describing to make sure this is covered in future versions.

Btw, on which channel is this happening?

lars18th:
Hi @catalinii ,

I agree with the idea of one unit test for thsi use case. However, until we can implement it, why not add a simple change of enlarging the expiration time? So instead of...
https://github.com/catalinii/minisatip/blob/1d353862a63325a74cf0a8dca05f1d22f2ce8ae1/src/pmt.h#L21
we can use 45 seconds... `#define MAX_CW_TIME 45000 // 45s`

And after, we can preprare the unit test. My suggestion is to prepare a home made stream (from the BBB free video) encrypted manually. The service will have 3 streams (one video, two audios), and we can use the three potential key changes in each stream:
- In the video the key change is done in a packet with a PUSI mark.
- In the first audio the key change is done only in a packet without a PUSI mark, and after the key change in the video.
- And in the second audio the key change is done not only without the PUSI mark and after the video, but also in a flip-flop mode (like EVEN-ODD-EVEN-ODD...).

If you want I can prepare this testing stream and you can prepare the unit test code. You agree?
Regards.

lars18th:
Please, @catalinii merge the PR #880

> I agree with the idea of one unit test for thsi use case. However, until we can implement it, why not add a simple change of enlarging the expiration time? So instead of...
> https://github.com/catalinii/minisatip/blob/1d353862a63325a74cf0a8dca05f1d22f2ce8ae1/src/pmt.h#L21
> 
> we can use 45 seconds... `#define MAX_CW_TIME 45000 // 45s`

catalinii:
I did, can we close this PR ?

lars18th:
> I did, can we close this PR ?

OK. However, the problem is here. Although I'm still trying another solution to the poblem.
Regards.

PR #878, opened 2021-07-03
Coverity disable false positive utils.c

9000h:
// coverity[use : FALSE] can disable false positive warnings


9000h:
or is providing a modeling.c better with content like
```
long random(void) {
  /* ignore */
}
```

PR #881, opened 2021-07-05
fix dad specinv dvb-c

9000h:
ver=1.2 DVB-C need specinv not inversion


retrofan1979:
O thanks, will do this evening.

PR #882, opened 2021-07-06
correct fe_specinv values

9000h:
correct fe_specinv values to use the right values for dad


9000h:
any issues with this, or would you prefer to revert https://github.com/catalinii/minisatip/pull/881

catalinii:
fe_inversion and fe_specinv seems so be duplicate. We can do fe_inversion and remove specinv everywhere

9000h:
I hope all is fine now.

Issue #879, opened 2021-07-04
Force setting for QAM


retrofan1979:
Hi,

I cannot get the Android app Aurora TV work with my Fritz WLAN repeater DVB-C when using minisatip. I think it's because the app doesn't ask for QAM parameter.

https://play.google.com/store/apps/details?id=com.futarque.amoeba

Is it possible to force this setting with an option for minisatip? Maybe we can try with a beta if this is indeed the problem? Thanks.

9000h:
it's probably an outdated/buggy app an no development anymore, go for something better like the
https://play.google.com/store/apps/details?id=de.avm.android.fritzapptv
which works direct with your fritz dvb-c repeater or if there is a need to use minisatip
https://play.google.com/store/apps/details?id=com.cmuv.SatIPViewer

retrofan1979:
Thanks, but my Panasonic TV does work with Aurora TV, so I don't understand why the Fritz and minisatip don't. I'm curious about it, that's why I'm asking.

9000h:
I'm confused now Aurora TV is a client, how should it a Panasonic TV use it as server?

retrofan1979:
My Panasonic TV has feature to act as SAT>IP server.

9000h:
ok got it, the Fritz SAT>IP implementation is a bit non standard and it's a non certified server, minisatip should probably work but as it's very old and no support anymore it's maybe wast of time.

9000h:
Aurora TV seem to work for DVB-S/S2 at least with minisatip, on DVB-T2 the scan did find nothing and a DVB-C scan did crash the app
did try the with minisatip (DVB-T2/DVB-C) and a Panasonic TV (DVB-C) as server on a Samsung Tablet

retrofan1979:
Okay, thanks for testing. Even more interesting the Panasonic succeeds to find DVB-C channels. Is my guess correct about QAM parameter?

9000h:
I'm unable to help as the app did crash on my Samsung tab when I try to do a scan on DVB-C.

retrofan1979:
Hi, I also tried this HDHomeRun Proxy, but this one also fails, but it claims it works with minisatip as well.
Don't know what the problem is. 474 MHz is the homing channel.
URLGet is the URL for minisatip.

https://github.com/dsaupf/HDHRProxyIPTV

I tried this setting in mapping .INI file

###########################################
# Description tags that may contain this file:
#
# [MAPPING_LIST]
[...]

PR #887, opened 2021-07-19
satipc quirk for Gainatech EyeTV Netstream 4C

9000h:
Gainatech Netstream 4C did reply  with 503 when fe=1 is in the  request URL
fixes https://github.com/catalinii/minisatip/issues/879

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -204,7 +204,7 @@ int satipc_reply(sockets *s) {
     if (rc == 404)
         sip->restart_needed = 1;
 
-//quirk for Gainatech Netstream 4C when fe=1 is in the URL
+//quirk for Gainatech EyeTV Netstream 4C when fe=1 is in the URL
     if (rc == 503)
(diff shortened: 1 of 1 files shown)


retrofan1979:
Sorry, but it's not Gainatech but Geniatech.

retrofan1979:
Topic 879 is closed, but what about the quirk for Aurora TV (QAM AUTO not allowed when mtype not specified in URL) and Fritzbox? (Error 408)

Personally I think it's better not to use quirks at all, but just deleting parameters in request URL which are not allowed for DVB-C (and DVB-T) as mentioned in the SAT>IP 1.2.2 specifications. And adding mtype when not in request URL as this one is mandatory!

Undo this change?

retrofan1979:
Sorry, but it's not Gainatech but Geniatech.

retrofan1979:
Sorry, but it's not Gainatech but Geniatech.

retrofan1979:
Thanks for this fix!

PR #915, opened 2022-01-14
Fix clean_psi() call when processing the PMT

### LINKED ISSUE #340: Bug: "-t --cleanpsi" parameter don't work
Hi @catalinii ,

In current releases the parameter "-t --cleanpsi" doesn't work!

I check it in this way:

- I open one client using the http interface (more simple to check the parameters). For example with pids "0,17,18" & PMT, Video/Audio, EMM of the channel. This starts to decrypt and play it.

- I open a second client that shows the PAT/PMT tables (TSReader, TransEdit, etc.) over the same frequency and/or the same program. The result is: The program continues to be signalized as encrypted.

Please, can you review the code?
Thank you!

lars18th:
When calling to clean_psi() the pointer is the start of the PMT table and not the start of the TS table. So it's required to pass the pid value.
Fixes #340

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -1246,8 +1246,8 @@ int clean_psi_buffer(uint8_t *pmt, uint8_t *clean, int clean_size) {
     return nlen + 3;
 }
 
-void clean_psi(adapter *ad, uint8_t *b) {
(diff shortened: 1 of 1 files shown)


catalinii:
Thanks @lars18th

Issue #911, opened 2021-12-28
[Q] Signal SNR in dB ==> Improve percentage conversion


lars18th:
Hi @catalinii ,

In my E2 STB I discovered that the SNR value that gets the minisatip is not consistent with the one reported by the tuner.

For example I use `--multiplier 0:1.0-2.0`, so the SNR value is duplicated by minisatip. And the behaviour is correct, as the UI of minisatip prints the double of the SNR value readed. However, I feel the value used is incorrect. Ler to explain why I think this. I've executed the `dvb-fe-tool -a 0 -g -m` command and I discovered why the SNR value is so low. Please see the output:
```
Lock   (0x1f) Signal= 89,00% C/N= 12,16dB
          Layer A: C/N= 82,00%
```

When the tool prints this, the UI of the minisatip shows 24% of SNR and 89% of STR. But why?
Because SNR is ~12*2. So if you have noted what the driver reports, the value used for SNR it's in **dB**, and not in percentage.

Then, I suggest a improve the SNR calculation in case the driver reports it in dB.

Regards.

catalinii:
Hey,

The satip standard requires the values to be relative (in percentages) rather than DB* values.

You can see here how they are converted: https://github.com/catalinii/minisatip/blob/master/src/dvb.c#L1392

If you have a better suggestion than what the current code has, let me know

lars18th:
Hi @catalinii ,

> If you have a better suggestion than what the current code has, let me know.

So, yes I've one! 😉 

I do some research and here are my results:

- When tuning one transponder (in the E2 STB) while executing minisatip with debug enabled for the dvb module _and_ with the dvb-fe-tool running as well, I see these values:
```
[29/12 08:52:20.984 signal]: get_signal_new adapter 0: status 31, strength 0 % -> 59636, snr 12050 dB -> 7896, BER: 0, err 0

Lock   (0x1f) Signal= 91,00% C/N= 12,05dB
          Layer A: C/N= 82,00%
```

[...]

PR #914, opened 2022-01-13
progress-bar-rewrite

kellerman:
Progress bars rewritten to a pure HTML/CSS solution.
pBar jQuery plugin removed.
Added dB display over SNR progress bar.
Added animation for the progress bars when the table is being updated.
Color scale depending of the value (5 colors)

Preview https://i.imgur.com/tDuWPJc.png
Open to any modifications to colors, borders, brightness, opacity etc...

PR #918, opened 2022-01-20
WEB: New progress bar & dB with DVB adapters

lars18th:
This PR fixes and improves #911. It's a teamwork of @kellerman and me.

The Progress Bar has changed in the UI with new format. Futhermore, it prints the real dB value if the adapter provides it.
Implemented the get of the dB value for DVB adapters that report in this scale.

### DIFF
--- a/html/status.html
+++ b/html/status.html
@@ -366,18 +366,21 @@
 						var snr = parseInt(state['ad_snr'][i] / 2.55);
 					else
 						var snr = parseInt(state['ad_snr'][i] / 0.15);
-						
-						var snrdb = "16.5"; //ad_db value needs to go here
-					
-						//SNR progressbar color scaling
-						var snr_color;
-						if (snr<=25 && snr>=0) snr_color="progress-1";
-						if (snr<=42 && snr>25) snr_color="progress-2";
-						if (snr<=60 && snr>42) snr_color="progress-3";
-						if (snr<=80 && snr>60) snr_color="progress-4";
-						if (snr<=100 && snr>80) snr_color="progress-5";
+						var snrdb = parseInt(state['ad_db'][i] / 0.1);
 
-						signal += "<div class='level'><div class='db'>" + snrdb + "dB</div>SNR <progress class='psnr " + snr_color + "' value='" + snr + "' max='100' />" + snr + "%</div><br />";
+						if (snrdb >= 65535) // "65535" is the MAX_DB value
+							signal += "<div class='level'>SNR <progress class='psnr " + snr_color + "' value='" + snr + "' max='100' />" + snr + "%</div><br />";
+						else {
+							//SNR progressbar color scaling
+							var snr_color;
+							if (snr<=25 && snr>=0) snr_color="progress-1";
+							if (snr<=42 && snr>25) snr_color="progress-2";
+							if (snr<=60 && snr>42) snr_color="progress-3";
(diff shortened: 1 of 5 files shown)


lars18th:
Supersedes #914.

html/status.html
```diff
@@ -229,22 +351,44 @@
...
 					if (ad_type == 1 || ad_type == 3)
 						var snr = parseInt(state['ad_snr'][i] / 2.55);
 					else
 						var snr = parseInt(state['ad_snr'][i] / 0.15);
-					var snr0 = Math.max(0, snr - 10);
-					signal += "<div class='level'>SNR <div class='pBar' data-from='" + snr0 + "' data-to='" + snr + "'></div> " + snr + "%</div><br />";
+						var snrdb = parseInt(state['ad_db'][i]);
+
+						if (snrdb >= 65535) // "65535" is the MAX_DB value
+							signal += "<div class='level'>SNR <progress class='psnr " + snr_color + "' value='" + snr + "' max='100' />" + snr + "%</div><br />";
```
catalinii, line 372: Can we not set a value if the SNR is not reported ?

src/dvb.c
```diff
@@ -1700,9 +1700,10 @@ void get_signal(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
 // returns the strength and SNR between 0 .. 65535
 
 int get_signal_new(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
-                   uint16_t *snr) {
+                   uint16_t *snr, uint16_t *db) {
 #if DVBAPIVERSION >= 0x050A
     *status = *snr = *ber = *strength = 0;
+    *db = 65535;
```
catalinii, line 1706: use MAX_DB

catalinii:
I would suggest using value 0 instead of 65535 for situation where the signal cannot be read.
By default the adapter vlues are all initialized to 0, and there are other cases such as satip adapters, so the ad->db value is not set to 65535. In that case the code may not work properly.

html/status.html
lars18th: You say "SNR" or "dB" ?
The code does this:
- When dB is 65535, then we don't have the value. So it isn't printed.
- The default value (in the code) for all adapters are 65535. So, only if a value is reported it will be printed.
- When no SNR is reported the default value is 0. So this value is printed (with the progress bar at zero value).

The only case not considered here is when the "status.html" page gets the "status.json" and the `ad_db` parameter is not reported. In this case, we can assume that the value will be 0. Therefore a "0.0 dB" will be printed. But this is an internall error, because minisatip will exports the `ad_db`. So then we don't need to worry about this extreme case. Right?

src/dvb.c
lars18th: Hi @catalinii ,

> use MAX_DB

To do this I need to add `#include "adapter.h"` in "dvb.c". My idea is not to smudge the code. But if you prefer to do this change, then please agree with that.

lars18th:
> I would suggest using value 0 instead of 65535 for situation where the signal cannot be read. By default the adapter vlues are all initialized to 0, and there are other cases such as satip adapters, so the ad->db value is not set to 65535. In that case the code may not work properly.

This is false. I considered this case. All tuners are initialized with the 65535 value for the dB. See the line 281 in the `int init_hw()` function. 😉

PR #920, opened 2022-01-21
Update README (and HELP)

lars18th:
Some cosmetic changes in README and HELP.
Added new documentation about char `~` with the option `-U --sources`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,13 @@ https://paypal.me/minisatip
 Usage:
 -------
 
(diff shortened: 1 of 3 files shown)


catalinii:
@lars18th there is tools/gen_readme which will generate the readme based on minisatip code.

Please use it to update the readme

lars18th:
> @lars18th there is tools/gen_readme which will generate the readme based on minisatip code.
> 
> Please use it to update the readme

Done! Please commit #926 and autogenerate the new README.
Regards.

PR #925, opened 2022-01-26
Add Dark Mode and update logos

bccrew:


bccrew:
I added a dark mode like seen in the screenshot below ;)

image

html/status.html
```diff
@@ -606,6 +659,9 @@
 			getData();
 			reloadPage();
 		});
+
```
catalinii, line 662: LGTM. Can you remove those lines please ?

bccrew: @catalinii done :)

PR #928, opened 2022-01-27
Optional pids=all filtering encypted packets (v2)

lars18th:
This patch incorporates a new parameter in the oscam option to disable the default behaviour of filter out all packets when `pids=all` is requested. With this option on (disabled by default) when `pids=all` is running, then instead of filter out all encrypted packets they are passed. Without this patch, when not using the -E parameter (disabled by default too) you receive all encrypted packets converted to NULL stuff. This new optional parameter solves this problem, and when you request the full transport stream (pids=all) then encrypted packets will pass untouched.


lars18th:
Supersedes #821.

lars18th:
Hi @catalinii ,

Ready to merge! 😉

PR #936, opened 2022-02-08
Change PCMCIA speed to 96 Mbps whenever a CI+ CAM is inserted

Jalle19:
I stumbled upon a CAM that always ended up taking the "ai_version == 1" path, meaning the CAM data rate would never get bumped.

The spec says:
> CICAMs shall support 96 Mbit/s. Hosts shall support 72 Mbit/s, support for 96 Mbit/s is optional.

...so it should be safe to always use 96 Mbit/s for CI+ CAMs.

There's another tiny change that I couldn't easily split to a separate commit and that is to log that the "date time" resource is connected.

### DIFF
--- a/src/ca.c
+++ b/src/ca.c
@@ -2913,6 +2913,7 @@ static int ca_lookup_callback(void *arg, uint8_t slot_id,
                               void **arg_out, uint32_t *connected_resource_id) {
     ca_device_t *d = arg;
     d->session->ca = d;
+    *connected_resource_id = requested_resource_id;
 
     LOGM("===================> %s: slot_id %u requested_resource_id %x",
          __func__, slot_id, requested_resource_id);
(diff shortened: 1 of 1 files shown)


Yuri666:
you false. check log

[08/02 23:09:37.625 CA0]: --------------------S_SCALLBACK_REASON_CAMCONNECTED---------APP_AI_RESOURCEID-------------------------
[08/02 23:09:37.626 CA0]: setting CI+ CAM data rate to 96 Mbps

Jalle19:
I'm not getting that line in my logs without this change.

src/ca.c
```diff
@@ -1123,11 +1121,6 @@ static int ciplus13_app_ai_data_rate_info(ca_device_t *d,
                                           ciplus13_data_rate_t rate) {
     uint8_t data[] = {0x9f, 0x80, 0x24, 0x01, (uint8_t)rate};
 
-    /* only version 3 (CI+ 1.3) supports data_rate_info -  no it isn't, 1.2
-     * support too*/
-    if (d->ca_ai_version < 2)
-        return 0;
```
Jalle19, line 1129: My CAM ended up here

Yuri666:
so, your CAM not support 96 mbps. I have such CAM. It ci+ 1.1

Jalle19:
Well, it still works even when I force it to 96 Mbps 🤷 And the spec says it's mandatory for any CI+ CAM (unless that's a more recent addition).

Yuri666:
Ok, probally you right. date_rate APDU just inform CICAM about supported data rate by HOST.
HOST always support 96 mbps, so we shall always to send 96 mbps APDU to ci+ CAM.

Jalle19:
Yeah that's my thinking as well. I don't know what circumstances could cause the host to _not_ support 96 Mbps? Maybe some really old laptop with a native PCMCIA slot? I think it's safe to assume users use DD, TBS or Enigma STBs, so full speed should be taken for granted.

@Yuri666 can you try this with your 1.1 CAM?

Yuri666:
> @Yuri666 can you try this with your 1.1 CAM?

No difference behavior between 72 and 96 mbps. Such as expected - this CAM can decode only one channel, so 72 mbps is enough at any way and don't matter what speed support host.

Jalle19:
Alright, thanks for testing! Do you think this is okay to merge?

Yuri666:
yes, all ok

Yuri666:
Ok, I re-read the spec again. Application information resource version less than 3 don't support data_rate_info APDU. It may hang some CAMs. So, this PR not valid, sorry.

Jalle19:
> Ok, I re-read the spec again. Application information resource version less than 3 don't support data_rate_info APDU. It may hang some CAMs. So, this PR not valid, sorry.

Hmm, that's unfortunate. Should we try to find a CAM that actually hangs before reverting?

Yuri666:
I hear about it in tvheadend forum, but didn't seen in real. I don't think it's very important, and not needs to reverting before anybody claim.

Jalle19:
Alright. It's strange that AI version 3 seems to be mandatory for CI+ 1.4 compatible hosts to announce, yet my CI+ 1.4 CAM chooses to use AI version 1 🤷 I originally came up with this change because I thought it was a bug, but it seems the original code was correct, even though some comments were technically wrong.

I'll whip up a revert with some clarifications in the near future, don't want to deal with potential support issues 😄

PR #942, opened 2022-02-13
Various CI resource tweaks

Jalle19:
~Makes the Kabelio CAM in https://github.com/catalinii/minisatip/issues/910 initialize, although some issues still seem to remain.~

@Yuri666 what do you think?

### DIFF
--- a/src/ca.c
+++ b/src/ca.c
@@ -2820,9 +2820,9 @@ static int ca_session_callback(void *arg, int reason, uint8_t slot_id,
              "%i",
              slot_id, resource_id, session_number);
 
-        if (resource_id == EN50221_APP_RM_RESOURCEID) {
+        if (resource_id == EN50221_APP_RM_RESOURCEID ||
+            resource_id == TS101699_APP_RM_RESOURCEID) {
             LOG("--------------------S_SCALLBACK_REASON_CAMCONNECTED---------"
-                "EN50221_"
                 "APP_RM_RESOURCEID-------------------------");
             en50221_app_rm_enq(d->rm_resource, session_number);
         } else if (resource_id == EN50221_APP_AI_RESOURCEID ||
--- a/src/ca.c
+++ b/src/ca.c
@@ -68,7 +68,7 @@ alternative source
 #define TS101699_APP_AI_RESOURCEID MKRID(2, 1, 2)
 #define CIPLUS_APP_AI_RESOURCEID MKRID(2, 1, 3)
 #define CIPLUS_APP_AI_RESOURCEID_MULTI MKRID(2, 1, 4)
-#define TS103205_APP_AI_RESOURCE_ID MKRID(2, 1, 5)
+#define BLUEBOOK_A173_2_APP_AI_RESOURCE_ID MKRID(2, 1, 5)
 #define CIPLUS_APP_DVB_RESOURCEID MKRID(32, 1, 2)
 #define TS103205_APP_DVB_RESOURCEID MKRID(32, 1, 3)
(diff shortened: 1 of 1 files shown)


Yuri666:
SAS support not unfinished. It sends fake host_id to CAM and end session. Without SAS support some CAMs don't work. I have such CAMs.

Yuri666:
btw, Kabelio CAM still not initialized properly with this changes

Yuri666:
hmm, really. sorry forgot to delete the comment

Jalle19:
Dropped the SAS commit, reworked the resource announcement to not include version 5 of Application Information.

Jalle19:
@Yuri666 what do you think about the remaining changeset? Works for me at least with a 1.4 CAM.

Yuri666:
You remove AI resource announcement, but also remove callback processing this resource. Maybe shall leave callback in code if in futures we'll needs to enable AI v.5, we can only add it at one line.

catalinii:
Do you think this would work faster with remote access?

Jalle19:
@Yuri666 sure, I can leave the callback case values in place.

Jalle19:
@catalinii for https://github.com/catalinii/minisatip/issues/910, yes, probably

Yuri666:
Nice. I suggest also to remove this code, because it wrong:
https://github.com/catalinii/minisatip/blob/master/src/ca.c#L75
https://github.com/catalinii/minisatip/blob/master/src/ca.c#L2849

Jalle19:
@Yuri666 done, good catch!

Yuri666:
> @Yuri666 I had to re-add older host control and content control resources for my other CAM to initialize. Not sure what the best approach here is - clearly it's not possible to go by the recommendation in the spec.

You sure what your CAM needs two versions both resources? Can you make log CAM initialization with -l ca option?

Yuri666:
> @Yuri666 FWIW I implemented the date time resource properly here: https://github.com/catalinii/minisatip/compare/master...Jalle19:datetime-test?expand=1
> 
> Can't say whether it makes any difference, but at least now the date is being sent to the CAM as often as the CAM wants 🤷 Will PR once we can sort out this resource mess.

I made same alredy in test branch in my repository. No filling any difference with my CAMs. It doesn't really matter because CAM takes time from the stream

Yuri666:
> Yeah, if I comment out the older host control and/or content control versions it gets stuck when the CAM is supposed to start the content control stuff.
please try to leave only CIPLUS_APP_DVB_RESOURCEID of dvb control and all as is of content control - is it will work?

Jalle19:
We should try https://github.com/OpenViX/enigma2/blob/58c05ef12d11a98a698c368ae2bde4664b550b8b/lib/dvb_ci/dvbci_resmgr.cpp#L82 since everyone seems to say their stuff works on Enigma 🤷

Jalle19:
> We should try https://github.com/OpenViX/enigma2/blob/58c05ef12d11a98a698c368ae2bde4664b550b8b/lib/dvb_ci/dvbci_resmgr.cpp#L82 since everyone seems to say their stuff works on Enigma shrug

Tried it, got "Your host is not fully CI plus compliant" from the CAM :shrug:

Jalle19:
I dropped the "use the highest version of each resource only" commit since it didn't seem to actually work. Also re-added the handling of AI resource version 5 so it's only a matter of announcing it if we want to try to support it.

Jalle19:
Been using this for a long time locally with my two CAMs, everything seems to work smoothly. I've got other fixes lined up that are based on this branch so I will go ahead and merge this and PR the next batch.

If there are any issues with these please let me know so we can keep master working for everyone!

PR #945, opened 2022-02-21
Fix calculation of signal values

lars18th:
When the DVB adapter reports dB signal, fix the result value (it's less in the order of 100 smaller times).
For Enigma STB devices assume that the driver reports dB allways.

### DIFF
--- a/src/dvb.c
+++ b/src/dvb.c
@@ -1676,9 +1676,10 @@ fe_delivery_system_t dvb_delsys(int aid, int fd, fe_delivery_system_t *sys) {
 // returns the strength and SNR between 0 .. 65535
 
 void get_signal(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
-                uint16_t *snr) {
+                uint16_t *snr, uint16_t *db) {
     *status = 0;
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @catalinii ,

The patch is ready to merge.

catalinii:
@lars18th I am fonfused related to the assumption that e2 will always report db always.

As every driver developer is free to use whichever measurement this ia probably not a generic implementation.

Thank you

lars18th:
Hi @catalinii ,

As I writed in https://github.com/catalinii/minisatip/issues/911#issuecomment-1048029529 I'm working on fixing this. However, it will be preferable to merge this PR. Then it will be more easy to fix it. Please think that this patch incorporates more tunning maps, fixes the calculation of dB->Signal, and adds the dB conversion to the `get_signal()` function as well.

Regards.

Jalle19:
Can we merge this? If there's anything that could be improved it could be done in a new PR?

catalinii:
@lars18th merged this one, but please submit a PR later about the enigma2 handling of relative (non db) measures

PR #951, opened 2022-03-08
Properly handle CAMs that request date time updates in intervals

Jalle19:
Many CAMs request the current date and time once, but some want regular updates sent to it. Things seem to work without this interval support, but it's definitely supposed to be implemented according to the spec.

Been using this "in production" for a few weeks without any issues.

@Yuri666 any comments?

### DIFF
--- a/src/ca.c
+++ b/src/ca.c
@@ -219,7 +219,6 @@ typedef struct ca_device {
     int uri_mask;
 
     uint16_t ai_session_number;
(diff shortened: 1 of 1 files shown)


Yuri666:
you said everything correctly. there is no improvements, but the specification requires.

catalinii (approved):
Looks good for me. We can wait for @Yuri666 's comment or merge and change after....either  way is fine

PR #952, opened 2022-03-09
Fix update_cw() validated initialization

lars18th:
The initialization of the "validated" flag is incorrect when only one TS packet with the PUSI mark it's found. With only one positive check it's sufficient to grant that the key is correct. So, the variable requires to be iniatialized to 1 when "len > 0", and not when "len > 1". Futhermore one change in the LOG is included for improved readability.

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -729,7 +729,7 @@ void update_cw(SPMT *pmt) {
     LOGL(old_cw == cw ? DEFAULT_LOG : 1,
          "found CW: %d %s for %s PMT %d, old cw %d, packets %d, parity %d, pid "
          "%d: %s",
-         cw ? cw->id : -1, validated ? "[validated]" : "[pending]", pmt->name, pmt->id,
+         cw ? cw->id : -1, validated ? "[validated]" : "[not validated]", pmt->name, pmt->id,
          old_cw ? old_cw->id : -1, pmt->blen, pmt->parity, pmt->pid,
          cw_to_string(cw, buf));
 
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -659,6 +659,17 @@ int fill_packet_start_indicator(SPMT_batch *all, int max_all, SPMT_batch *start,
     return s;
 }
 
+void dump_cws() {
+    int i;
+    char buf[200];
+    uint64_t ctime = getTick();
+    LOG("List of CWs:");
+    for (i = 0; i < ncws; i++)
+        if (cws[i] && cws[i]->enabled && cws[i]->expiry > ctime) {
+            LOG("* CW %d: %s", cws[i]->id, cw_to_string(cws[i], buf));
+        }
+}
+
 void update_cw(SPMT *pmt) {
     SCW *cw = NULL, *old_cw = pmt->cw;
     char buf[300];
@@ -753,17 +764,6 @@ void update_cw(SPMT *pmt) {
     } else {
         if (opts.debug & LOG_PMT)
             dump_cws();
-}
-
-void dump_cws() {
-    int i;
-    char buf[200];
-    uint64_t ctime = getTick();
-    LOG("List of CWs:");
-    for (i = 0; i < ncws; i++)
-        if (cws[i] && cws[i]->enabled && cws[i]->expiry > ctime) {
-            LOG("* CW %d: %s", cws[i]->id, cw_to_string(cws[i], buf));
-        }
 }
 
 int send_cw(int pmt_id, int algo, int parity, uint8_t *cw, uint8_t *iv,
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -677,10 +677,7 @@ void update_cw(SPMT *pmt) {
     int i = 0;
     SPMT_batch start[10];
     int len = fill_packet_start_indicator(pmt->batch, pmt->blen, start, 9);
-    int validated = 1;
-
-    if (len <= 0)
-        validated = 0;
+    int validated = (len > 0);
 
     // If no CW exists and we can validate the CW, try to find one again
     // Helps for CI+ case
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -666,7 +666,10 @@ void update_cw(SPMT *pmt) {
     int i = 0;
     SPMT_batch start[10];
     int len = fill_packet_start_indicator(pmt->batch, pmt->blen, start, 9);
-    int validated = (len > 1);
+    int validated = 1;
+
+    if (len <= 0)
+        validated = 0;
 
     // If no CW exists and we can validate the CW, try to find one again
     // Helps for CI+ case
@@ -716,9 +719,9 @@ void update_cw(SPMT *pmt) {
         }
 
     LOGL(old_cw == cw ? DEFAULT_LOG : 1,
-         "found CW: %d %sfor %s PMT %d, old cw %d, packets %d, parity %d, pid "
+         "found CW: %d %s for %s PMT %d, old cw %d, packets %d, parity %d, pid "
          "%d: %s",
-         cw ? cw->id : -1, validated ? "[validated] " : "", pmt->name, pmt->id,
+         cw ? cw->id : -1, validated ? "[validated]" : "[pending]", pmt->name, pmt->id,
          old_cw ? old_cw->id : -1, pmt->blen, pmt->parity, pmt->pid,
          cw_to_string(cw, buf));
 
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

With this patch some of my glitches with the decoding process will disappear. Please, test it.
In addition, I've added a small change in the log for easier decode verification.

Regards.

lars18th:
Hi @catalinii ,

The movement of the dump_cws() function it's because the call from the update_cw() when not key is found and validated. Nothing more. Detected by the Coverity checks!

src/pmt.c
```diff
@@ -659,14 +659,28 @@ int fill_packet_start_indicator(SPMT_batch *all, int max_all, SPMT_batch *start,
...
     char buf[300];
     int64_t ctime = getTick();
     int i = 0;
     SPMT_batch start[10];
     int len = fill_packet_start_indicator(pmt->batch, pmt->blen, start, 9);
-    int validated = (len > 1);
+    int validated = 1;
+
+    if (len <= 0)
+        validated = 0;
```
Jalle19, line 683: Couldn't all this be simplified into `int validated = (len > 0);`?

lars18th: > Couldn't all this be simplified into `int validated = (len > 0);`?


No. That's the reason for the patch... As I say in the description **_several compilers fail to initialize correctly_**. You can check it in your system activating the log with `-v pmt`. If you see multiple times that the same key is checked then your compiler has troubles. So please, belive me: it's preferable writing the code in this way. The result never fails, and if the compiler wants to optimize it, it can do it. Therefore this patch doesn't introduce any penalty in the performance.

I've dedicated a lot of time to found this!
My recomendation is to merge this PR now.

Jalle19: Have you tried `volatile int validated = (len > 0);`?

Jalle19: The thing is you're both changing the logic _and_ adding some form of optimization hack. Are you sure it's not enough to just change the logic?

lars18th: Hi @Jalle19 (and @catalinii ),

Sorry, sorry! I revised the code, and the problem is another. The correct solution is the one you proposed:
```
int validated = (len > 0);
```
So, yes. In the original code some bug exist because with only 1 packet with the PUSI flag it's sufficient to do the validation. I don't know why the original author used `validated = (len > 1)` as it doesn't have sense.

Now, it's fixed, so please, merge the PR.
Regards.

src/pmt.c
```diff
@@ -716,9 +727,9 @@ void update_cw(SPMT *pmt) {
         }
 
     LOGL(old_cw == cw ? DEFAULT_LOG : 1,
-         "found CW: %d %sfor %s PMT %d, old cw %d, packets %d, parity %d, pid "
+         "found CW: %d %s for %s PMT %d, old cw %d, packets %d, parity %d, pid "
          "%d: %s",
-         cw ? cw->id : -1, validated ? "[validated] " : "", pmt->name, pmt->id,
+         cw ? cw->id : -1, validated ? "[validated]" : "[pending]", pmt->name, pmt->id,
```
catalinii, line 732: Why pending? This will be the CW used for decrypting

src/pmt.c
Jalle19: Very good, thanks for checking! Now the fix makes perfect sense

Jalle19 (approved):
Can't really comment on the logging change, but the logic change sounds good

lars18th:
> Can't really comment on the logging change, but the logic change sounds good

The logging change only does this: when a key is not found, then all keys are printed. This will help to found error/bugs/glitches with the key selection.

And regarding the logic change, I've reduced the glitches with the decoding. I'm absolutly convinced that the original author wants to check for 1 correct packet and not for more than one.

So, I hope @catalinii will merge it soon.

src/pmt.c
Jalle19: I'm guessing he means it's pending validation?

lars18th: Hi @catalinii ,

> Why pending?

Because this key will be pending until it be validated.

> This will be the CW used for decrypting

Sure? From what I've checked, if a key is not validated (aka "pending"), then it's because one of this:
1. The key is reported but not already checked (because missing PUSI packets).
2. The key is checked with a PUSI packet and the check failed.

Then what the current code does is to mainatin this key in the candidate list of keys. That until in the future a new one arrives or some change is done and the PUSI test will be passed. So a key can only be "validated" or "pending". Why not then print the two status? Printing the key without the "pending" label could induce to think that the key is correct, and in fact it's a key pending to validate it (or pending to be overwrited with another one because is not a valid key). 

Therefore, please accept this change to make more easier to debug key troubles.

lars18th: > I'm guessing he means it's pending validation?

So, you prefer to use the word "unvalidated"? It's that?

Jalle19: I think it's fine like this.

Jalle19:
@catalinii can we merge this?

src/pmt.c
lars18th: Changed to "not validated". So comment resolved. 😉

lars18th:
> Not validated may be better or just leave it as it is.

Done! Thank you for the comments and positive feedback.
Now ready to merge.

Jalle19:
@lars18th https://github.com/Jalle19/satip-axe/issues/28 apparently this commit is causing decryption issues for many people? :thinking:

lars18th:
> @lars18th [Jalle19/satip-axe#28](https://github.com/Jalle19/satip-axe/issues/28) apparently this commit is causing decryption issues for many people? 🤔

Hi @Jalle19 ,

I don't think so. In which case a key can be validated 1 time and failed in with next packet? This patch has resolved a lot of troubles in my system because I use a very small buffer queue. So, only 1 PUSI packet could be checked at one shot. And then this patch provides a more stable behaviour. I'm using it for a long time!

But I can't be sure for others. All systems require to do real world checks. So, to be completly sure if this PR has some side effect, the solution is to recompile with this minimal change in the "pmt.c" file:
```
void update_cw(SPMT *pmt) {
    SCW *cw = NULL, *old_cw = pmt->cw;
    char buf[300];
    int64_t ctime = getTick();
    int i = 0;
    SPMT_batch start[10];
    int len = fill_packet_start_indicator(pmt->batch, pmt->blen, start, 9);
-    int validated = (len > 0);
+    int validated = (len > 1);

    // If no CW exists and we can validate the CW, try to find one again
    // Helps for CI+ case
```
If after this patch the behaviour not changes, then the trouble observed by others has other cause. However, if this fixes something, then I'll prepare a new PR incorporating a new option to set or modify this value. You agree with that?

Regards.

catalinii:
I remember also having some instability (that's why I set to at least 2 packets with PUSI).

catalinii:
Maybe we should handle separately if we have just 1 packet with PUSI, like if there is no key found fallback to non validated logic

catalinii:
> Hi @catalinii ,
> 
> First: Reading the post where people writes about a trouble [Jalle19/satip-axe#28](https://github.com/Jalle19/satip-axe/issues/28) you can see this:
> 
> > > Are you having issues on FTA channels too or just encrypted ones?
> 
> > FTA channels as well
> 
> So don't be wrong! The trouble seems to not be related with this PR.
> 
> However, regarding the idea of using at least 2 packets with PUSI for the check, let me to comment:
> 
> * The basis is this: if one key can decrypt one PUSI packet, then it could decrypt any other with the same parity. It's imposible any other case (except if the packet checked has errors).
Correct
> * The current logic as I can see it's to use the non-validated key, but continue asking for a new key. This causes troubles in my system if the previous returned key is correct. And for this reason I submited this patch.

If you cannot validate the key is very hard to figure out which key is correct. This could happen if the provider switches from the new to old key fore few packets.


The other issue that falls in this case is that there is one packet but that packet could have errors. Due to how dvbcsa is designed the first 8 bytes will get corrupted even if one bit is corrupted in the entire packet.

Possible solution for this case is:
- packet error flag (not sure how reliable is this)
- validate using h264/h265 internal headers
- guess the partity
Of course getting this wrong leads to visial artifacts (user impact).
> 
> So if you're worried about "some instability", then perhaps it could be good to check in the function `fill_packet_start_indicator()` the ERROR FLAG of the header. In fact, if the source packet to check has errors, then it could generate a false positive (although with a very low probability).
> 
> What you think?

lars18th:
Hi @catalinii ,

Sorry, but I don't agree with _all_ you comment. Please, note this:

> If you cannot validate the key is very hard to figure out which key is correct.

The question is not this. The question is what happens if you valite the key with 1 PUSI packet instead of 2 or more. And the answer is: *nothing*. It never can't happen that with _two_ packets with the same parity one fails to decrypt and one goes right.

> This could happen if the provider switches from the new to old key fore few packets.

And this happens. In fact I see it a lot of times. But the current code works without problems: because the old key is not expired, so when the parity change occurs, it could happen that inside a small period packets with different parity appear. So instead of 1 switch, in fact multiple happens. But as I say the code works: it changes multiple times, and all works. In fact, the problem appears when you artificially force to have 2 PUSI checks. In this case, when the flip-flop switch occurs multiple times the key search fails as only 1 packet could be check... of the _old_ parity. And then the algorithm select the _last_ key that perhaps is the wrong key.

Therefore, instead of thinking that the current implementation is the cause of the errors that these users comment, my feel is that the root cause is another very different. The only change that I recommend is to add the check of the packet error flag in the function `fill_packet_start_indicator()` because a false positive could happen. But nothing more.

Regards.

Jalle19:
A check was added in https://github.com/catalinii/minisatip/commit/c8b10b220a4650b55ad3b13ca0f64f42a3bf3096

Jalle19:
> @lars18th [Jalle19/satip-axe#28](https://github.com/Jalle19/satip-axe/issues/28) apparently this commit is causing decryption issues for many people? thinking

This turned out to be a false alarm - satip-axe works correctly with v1.1.84 (which is the first release with these changes).

PR #953, opened 2022-03-12
Replace usleep() with nanosleep()

### LINKED ISSUE #528: RFE: suggestion replace obsolete usleep with nanosleep
I like to suggest replaceing usleep with nonosleep and use a util function like:
```
void myusleep(uint64_t usec)
{
    struct timespec ts = { .tv_sec = usec / 1000000,
                           .tv_nsec = (usec % 1000000) * 1000};
#if defined(HAVE_SELECT)
    select(0, NULL, NULL, NULL, &ts);
#else
    nanosleep(&ts, (struct timespec *)NULL);
#endif
}
```
According to usleep() manpage:
POSIX.1-2001 declares this function obsolete; use nanosleep(2) instead.
POSIX.1-2008 removes the specification of usleep().

/////////////
select(0, NULL, NULL, NULL, &ts); is the lazy method wine is using :-) and it's the same on windows (req. winsocket)
////////////
the other option would be
```
#include <ctime>
void myusleep(int ms)
{
    clock_t starting = clock();
    while( clock() - starting < ms ) {}
}
```

### LINKED ISSUE #607: ca.c:2678 Invalid usleep() argument nr 1.
https://github.com/catalinii/minisatip/blob/6ea7837967a13b5036c06266fa7e41da1859eae5/src/ca.c#L2678
in the man page
```
       The type useconds_t is an unsigned integer type capable of holding
       integers in the range [0,1000000].
```

Jalle19:
Code copied from oscam

Closes #528, closes #607

Untested, will start runtime testing tomorrow. Should be pretty straight-forward though,

### DIFF
--- a/src/ca.c
+++ b/src/ca.c
@@ -2408,7 +2408,7 @@ void *stackthread_func(void *arg) {
     LOG("%s: start", __func__);
 
     while (d->enabled) {
-        usleep(100 * 1000);
+        sleep_msec(100);
 
         // Check for slot errors
         int error;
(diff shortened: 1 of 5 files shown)


catalinii (approved):

Jalle19:
Still haven't had time to test this, won't have time before next week

Jalle19:
Testing this at the moment

Jalle19:
Has been working for 24 hours, I'm pretty confident this works. @catalinii feel free to merge!

PR #954, opened 2022-03-16
Reset decoding errors when starting to decode

lars18th:
When a new program is started to decode, at first some packets can't be decoded. This could be annoying when viewing the UI as you can interpret that some decoding errors have appeared. But this is not the case. Futhermore, if after the initial decoding some packets can't be decoded you need to remember the initial count to know the real non-decoded packets.

This patch adds a new behaviour: When first starting to receive a new program, the counter of non-decoded packets will continue to work as now. Therefore, if the program cannot be decoded, the counter can be viewed incrementing. But when the program starts to be decoded, then the counters are reset. So only if new decoding errors appear after the initial decoding you will see this counter in the UI.

This new behaviour is more consistent and user friendly.

### DIFF
--- a/src/dvb.h
+++ b/src/dvb.h
@@ -355,6 +355,7 @@ typedef struct struct_pid {
                  // tune is called, 3 disable when tune is called
     int packets; // how many packets for this pid arrived, used to sort the pids
     int dec_err; // decrypt errors, continuity counters
+    uint8_t is_decrypted;  // Set when first decrypted
     int16_t pmt, filter;
     char cc, cc1;
     int sock; // sock_id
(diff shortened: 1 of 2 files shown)


Jalle19:
Nice, I've been wondering why the UI says there are hundreds/thousands of DEC-ERRORS that then tend to remain constant.

catalinii:
@Jalle19 is not the case for this PR, but in general please check https://github.com/catalinii/minisatip/issues/841 to ensure there are no coverity issues (only when merging PRs).

Jalle19:
@catalinii thanks, I'll try to remember!

PR #955, opened 2022-03-18
Debug dump of CWs if the validation fails

lars18th:
Print debug data of all CWs not only when the no CW is found, but also when not validated.


lars18th:
Hi @catalinii ,

I've missed this in my previous PR related to this!

catalinii:
@lars18th do you think it would be better topeint just the CWs for that specific PMT?
I am afraid that this will bloat the output and potentially be confusing.

lars18th:
> @lars18th do you think it would be better topeint just the CWs for that specific PMT?

Why? This is only printed with the debug of the PMT enabled. And it will only appears if some trouble with the key is found. I'm not worried about that in any sense.

PR #961, opened 2022-03-31
Remove dependency on dvb-apps

catalinii:


catalinii:
cc @Jalle19 @Yuri666 can you guys test this PR?
For me this does not work and fails auth around 40% missing an call to CIPLUS_APP_CC_handler from the CAM.
Any pointers on what could be wrong ?

src/ca.c
```diff
@@ -172,59 +144,58 @@ struct ci_buffer {
...
-    struct en50221_app_send_functions sf;
-    struct en50221_app_rm *rm_resource;
-    struct en50221_app_ai *ai_resource;
-    struct en50221_app_dvb *dvb_resource;
-    struct en50221_app_ca *ca_resource;
-    struct en50221_app_datetime *dt_resource;
-    struct en50221_app_mmi *mmi_resource;
+    // Session Numbers: as all the objects are stored inside of the ca_device,
+    // don't store data inside of the sessions
+    int session_number;
```
Jalle19, line 192: How is this supposed to work? There's just a single session number here?

src/ca.c
```diff
@@ -419,6 +301,12 @@ unsigned char dh_g[256] =
 
 int dvbca_close_device(ca_device_t *c);
 static int ca_send_datetime(ca_device_t *d);
+int populate_resources(ca_device_t *d, int *resource_ids);
+int ca_write_apdu_session(ca_device_t *d, int session_number, int resource,
+                          const void *data, int len);
+int ca_write_apdu(ca_device_t *d, int resource, const void *data, int len);
+int find_session_for_resource(int resource);
+int asn_1_decode(int *length, unsigned char *asn_1_array);
```
Jalle19, line 309: IMO to keep this file more managable in size, ASN.1 functions could be moved to a different file

src/ca.c
```diff
@@ -2256,15 +2035,15 @@ static int ci_ccmgr_cc_sac_send(struct ci_session *session, const uint8_t *tag,
...
 }
 
-static int ci_ccmgr_cc_sac_data_req(ca_device_t *d, struct ci_session *session,
-                                    const uint8_t *data, unsigned int len) {
-    struct cc_ctrl_data *cc_data = session->private_data;
-    const uint8_t data_cnf_tag[3] = {0x9f, 0x90, 0x08};
+static int ci_ccmgr_cc_sac_data_req(ca_device_t *d, const uint8_t *data,
+                                    unsigned int len) {
+    struct cc_ctrl_data *cc_data = &d->private_data;
+    uint32_t data_cnf_tag = 0x9f9008;
```
Jalle19, line 2046: There are DEFINEs for most tags, IMO we should use defines for these as well instead of magic numbers

src/ca.c
```diff
@@ -2526,16 +2246,16 @@ static int ciplus_app_lang_message(void *arg, uint8_t slot_id,
     {
         LOG("country answered with '%c%c%c'", data_reply_country[0],
             data_reply_country[1], data_reply_country[2]);
-        uint8_t tag[3] = {0x9f, 0x81, 0x01}; /* host country reply */
-        ci_session_sendAPDU(d->session, tag, data_reply_country, 3);
+        /* host country reply */
+        ca_write_apdu(d, 0x9f8101, data_reply_country, 3);
```
Jalle19, line 2250: Same thing here regarding magic numbers

src/ca.c
```diff
@@ -2736,625 +2419,770 @@ static int ciplus_app_upgr_message(void *arg, uint8_t slot_id,
...
+    case TAG_PROFILE:
+        ca_write_apdu(d, TAG_PROFILE_CHANGE, NULL, 0);
         break;
-    case CIPLUS_TAG_CICAM_RESET:
-        //		hexdump(data,data_length);
-        if (ioctl(fd, CA_RESET, &info))
-            LOG_AND_RETURN(0, "%s: Could not reset ca %d", __FUNCTION__, d->id);
-        return 1;
+    case TAG_PROFILE_ENQUIRY:
+        // do we really need this to force CI only mode ?
```
Jalle19, line 2468: I'm not sure but I don't think we have to - CI CAMs should simply ignore whatever unknown resources the host declares

src/ca.c
```diff
@@ -2736,625 +2419,770 @@ static int ciplus_app_upgr_message(void *arg, uint8_t slot_id,
...
+    case TAG_DISPLAY_CONTROL: {
+        int cmd_id = buffer[0];
+        int mmi_mode = buffer[1];
+        if (cmd_id == MMI_DISPLAY_CONTROL_CMD_ID_SET_MMI_MODE) {
 
-    if (en50221_app_rm_enq(d->rm_resource, session_number)) {
-        LOG("%02x:Failed to send ENQ", slot_id);
+            LOG("mmi display ctl cb received for CA %u session_num %u "
+                "cmd_id 0x%02x mmi_mode %u",
+                d->id, d->session_number, cmd_id, mmi_mode);
```
Jalle19, line 2608: Here for example it seems like `d->session_number` isn't actually the session number for the MMI resource. Maybe this field should be removed from `ca_device_t` completely to avoid confusion?

src/ca.c
```diff
@@ -2736,625 +2419,770 @@ static int ciplus_app_upgr_message(void *arg, uint8_t slot_id,
...
+    DEFAPP(EN50221_APP_RM_RESOURCEID, APP_RM_handler, APP_RM_create),
+    DEFAPP(TS101699_APP_RM_RESOURCEID, APP_RM_handler, APP_RM_create),
+    // Application Information
+    DEFAPP(EN50221_APP_AI_RESOURCEID, CIPLUS_APP_AI_handler, APP_AI_create),
+    DEFAPP(TS101699_APP_AI_RESOURCEID, CIPLUS_APP_AI_handler, APP_AI_create),
+    DEFAPP(CIPLUS_APP_AI_RESOURCEID, CIPLUS_APP_AI_handler, APP_AI_create),
+    // Conditional Access Support
+    DEFAPP(EN50221_APP_CA_RESOURCEID, APP_CA_handler, APP_CA_create),
+    // TS103205_APP_CA_MULTISTREAM_RESOURCEID, // Multi-stream
+    // Host Control - Does not seem like other apps are annoucing it
```
Jalle19, line 2715: Not sure about this, AFAIK it's a required resource according to the spec

src/ca.c
```diff
@@ -2736,625 +2419,770 @@ static int ciplus_app_upgr_message(void *arg, uint8_t slot_id,
...
 }
 
-static int ca_mmi_enq_callback(void *arg, uint8_t slot_id,
-                               uint16_t session_number, uint8_t blind_answ,
-                               uint8_t exp_answ_len, uint8_t *text,
-                               uint32_t text_size) {
-    ca_device_t *d = arg;
-    char buffer[256];
+int asn_1_encode(int length, uint8_t *asn_1_array) {
+    if (length < 0x80) {
```
Jalle19, line 2798: Incidentally this fixes a bug in the previous implementation (was `< 127` when it should have been `< 128`). I had a fix for this coming up at some point.

src/ca.c
Jalle19: Same thing with the AES and other crypto stuff in this file, but maybe it should all be moved in a later PR.

catalinii: Once I get it working I can split to multiple files and fix the magic numbers

src/ca.c
catalinii: This is mostly for forcing using CI (not CI+) from the host side (if for example the certs are missing)

src/ca.c
Jalle19: Yeah, I realized when reading the whole thing

src/ca.c
Jalle19: True

src/ca.c
catalinii: we have 2 choices:
- use d->sesion_number (or rename it)
- carry session_number as argument across all the method as part of the CAM request
WDYT ?

src/ca.c
catalinii: yeah I saw multiple implementation each with different logic... so I took the one from dvb-apps.

catalinii:
Enabling Host control as suggested by @Yuri666 fixed few cams, mine is still not working so I will have to dig deeper.

Jalle19:
Tried a Kabelio CAM with this patch set, same issue as before (CAM never opens a content control resource) :shrug:

catalinii:
Do u have access to kabelio?
This patch works on enigma2 receivers as well

Jalle19:
@catalinii yes, for the moment

catalinii:
Can u share it?
The point is we can sniff enigma2 traffic and compare.

catalinii:
One note: datetime for enigma2 and minisatip/dvb-apps is different.
Minisatip sends an apdu of 5 bytes while enigma2 sends an apdu of 6 bytes (first one being 05).
My CAM fails with enigma2 way of sending the datetime.

catalinii:
Enigma code: https://github.com/openatv/enigma2/blob/7.0/lib/dvb_ci/dvbci_datetimemgr.cpp#L66

Yuri666:
Enigma code is wrong

Table 32: Date-Time Enquiry object coding
Syntax No. of bits Mnemonic
date_time () {
date_time_tag 24 uimsbf
length_field()= 5 or 7
UTC_time 40 bslbf
local_offset /* optional */ 16 tcimsbf

catalinii:
I will create a PR for enigma2

Jalle19:
> Can u share it?

Yes, please contact me by e-mail on sam.stenvall@iki.fi.

Issue #960, opened 2022-03-27
EMMs not sent to CAM when using DD CI


Jalle19:
I've been using a CAM for about two months now, and two times I've had to move the CAM temporarily to my TV in order to get entitlements updated. I suspect there's something wrong with the code, but I'm not really sure how to fix it.

@catalinii AFAICT you added support for this many years ago? Not sure when it stopped working.

Jalle19:
I'm also thinking that this could be part of the reason why Kabelio CAMs cannot be initialized fully (https://github.com/catalinii/minisatip/issues/910). Many are saying that some CAMs need to have data passed to them before they "wake up" and initialize the content control resource, and perhaps missing EMMs is why this doesn't seem to have any effect when using minisatip.

Yuri666:
To test your assumption, needs to try to initialize the Kabello CAM in dvb card with imbedded CI interface - it pass full transonder stream to the CAM.

catalinii:
The problem was reported before however was not able to repro. If you tune to a TP i. The beginning and leave it that way for let's say an hour, the CAM will get updated and EMMs are working fine. I think something happens of you keep changing the channels.

If you get a repro I can take a look.
This is where the mapping of emm pids happens: https://github.com/catalinii/minisatip/blob/master/src/ddci.c#L1250

Jalle19:
My logs look like this when grepping for that particular log line:

```
$ journalctl -u minisatip --since "2022-03-27 19:00:00" | grep 'CAT pos\|mapped adapter'
Mar 27 19:50:03 vladimir minisatip[22640]: [27/03 19:50:03.372 AD0]: mapped adapter 0 (0) pid 1 (1) to 1, pmt 12
Mar 27 19:50:03 vladimir minisatip[22640]: [27/03 19:50:03.372 AD0]: mapped adapter 0 (0) pid 1464 (1464) to 1464, pmt 12
Mar 27 19:50:03 vladimir minisatip[22640]: [27/03 19:50:03.372 AD0]: mapped adapter 0 (0) pid 2758 (2758) to 2758, pmt 12
Mar 27 19:50:03 vladimir minisatip[22640]: [27/03 19:50:03.372 AD0]: mapped adapter 0 (0) pid 450 (450) to 450, pmt 12
Mar 27 19:50:03 vladimir minisatip[22640]: [27/03 19:50:03.372 AD0]: mapped adapter 0 (0) pid 550 (550) to 550, pmt 12
[...]

PR #962, opened 2022-04-19
Fix issue #910

catalinii:
- Blacklisting is no longer needed since the assignment can be specified in ddci.conf

- Fix EMM handling

### DIFF
--- a/src/ddci.c
+++ b/src/ddci.c
@@ -139,10 +139,9 @@ int add_pid_mapping_table(int ad, int pid, int pmt, ddci_device_t *d,
             }
         }
     if (add_pid && force_add_pid) {
-        if (pid != 1) {
-            m->pid_added = 1;
-            mark_pid_add(DDCI_SID, ad, pid);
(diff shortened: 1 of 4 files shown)


Jalle19:
Most likely fixes https://github.com/catalinii/minisatip/issues/960 too, will test later today

Jalle19 (approved):

src/ddci.c
```diff
@@ -139,10 +139,9 @@ int add_pid_mapping_table(int ad, int pid, int pmt, ddci_device_t *d,
             }
         }
     if (add_pid && force_add_pid) {
-        if (pid != 1) {
             m->pid_added = 1;
```
Jalle19, line 142: Indentation is wrong now

Jalle19:
Seems to at least not make anything worse (descrambling still works) for my CAMs :+1: too early to tell whether EMMs are being sent or not, entitlements expire on April 30th.

PR #963, opened 2022-04-29
Fix compilation overflow

lars18th:
Fixes:

> on my box, strength from 0..100 seems to not be the case (vuuno4kse).
> dvb.c:1733:28: runtime error: signed integer overflow: 65520 * 65535 cannot be represented in type 'int'

### DIFF
--- a/src/dvb.c
+++ b/src/dvb.c
@@ -1730,7 +1730,7 @@ void get_signal(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
     int64_t strengthd = 0, snrd = 0, init_snr = 0;
     float tf;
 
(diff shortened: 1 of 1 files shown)


catalinii:
My point was that one some boxes is 0..100 (relative) and on others is 0..65535...

lars18th:
> My point was that one some boxes is 0..100 (relative) and on others is 0..65535...

Sure! For this reason I've in my TODO to add a manual configuration of each tuner to enforce the mode. However, I'm very busy at time to implement the modification.

lars18th:
Hi @catalinii ,

I close this PR because is already included in #1030 .

PR #964, opened 2022-04-30
Fix master tests

catalinii:


Jalle19:
Can you move the PMT related changes to a separate commit?

catalinii:
The PMT changes prevent the AXE compilation. Need to fix also coverity which fails because of netceiver module

catalinii:
The reason is that I rebuild minisatip-build-image with an updated version of the vdr-plugin-mcli

PR #968, opened 2022-05-17
Send diseqc command with 18V

lars18th:
When multiple tuners share the same wire in a master-slave configuration, it's preferable to send all diseqc commands using the 18V voltage. Then commands are more easy to read from devices when tuners inject incorrect power or slave tuners are miss configured. This behaviour increase the robustness of the diseqc commands.


lars18th:
Hi @catalinii and @perexg ,

Please check it. This improves the switching in my house using multiple tuners connected to the same wire.

Regards.

lars18th:
Hi,

No comments? Take note that the CI error it's caused by an internal error of the service. The code is clean. In my environment this change has reduced the tunning errors caused by switches to near zero. And the original idea is not mine.

Regards.

lars18th:
Hi,

No one tested this patch ❓ 😮

Jalle19 (approved):

Jalle19:

PR #970, opened 2022-05-17
Add FreeBSD support

decke:
Thanks a lot for that really nice project! During one of the lockdowns in 2020 I played around with minisatip on FreeBSD and created a few patches to get it compile properly. The submitted patches are from that time but it took me quite long to finally upstream them because minisatip compiles, starts fine, detects an adapter and the webui works but I cannot seem to get a stream. I still need to figure out why and fix it.

Compiling on FreeBSD works like this:

```
./configure
gmake EXTRA_CFLAGS=-I/usr/local/include EXTRA_LDFLAGS=-L/usr/local/lib
```

I also created a FreeBSD port for minisatip so users might prefer to do:

`pkg add minisatip`

Port: https://cgit.freebsd.org/ports/tree/multimedia/minisatip

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -1851,6 +1851,7 @@ int main(int argc, char *argv[]) {
         if ((ssdp1 = udp_bind(opts.disc_host, 1900, 1)) < 1)
             FAIL("SSDP: Could not bind on %s udp port 1900", opts.disc_host);
         if (opts.bind_dev) {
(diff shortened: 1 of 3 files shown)


decke:
I created a new test setup with FreeBSD 13.1/amd64, webcamd 5.17.1.0 and minisatip 1.2.1 and it works like a charm. Thanks a lot for merging!

PR #977, opened 2022-07-20
Use coverity from the base image

darinatoda:


Jalle19:
The workflow is still using ubuntu-latest, I believe that's why it's still failing

PR #979, opened 2022-07-31
Fix -A in combination with -S for AXE devices

Jalle19:
Since `source_enabled_for_adapter` always returned `1`, the wrong adapter ended up being chosen if the list of sources per adapter has been modified using `-S` (useful when having two Unicable inputs).

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -667,9 +667,6 @@ int source_enabled_for_adapter(adapter *ad, transponder *tp) {
         return 1;
     if (tp->sys != SYS_DVBS && tp->sys != SYS_DVBS2)
         return 1;
-#ifdef AXE
-    return 1;
-#endif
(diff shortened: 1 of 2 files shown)


Jalle19:
I'm still doing some testing, but at least with this change I got the right input to be used depending on the `src=` value in the request. Without this patch, the first available adapter was always chosen (since all adapters had all sources enabled).

Jalle19:
Finished testing, now `-S 0-2:0,3:3 -A 0:0:0,0:1:0,0:2:0,1:3:0` does what `-S 0-2:0,3:3 -U 0-2:3` used to do.

Jalle19:
Next up I'll try to make the input selection smarter so any tuner can use any Unicable input.

catalinii (approved):

Issue #978, opened 2022-07-31
Problems tuning Eutelsat 9ºE multistream channels


patacalas:
Hi!
I have a Gigablue UHD Quad 4K running Minisatip standalone. 
My receiver is enhanced with the PnP DVB-S2X dual tuner so I can receive all Italian multistream channels from Eutelsat 9ºE on Enigma2.
I've enabled frontends 1,5-7,9 on Minisatip, the first four are connected to Unicable LNBs serving Astra, Hotbird and Hispasat using the FBC tuner (no problem with this part) and frontend number 9 is connected to a Legacy LNB to Eutelsat 9ºE to receive Italian channels.

I have 2 problems with Eutelsat 9ºE multistream channels and Minisatip: 

The first one is that Minisatip can't tune transponders with 16APSK modulation (I have good reception of this channels using Enigma2). My client receiver is a Vu+Solo SE using Enigma2 Sat-IP Client. I guessed the problem should be the client receiver, as it isn't a DVB-S2X, but generating the rstp request with VLC and mtype=16apsk param it doesn't work either.

The second problem is related to the transponder 12149V. I have no problems tuning 8PSK multistream transponders, with the exception of 12149V. For some reason, this transponder works like a charm using Enigma2 but Minisatip can't tune it.  The rest of 8PSK multistream transponders work great in both Enigma2 and Minisatip.

I've tested different releases of Minisatip, from v1.1.81 to v1.2.7 and those problems happen in all of them. 
After several tests, I have installed v1.1.92 because Oscam (running on client receiver) doesn't work in newer versions, I get DEC-ERRORS on Minisatip and the client receiver isn't able to decrypt channels. I guess this problem with Oscam could be solved with some configuration parameter that I can't find (I'm a newbie in Minisatip).

I use the following command to start Minisatip:
`minisatip -e 1,5-7,9 -S 5-7:1 -j 1:16-1530,5:17-1566,6:18-1602,7:19-1638 --satip-xml http://192.168.194.39:8554/ -R
/usr/share/minisatip/html`

Thanks in advance.

catalinii:
Hi,

Can you try to upload the logs of:

./minisatip -f -l http,dvb

And then try the vlc using that URL?

patacalas:
Hi @catalinii 
I'm sharing logs related to the issues.

### First issue: tuning 16APSK modulation transponders is impossible in Minisatip.
[...]

PR #983, opened 2022-08-02
Use Gold mode for dvb-s2 if nothing specified #978

catalinii:
Before this PR to set correctly the DTV_SCRAMBLING_SEQUENCE_INDEX you need to send also plsm=[root|gold]. If pls_code is specified then there is no need to specify root or gold and can use the pls_code directly for DTV_SCRAMBLING_SEQUENCE_INDEX


Jalle19 (approved):
Don't really know what this option does but if it seems to work then it's fine

PR #986, opened 2022-08-04
Add Debian packaging

Jalle19:
@marillat let's continue the discussion here


Jalle19:
@marillat what's the command to build a .deb package? I could add it to the CI/CD pipeline, then we can start refining it.

marillat:
dpkg-buildpackage from dpkg-dev package

marillat:
I added a minisatip user to start minisatip with systemd in 1.2.10-dmo2

Jalle19:
@marillat awesome, I incorporated the changes here and amended the commits to make you the author.

What's next here? I guess we should document how to build a package, then I can test that it actually works (I happen to run minisatip on Debian too, but so far I've manually installed it and hacked together a similar systemd unit).

Jalle19:
@marillat please review my last three commits, are they okay?

marillat:
We have the same debian directory (except  [ade0252](https://github.com/catalinii/minisatip/pull/986/commits/ade0252f742b0f80423e371e790594ad19be3d74)) so it's perfect.

catalinii (approved):

Jalle19:
@marillat can we drop the `nodejs-jquery` dependency? The file is after all bundled here and I don't see that changing any time soon.

Also, can we change the package name to just `minisatip` instead of `minisatip-dmo`?

Jalle19:
Not really sure how we could automate the generation of new packages (with bumped version number and associated change log entries), I would need help with that. Would definitely be easier if we didn't do automatic releases :/

@catalinii @marillat thoughts?

marillat:
> @marillat can we drop the `nodejs-jquery` dependency? The file is after all bundled here and I don't see that changing any time soon.
> 
> Also, can we change the package name to just `minisatip` instead of `minisatip-dmo`?

Yes you can drop `nodejs-jquery` and remove -dmo to the source name. Should be done in the control and changelog files.

marillat:
> Not really sure how we could automate the generation of new packages (with bumped version number and associated change log entries), I would need help with that. Would definitely be easier if we didn't do automatic releases :/

automatic release is boring must be done by an human.

catalinii:
@Jalle19

Generally it needs to be added to this list of files: https://github.com/catalinii/minisatip/blob/master/.github/workflows/binaries.yml#L51

 If needed, I can do that (posting debian package in the binary section).

Jalle19:
@marillat can you review the last few commits?

marillat:
seems to be OK.

Jalle19:
Haven't gotten around to testing the package yet, but let's merge this and continue iterating

marillat:
I noticed that debhelper-compat version has been downgraded from 13 to 12, Why ?

marillat:
I noticed that deluser is called with --quiet
Probably a bad idea if  something goes wrong user will never seen the deluser errors.

Jalle19:
> I noticed that debhelper-compat version has been downgraded from 13 to 12, Why ?

Ubuntu 22.04 doesn't have 13, so I didn't manage to build it without lowering compat to 12.

Jalle19:
Feel free to make pull requests if you have improvements!

PR #987, opened 2022-08-07
Obey -M when parsing SAT>IP signal strength values too

### LINKED ISSUE #965: Signal strength and SNR values (doesn't take -M parameters)
When using -M parameter, it doesn't do anything with it when I set the STR/SNR values to 0 for all adapters for example (to get 100%) or any other value in the status page of minisatip on port 8080. Is this a bug?

capture str snr

Jalle19:
Fixes #965 

Renamed `snr` to `quality` to make it clear where the odd maximum value (15) comes from (the spec).

@retrofan1979 this should really work now, can you test it?

Server:
Screenshot-20220807112504-166x108
Client (using server via `-s`):
Screenshot-20220807112455-177x121

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -435,41 +435,41 @@ int satipc_timeout(sockets *s) {
 }
 
 void set_adapter_signal(adapter *ad, char *b, int rlen) {
-    int strength, status, snr;
+    int strength, status, quality;
     char *tun, *signal = NULL;
 
     tun = strstr((const char *)b, "tuner=");
     if (tun)
         signal = strchr(tun, ',');
     if (signal) {
-        sscanf(signal + 1, "%d,%d,%d", &strength, &status, &snr);
+        sscanf(signal + 1, "%d,%d,%d", &strength, &status, &quality);
         // Workaround for faulty servers (level=0)
         // Ex. XORO:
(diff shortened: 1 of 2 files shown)


retrofan1979:
Sorry, not over here. I never see 100%/100%.
[minisatip.log](https://github.com/catalinii/minisatip/files/9276500/minisatip.log)

Jalle19:
> I assume this commit is present when compiling the source again?

You have to compile a version that includes this commit - it's not enough to compile master. To do that you can `git clone https://github.com/Jalle19/minisatip` and then `git checkout satipc-multiplier`, then compile normally.

retrofan1979:
OK, I see. I usually only compile master indeed. ;-)
git checkout satipc-multiplier gives an error:
fatal: not a git repository (or any of the parent directories): .git

Also: This branch is [10 commits behind] catalinii:master.

Jalle19:
You forgot to enter the `minisatip` directory you just cloned

retrofan1979:
O ofcourse, sorry.
Yes, it works now! Finally, thanks!

retrofan1979:
Another question: can you please also let minisatip set qam parameter if it's not specified by client?
Normally it's qam256 at default, but sometimes it's qam64.

Jalle19:
> Another question: can you please also let minisatip set qam parameter if it's not specified by client? Normally it's qam256 at default, but sometimes it's qam64.

Can you open a separate issue for that?

retrofan1979:
just did

catalinii (approved):

PR #990, opened 2022-08-14
Refactor utils.h into multiple smaller units

Jalle19:
It seems like utils have grown over time and it has slowly become so integrated that almost every module requires it in one way or another. I figured I'd get the ball rolling on factoring out things into separate modules/files.

This is really just a first step - there's more that could be factored out.

@catalinii what do you think? I think it could help other people get a grasp of the codebase more easily if we untangle it a bit? Certainly helps me!


catalinii (approved):
I was thinking for a long time that improving readability of the code can allow more people to potentially contribute as I think the code is hard to read at the moment...but never got to it.

Thanks for putting the PR up

Jalle19:
I'll squash the fixups and will runtime test this for a few days before merging

Jalle19:
Note to self to add copyright block comments to some of the new files

Issue #992, opened 2022-09-16
BUG: Player glitches when using "satipc"


lars18th:
Hi,

I've advanced a litte more with the trouble that I've described a lot of time ago: When connecting with multiple users at the same time, in some cases the already playing user observes a glitch in the play when the second client starts.

Here my configuration:
- Two DVB-S2 enigma boxes with minisatip running on it as SAT>IP servers.
- One PC running minisatip as a central SAT>IP server, serving to clients (it uses "satipc" with RTSP-over-TCP).
- Different clients using the HTTP protocol (mainly VLC).

Here how yo reproduce the glitch:
- Start minisatip in the PC.
- After 30s. (when all tuners will close) connect the first client (VLC) to one tuner (could be any FTA channel).
- After 15s. connect a second VLC instance to one FTA channel in a different transponder (then the second tuner will be started).
- For 1sec. the first client shows a glitch in the play.
- However, only small CC errors will appear in the stream. All streams are more or less clean, but the first client receives the stream with an interval of ~1000ms of missing data.
- Note: If the second tuner is already open (in sleeping mode), then the glitch doesn't appear at all.

I see in the LOGs (with -v satipc) that the thread that is MANAGING data from the first SAT>IP client, is stalled for sometime when the thread of the second tuner makes the call to the other SAT>IP server. So more or less you see (1 is first satipc thread, 2 is the second one):

`1111111111111111111111111111111122222222222222222222222222221212121212121212121212121212121212`

But what you're expeting is something like : 

`1111111111111111111111111111111122121212121212121212121212121212121212121212121212121212121212`

Therefore, I suspect that in the implementation of the "satipc" module, something is blocking the READING SOCKET, or the writing in this socket, I don't know. Any idea @catalinii ?

catalinii:
hi @lars18th,

I did not see anything that can cause this. To get more info please add -l lock and you can see the messages related to locking maybe something comes up.
Feel free to upload the logs as well.
Thanks

lars18th:
Hi @catalinii ,

I doing a deep trace of the logs. However, something is strange: For every NEW stream, I see that after the first PLAY command (the remote SAT>IP server is minisatip too), the "satipc" module send _THREE_ consecutive DESCRIBE commands:
`"DESCRIBE rtsp://192.168.1.90:554/stream=1 RTSP/1.0"`
[...]

PR #997, opened 2022-10-05
Fix satipc module

lars18th:
This patch fixes some small issues in the "satipc" module:
- Removes redundant thread call.
- Set correct DVR demux buffer size when using RTSP_OVER_TCP.
- When reopening the RTSP channel to a minisatip SAT>IP server send anyway the initial OPTIONS message.

In addition it adds two simple new logs to the adapter module.


lars18th:
Hi @catalinii ,

This patch is based on some considerations described in #992. However, it's completly independent. So please merge it.

Regards.

catalinii (approved):

PR #999, opened 2022-10-07
Expose command line and CAM information to web interface

### LINKED ISSUE #981: Show command used to run minisatip on the status page
The footer could contain e.g. `-f -g -j 0:4-984,1:5-1020,2:6-1056,3:7-1092 -Z *:0 -S 0-2:0,3:3 -A 0:0:0,0:1:0,0:2:0,1:3:1`

### LINKED ISSUE #982: Show service CAID in Cryptinto column
At least when using CI modules this piece of information is not present anywhere.

Jalle19:
This PR consists of four distinct parts:

* Extracting the API related functions to a separate module (symbols, variables). Now that it's separated it's easier to potentially refactor it in the future (e.g. the bandwidth handling still has some nasty cross-dependencies between modules).
* Exposing CAM information via new "symbols". I figured it's better to use a separate symbol array for these since CA support can be disabled at compile-time.
* Storing and exposing the command-line used to start the application
* Updating the web interface to expose the newly added information

I haven't been able to runtime test the CA changes yet, only modified the state locally in the web interface to emulate a CI device.

Screenshots:
Screenshot-20221007105639-248x127
Screenshot-20221007104212-681x147

When compile testing this on a different computer (with GCC 10 instead of GCC 9) I noticed that the GCC 10 complains about multiple definitions of variables by default (https://gcc.gnu.org/gcc-10/porting_to.html), so I figured I'd add the `-fno-common` option to Makefile.

Closes #981, closes #982


src/ca.c
```diff
@@ -2885,17 +2887,13 @@ char *get_ca_pin(int i) {
 void set_ca_pin(int i, char *pin) {
     if (!ca_devices[i])
         ca_devices[i] = alloc_ca_device();
-    if (!ca_devices[i])
-        return;
```
Jalle19, line 2889: I dropped these early returns because there was at least one place where this was _not_ done, so I figured it's pointless since we'll crash anyway.

src/ca.h
```diff
@@ -206,6 +206,9 @@ struct ca_device {
     uint64_t datetime_next_send;
 };
 
+extern ca_device_t *ca_devices[];
```
Jalle19, line 209: Had to move this here since the newly added test case uses it

Jalle19:
AFAICT I fixed Coverity CID 400978.

Jalle19:
@catalinii ready for review

src/ca.c
catalinii: It is not really pointless... my tests on no free memory shows that minisatip still works fairly good.

catalinii (approved):

src/ca.c
Jalle19: I can drop the change and instead add the check where it's missing?

Jalle19:
Still need to find some time to test this before merging.

Jalle19:
@pannlala found some use for your reader after all!

@catalinii turns out the logic that saved which CAIDs each CAM supports was broken. It's fixed in the latest commit.

Normal operation (no `-c` parameter):
Screenshot-20221007221613-549x194

Using `-c` to limit the list of supported CAIDs (`-c 1:1-0100-0500`:
Screenshot-20221007221649-297x167

src/ca.c
catalinii: Sounds good

src/ca.c
```diff
@@ -2977,3 +2978,34 @@ void set_ca_multiple_pmt(char *o) {
         }
     }
 }
+
+char *get_ca_caids_string(int i, char *dest, int max_len) {
+    if (!ca_devices[i])
+        ca_devices[i] = alloc_ca_device();
+    if (!ca_devices[i])
+        return NULL;
```
Jalle19, line 2995: @catalinii added the check here and dropped the earlier commit that removed these checks

src/minisatip.c
```diff
@@ -565,6 +567,25 @@ By default every CAM supports 4 channels\n\
     exit(1);
 }
 
+char* get_command_line_string(int argc, char *argv[]) {
+    int i;
+    size_t len = (argc * 2) - 1; // make room for ", " between arguments
```
Jalle19, line 572: Not sure what I was thinking here, should be argc - 1.

catalinii:
Please squash the commit that is merged when merging for the history to be cleaner.

Jalle19:
Will do next time!

PR #1001, opened 2022-10-09
Implementation of smooth undecrypted packets

lars18th:
Now when "drop_encrypted == 2" instead of filter out all non-PCR packets, we convert all undecrypted packets to TS packets with stuffing removing the PES payload. This will maintain the number of TS packets and the CC counters. Therefore, the resulting bitrate is identical to that of the original encrypted stream. And that it's more easy to handle for regular hardware and software decoders. Producing then a more smooth play when some part of the stream can't be decrypted.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -348,6 +348,7 @@ Help\n\
 	- note: * as adapter means apply to all adapters\n\
 \n\
 * -E Allows encrypted stream to be sent to the client even if the decrypting is unsuccessful\n \
+       Duplicating it (-E -E) all unencrypted packets are send as stuffing TS packets. Usefull for regular player clients.\n\
 	- note: when pids=all is emulated this pass NULLs too\n\
 \n\
 * -Y --delsys ADAPTER1:DELIVERY_SYSTEM1[,ADAPTER2:DELIVERY_SYSTEM2[,..]] - specify the delivery system of the adapters (0 is the first adapter)	\n\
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @catalinii ,

I'm using this patch for around 4 months. Previously I'm using another one for one year that recreates PTS and DTS values. However, this is be far more robust with Android TV, Fire TV, VideoLAN, FFmpeg and similar decoders. So I want to share it with others. My recomendation is to default enable it (with `-E -E`), but I prefer to not change the semantics of the already implemented parameters, so for this reason the current use of the `-E` parameter (with one instance and without it) continues equal.

Regards.

Jalle19:
Would there be any downside to doing this the new way when -E is specified? That way we wouldn't need two different modes.

lars18th:
Hi @Jalle19 ,

No downside on use this alternative mode. But one difference exists between them: the current mode 1 reduces the bitrate to the minimum, while the new mode 2 maintains the original bitrate. Therefore, if the user wants to do some external check based on bitrate to identify/detect the error, then the mode 1 continues to be valid. This ancient mode (implemented by myself) only pass PCR timestamps because they're mandatory to not break the specs of a MPEG-TS. So, both modes are usefull. And because I don't want to break anything I prefer to leave the same behaviour when the users enable the `-E` parameter.

In summary, it seems to me IMHO that we can leave both modes.
You agree?

Jalle19:
Sounds good

Jalle19 (approved):

Jalle19:
@catalinii can we merge this?

catalinii:
Ues thanks for the reminder

PR #1002, opened 2022-10-11
Define character encoding when serving state.json

### LINKED ISSUE #398: Web-GUI: german umlauts not working in channel names
German umlauts like ü and ö are shown as question mark in column of the stream.
Tested with Google Chrome and Mozilla Firefox.

umlauts

Looks like this names are written with JavaScript, I am not sure if utf8 is possible...

Jalle19:
Fixes #398

Before:
Screenshot-20221011145440-1245x388

After:
Screenshot-20221011145139-1243x348

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -1558,8 +1558,10 @@ int read_http(sockets *s) {
     if (strcmp(arg[1], "/state.json") == 0) {
         char *buf = malloc1(JSON_STATE_MAXLEN);
         int len = get_json_state(buf, JSON_STATE_MAXLEN);
+        // Browsers will default to UTF-8 if no character set is specified, and 
+        // DVB networks rarely seem to use UTF-8 so let's go with ISO-8859-1 here.
         http_response(s, 200,
-                      "Content-Type: application/json\r\nConnection: close",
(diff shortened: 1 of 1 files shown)


Jalle19:
@lars18th @Yuri666 care to test this? I have _very few_ service names with "special characters" in them so it's hard to tell if this is a definite fix or not.

catalinii (approved):

Yuri666:
Not work for my provider:
image
To my provider needs charset=ISO-8859-5:
image

Jalle19:
I think this needs a more complex fix then. The service name needs to be converted to Unicode from the character set indicates in the service information. I'll see if the logic can be easily copied from e.g. tvheadend's implementation.

Jalle19:
This is more than I bargained for so I'll close this for now

catalinii:
Most of tvh implementation seems to be here.
I think the charset is specified as part of the string. Before we bring something complete maybe we can use the encoding that is defined in the SDT. At least for the ppl watching channels from one country it will work.

catalinii:
https://github.com/tvheadend/tvheadend/blob/5f9404117f59ad1f5aa7ca542ce39d9e064e8209/src/input/mpegts/dvb_support.c#L260

PR #1006, opened 2022-10-14
Update README.md

lars18th:
After PR #1001 the `tools/gen_readme` script it's necessary to update the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ make
 Usage:
 -------
 (Message automatically generated from "minisatip --help")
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii and @Jalle19 ,

Perhaps an autoupdate of the README.md with github actions could be interesting. What you think?

Jalle19:
In my opinion it would be better if we'd stop printing the whole usage text in the README, then there would be nothing to keep up to date :shrug:

Jalle19:
Especially since different compilation flags yield different available options

PR #1014, opened 2022-10-23
Include the git revision in the version number

Jalle19:
@catalinii what do you think? IMO this is helpful.


Jalle19:
```
 make[1]: Entering directory '/__w/minisatip/minisatip/src'
fatal: detected dubious ownership in repository at '/__w/minisatip/minisatip'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/minisatip/minisatip
fatal: detected dubious ownership in repository at '/__w/minisatip/minisatip'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/minisatip/minisatip
```

This seems to be a recent regression in GitHub runners (see https://github.com/actions/runner/issues/2033), added a fix for it in the Debian packaging PR already, will add here as well.

catalinii:
I am ok if you add it. The purpose of minor minor version was to track individual commits.

But the general issue is that people sometimes build outside of git directories or directly from zip files (such as LibreELEC, ... ) and then the version is messed up.

catalinii (approved):

Jalle19:
Indeed, non-standard builds end up with "1.2." as version number now, with this change it would at least contain the git hash.

I'm thinking of improving this further with a runtime generation of the version string instead of the macro used now. That would eliminate the "1.2." issue as well.

PR #1013, opened 2022-10-22
Automated Debian packaging

Jalle19:
Have to PR this to run the GitHub workflow, still work in progress

PR #1016, opened 2022-10-25
Update Script install minisatip and update README.md Autostart after system reboot

dnmnhat:


Jalle19:
Closing this in favor of https://github.com/catalinii/minisatip/pull/1013, the end goal is to make it easy to install minisatip just like other Debian/Ubuntu packages. Getting the packages built automatically is the first step towards that.

PR #1021, opened 2022-11-07
Use demux filters for pmts as previously. This fixes PMT processing on vu+

catalinii:
This also fixes ddci tests.


catalinii:
cc @Jalle19

Jalle19:
Nice to see that the build passes, thanks for fixing!

PR #1022, opened 2022-11-21
Add pid 18 to be sent to the DDCI

catalinii:


Jalle19:
What does this do?

catalinii:
It maps pid 18 to the first pmt. My CAM needs this to update regularly.

Jalle19:
Getting a lot of this after this commit:

```
PID Continuity error (adapter 0, pos 132): pid: 1016, Expected CC: 8, Actual CC: 4, CC Before 7
```

catalinii:
I get that too ...was thinking is specific to my provider/ setup

catalinii:
Will look into it

Jalle19:
Saw that you reverted this. Is ddci_create_epg() really needed? It seems pretty arbitrary.

catalinii:
Certain CAMs does not request pin number when switching the channel when ddci_create_epg is used. Will investigate more when my CAM disables next time.

Jalle19:
What if you add PID 18 like you did but disable ddci_create_epg()?

catalinii:
I am not even sure pid 18 helps my case. But if I do that, I will choose for my CAM to work (which is an old model) over an official CAM that is used a lot across Europe.

PR #1030, opened 2023-01-10
Add Tuner status type enforcement

lars18th:
For some adapters the DVB API version 5.5 is not available. In these cases the software can't distinguish the status reporting of the tuners. Therefore, the values of Strength and/or SNR could be incorrect. If you know how the driver of the tuner reports the data (in percentage or in dB), then you could tweak the values with the `-M --multiplier` parameter.

This patch adds the option to enforce percentage or decibel modes at tuner level (because you can mix different hardware and drivers in the same machine). Example:

`-M 0:%1.0-#1.0 -M 1:%1.0-#1.0 -M 2:2.0-1.0 -M 3:2.0-1.0`

In this case the first two tuners (DVB-S) use default values enforcing "percentage" for Strength and "dB" for SNR. And for the other two tuners (DVB-T2) the types aren't enforced but the Strength is doubled.

This removes all dependencies for Enigma2 boxes, and you will have the opportunity to set the correct types/values for all of your tuners.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -1964,7 +1964,7 @@ void set_signal_multiplier(char *o) {
     int i, la, a_id;
     float strength_multiplier, snr_multiplier;
     char force_tuner_signal = TUNER_FORCE_NO;
-    char force_str[32];
+    char force_str[128];
     char buf[1000], *arg[40], *sep1, *sep2;
     adapter *ad;
     SAFE_STRCPY(buf, o);
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -108,6 +108,8 @@ adapter *adapter_alloc() {
 
     ad->strength_multiplier = opts.strength_multiplier;
     ad->snr_multiplier = opts.snr_multiplier;
+    ad->force_strength_mode = 0;
+    ad->force_snr_mode = 0;
     ad->db_snr_map = 1.0;
 
     ad->drop_encrypted = opts.drop_encrypted;
@@ -1962,6 +1964,7 @@ void set_adapters_delsys(char *o) {
 void set_signal_multiplier(char *o) {
     int i, la, a_id;
     float strength_multiplier, snr_multiplier;
+    char force_strength_mode, force_snr_mode;
     char buf[1000], *arg[40], *sep1, *sep2;
     adapter *ad;
     SAFE_STRCPY(buf, o);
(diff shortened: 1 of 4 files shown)


lars18th:
Hi @catalinii and @Jalle19 ,

This PR is the final step of the work started at #911. Please, check it. I'm using it in my two E2 stb with good results.

Regards.

src/dvb.c
```diff
@@ -1733,13 +1733,22 @@ void get_signal(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
         }
     }
 
-    if (opts.enigma) {
-        // In Enigma2 STRENGTH:[0..100] and SNR:dB*1000
-        int64_t strengthd = 0, snrd = 0, init_snr = 0;
-        float tf;
+    if (ad->force_strength_mode == 1) {
```
catalinii, line 1736: As with the previous code, this is not the right place to change the values returned by the hardware.
This method should only get the values from hardware.

I propose adding this code to a separate method and calling it from here: https://github.com/catalinii/minisatip/blob/master/src/dvb.c#L1875

src/adapter.h
```diff
@@ -64,6 +64,8 @@ struct struct_adapter {
     uint16_t db;  // if MAX_DB then no value, else value is dB*10 of the adapter
     float strength_multiplier, // final value: strength * strength_multipler,
         snr_multiplier;        // same for snr
+    char force_strength_mode,  // 0: No force (default) 1: Force Relative 2: Force Decibel
```
catalinii, line 67: The values used here should be used as constants rather than a comment here... the constants should be used to actually decide the final value.

src/dvb.c
```diff
@@ -1733,13 +1733,22 @@ void get_signal(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
...
+        // STRENGTH: [0..100]
+        int64_t strengthd = 0;
 
         strengthd = (*strength * 65535) / 100.0;
 
+        *strength = (int)strengthd;
+    } else if (ad->force_strength_mode == 2) {
```
catalinii, line 1743: Please implement also this part.

src/dvb.c
```diff
@@ -1751,8 +1760,9 @@ void get_signal(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
             init_snr = 999; // No more than 99.9 dB
 
         *db = (int)init_snr;
-        *strength = (int)strengthd;
         *snr = (int)snrd;
+    } else if (ad->force_snr_mode == 1) {
```
catalinii, line 1764: same here

src/dvb.c
lars18th: Done!

src/adapter.h
lars18th: Done!

src/dvb.c
lars18th: Done!

src/dvb.c
lars18th: Done!

lars18th:
Hi @catalinii ,

All suggested changes are now implemented. So please, check it! 😉

src/adapter.c
```diff
@@ -2004,9 +2018,18 @@
                     a[j]->snr_multiplier = snr_multiplier;
                 }
         }
-        LOG("Setting signal multipler for adapter %d strength_multiplier %.2f "
+        if (force_tuner_signal) {
+            sprintf(force_str, " (forcing driver:%s%s%s%s)",
```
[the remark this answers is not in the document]
lars18th: Fixed!

src/dvb.c
```diff
@@ -1873,6 +1932,9 @@ int dvb_get_signal(adapter *ad) {
 
         if (ad->new_gs == OLD_SIGNAL)
             get_signal(ad, &status, &ber, &strength, &snr, &dbvalue);
+
+        if (ad->force_tuner_signal != TUNER_FORCE_NO)
```
catalinii, line 1936: Thanks, the PR looks great.
Do you want adapt_signal to be applied to both cases (get_signal and get_signal_new) ?

lars18th: Hi @catalinii ,

> Do you want adapt_signal to be applied to both cases (get_signal and get_signal_new) ?

I don't think so! The reason is because the `get_signal_new()` function uses the driver information. And we need to trust the driver. The problem is with the `get_signal()` function in which the driver doesn't indicates the type of data. Then in this case the user requires to "tweak" the information. For example, I have one E2 box with four tuners, and each one requires different tweaks... because it's driving different physical tuners with different drivers. Therefore this PR enables the correct configuration. I hope you will merge it soon.

And please note that the duplicated code (i.e. `get_snr_decibels()` and `get_strength_decibels()`) is already used in `get_signal_new()`. So from my point of view the PR is complete.

Regards.

catalinii: The code above allows it to execute even when new_signal is used

lars18th: > The code above allows it to execute even when new_signal is used

You're right! Now it's fixed. Only executed when calling to "old" `get_signal()`.

Ready to merge it?

catalinii: Yes thx

lars18th:
Hi @catalinii and @Jalle19 ,

Ready to merge it?

PR #1031, opened 2023-01-12
Hide HTTP response of a reGET

lars18th:
When one HTTP client wants to change the pid list it could send a new GET command identical to the first one with different pids. The current sever code can handle this non-standard use of the HTTP protocol. But without this extension the HTTP response is introduced in the middle of the streaming. And this generates noise from the client point of view. This new implementation changes the behaviour checking if the client sends multiple GET commands. In this case the response for the successive requests is automatically hidden. The result is that the streaming continues from the server to the client without any interruption.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -2020,6 +2020,14 @@ void http_response(sockets *s, int rc, char *ah, char *desc, int cseq, int lr) {
     else
         proto = "RTSP";
 
+    // Check if the request is a "reGET" (successive non-standard HTTP request while streaming)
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii ,

This is based on the same concept of my patch #576 created years ago. But more simple and robust. I know that not all users use this functionality. But I'm using it for years without troubles. Now, with this patch any client that wants to use HTTP only (instead of RTSP over TCP) could change the pid list without noise in the streaming.

I hope you want to merge it.
Regards.

Issue #1011, opened 2022-10-22
BUG: Tuning to the same frequency with some default parameters implies the solection of a different tuner.


lars18th:
Hi @catalinii ,

I discovered that when using the "satipc" module, if some client requests a transponder with a small _different_ parameters, but for the same frequency, then the "satipc" module requests a different SETUP for the remote SAT>IP server. Then this will open a different tuner, even for the same frequency. This is for example what receives the remote server:
```
SAT>IP: SETUP rtsp://192.168.1.53:554/?src=1&freq=11236&msys=dvbs&mtype=qpsk&pol=h&sr=0&pids=none RTSP/1.0
SAT>IP: SETUP rtsp://192.168.1.53:554/?src=1&freq=11236&msys=dvbs&mtype=qpsk&pol=h&sr=27500&fec=56&ro=0.35&plts=off&pids=none RTSP/1.0
```
Why not "compact" the request in this case?

The first SETUP is sended by the "satipc" module when the client requests only "src", "freq", "pol", "msys" and "mtype", with default values for the rest. And the second SETUP correspond to the request from the "satipc" module when a different client wants the same transponder but indicates all values ("sr", "fec", "ro" and "plts").

These different SETUPs are interpreted as different tunning parameters in the remote SAT>IP server. And then implies to use two tuners.

Any idea to fix this?

Jalle19:
Is the behavior identical (two tuners are used) if `satipc` is not involved? Or will the tuner get reused then?

lars18th:
Hi @Jalle19 ,

I think you're right. After thinking more on it, I suspect that the problem is not related with the "satpic" module, but with the tuner selection. However, I discovered this issue using this module.

More or less, the problem could be related with the function `compare_tunning_parameters()`:
https://github.com/catalinii/minisatip/blob/792820fab4e8bd722e5a47fc378371fe91bd1902/src/adapter.c#L1206-L1226

So, the solution perhaps implies to store the "default" (aka "not indicated") value in some part, and use this information to compare the tunning data.

What you think @catalinii ?

catalinii:
Probably the issue is in compare_tunning_parameters but the values of fec, ro, plts are not compared. Can you print if compare_tunning_parameters is really the issue?

lars18th:
> Probably the issue is in compare_tunning_parameters but the values of fec, ro, plts are not compared. Can you print if compare_tunning_parameters is really the issue?

Hi @catalinii ,

[...]

PR #1037, opened 2023-01-17
Fix compare tunning if SR=auto in current tunning

lars18th:
When comparing new and current tunning values, the SR could be AUTO. Not only for the "new" tunning, but also for the "current" tunning. This fixes the tunning when the "current" adapter tuning is using AUTO mode for SR, and the "new" request is forcing some value. In this edge case, we can consider that the current running value is the one corresponding to the current tunning mode. That's we trust the new request.

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -1218,7 +1218,7 @@ int compare_tunning_parameters(int aid, transponder *tp) {
         get_absolute_source_for_adapter(aid, tp->diseqc, tp->sys) !=
             ad->tp.diseqc ||
         (tp->pol > 0 && tp->pol != ad->tp.pol) ||
-        (tp->sr > 1000 && tp->sr != ad->tp.sr) ||
+        (tp->sr > 1000 && ad->tp.sr > 1000 && tp->sr != ad->tp.sr) ||
         (tp->mtype != 6 && ad->tp.mtype != 6 && tp->mtype != ad->tp.mtype))
 
         return 1;


lars18th:
Hi,

This fixes #1011. The only problem that I can see, it's if the new request is enforcing a different SR mode that the _real_ currently used. For example, if some client is receiving a valid stream and it has tunned it using SR=AUTO. And after another client is doing some scaning over the same frequency and it uses a SR that _is different_ from the real one. Only in this case this patch could generate a trouble. So I assume that this very edge case could not be considered. However, if someone understand that we need to protect to it, then it will be necessary to store the _real_ tunning parameters when requesting with the AUTO mode for some options. In this case, if the client requests SR=AUTO, then the adapter will complete the information this information. And therefore this patch will can't generate any side effect, as the adapter->sr value will never be 0 (aka auto).

Please test it and merge.
Regards.

Jalle19:
Can you add a test case for this to https://github.com/catalinii/minisatip/blob/master/tests/test_dvb.c ?

lars18th:
> Can you add a test case for this to https://github.com/catalinii/minisatip/blob/master/tests/test_dvb.c ?

I'll try.

lars18th:
Hi @Jalle19 ,

Because I'm not an specialist writing Unit Tests, I prefer if someone write them. In the meantime, please merge this patch. The only edge case not considered is when a new client requests the same transponder with specific SR, and at the same time other user is using it with AUTO _and_ the SR from the new client is not the same. In this case, nothing is interrupted. The new user will be attached to the tuner used by the current user. The problem is that the new client will receive the stream _using an incorrect value of the SR_ in the request. The true is that the request in fact requires to fail because the incorrect tunning parameters. But it's a very edge case... and perhaps it could be a problem in a blind scaning. But I feel it's preferable to overcome this case and merge the patch.

Jalle19 (approved):

lars18th:
Hi @catalinii ,

This PR is approved by @Jalle19 . Any comment from you? From my point of view it's ready to merge it.

catalinii:
I agree with @Jalle19 related to unit tests and in general we should this about this aspect too.

Generally is the responsibility of the one submitting a patch to add unit tests

lars18th:
> I agree with @Jalle19 related to unit tests and in general we should this about this aspect too.
> 
> Generally is the responsibility of the one submitting a patch to add unit tests

OK. But I don't have sufficient experience with them. I'll try in the future to implement one.

Regards.

Issue #1034, opened 2023-01-15
Request: Revert some changes on commit 2fb66a7


lars18th:
Hi @catalinii ,

I have some comments about changes in this commit: https://github.com/catalinii/minisatip/commit/2fb66a703e0d00d7eb99222abc704bb05a18bc0a

One of the changes is related to part of my fix in PR #997. I've spend years of use to discover that the `satipc` module will not work sufficient reliable because the receive buffer has not updated with a correct size. The default _network_ buffer size is incompatible with the work done by the `satipc`. Futhermore, the `dvb` module continues having the option to set the buffer size. And this has sense because the user requires to tweak it based on the physical hardware used. However, for the `satipc` (in my environment the most relevant and important part of the tool) you've now removed the option to adjust correctly the size of the DVR reading buffer. The commit log indicates that the user can adjust it using _system_ variables (`/proc/sys/net/core/rmem_default` and `/proc/sys/net/core/wmem_default`). But this is a false statement. Almost I see these cases in which is false:
- When minisatip is running inside a server not administred by the user.
- When minisatip is running in a Docker container.
- When minisatip is running in a server shared with other services that require different values for these system variables.

And perhaps I forgot to mention some other cases.

So, please consider one of these two options:
1) Revert the part of this commit that removes the `satipc` adjust to the DVR buffer. In fact the lines: 571-572 *and* 589-590.
2) Or add the option to adjust independently the DVB read buffer and the SATIPC read buffer.

Sorry if this sounds quite rude. It's not my objective to say something incorrect. I only want to comment that this generates a lot of side effects that perhaps you don't have considered. And perhaps the best is to implement the option 2. What you think?

Regards.

catalinii:
@lars18th opposing opinions as long as they are respectful are quite welcome.

The reason I removed those lines are because I noticed issues in my setup.
My DVB hardware using minisatip server is 190ms away from the minisatip satipc. I have optic fiber on both ends with no bandwidth issues in between.

If you look at the kernel code: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/core/sock.c?h=v6.2-rc4#n1638 SO_RVCBUF sets sk->sk_rcvbuf, for which the default value is rmem_default: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/core/sock.c?h=v6.2-rc4#n3390.
[...]

PR #1038, opened 2023-01-18
Restore socket buffer configuration

lars18th:
This patch restores the lost option to set the socket buffers. However, now it's possible to set independently the read socket buffer for the satipc connections, and the write socket buffer for the clients. By default the buffer uses the kernel configuration. However, with two new parameters the user can overwrite them.


lars18th:
Hi @catalinii ,

Ready to merge! This PR includes all your requrements discussed in #1034 .

lars18th:
Hi @Jalle19 ,

Please, could you test it too!

catalinii:
THabks @lars18th

PR #1040, opened 2023-01-19
Store in the UI status previous data

lars18th:
In the HTML status page now the previous status.json data will be stored at every iteration. Then it will possible to print difference of values betwen current and previous state data. One use of this is to print the decoding errors only from the last iteration.

### DIFF
--- a/html/status.html
+++ b/html/status.html
@@ -358,7 +358,7 @@
 			}
 		}
 
-		function getTable(state) {
+		function getTable(state,oldstate) {
 
 			if (!state)
 				return '';
@@ -597,8 +597,10 @@
 				}
 
 				var decerr = state['ad_decerr'][i];
+				var decerrpre = oldstate['ad_decerr'][i];
+				decerrpre = decerr - decerrpre;
 				if (decerr > 0)
-					myTable += "DEC-ERRORS: " + decerr + "<br />";
+					myTable += "DEC-ERRORS: " + decerr + " (" + decerrpre + ")<br />";
 				myTable += "</small></td></tr>";
 			}
 
@@ -623,6 +625,7 @@
 		var stats = 0;
 		var statsrefresh;
 		var table = null;
+		var oldData = null;
 
 		var hashTag = location.hash;
 		if (typeof hashTag.split('#')[1] !== "undefined") {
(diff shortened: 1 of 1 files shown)


html/status.html
```diff
@@ -667,7 +672,8 @@
 					"emptyTable": "No tuner found/active"
 				}
 			});
-
+			delete oldData;
```
Jalle19, line 675: This does nothing and can be removed

html/status.html
```diff
@@ -358,7 +358,7 @@
 			}
 		}
 
-		function getTable(state) {
+		function getTable(state,oldstate) {
```
Jalle19, line 361: Can you please add spaces between parameters (same thing where this function is called further down)?

html/status.html
```diff
@@ -597,8 +597,10 @@
 				}
 
 				var decerr = state['ad_decerr'][i];
+				var decerrpre = oldstate['ad_decerr'][i];
+				decerrpre = decerr - decerrpre;
 				if (decerr > 0)
-					myTable += "DEC-ERRORS: " + decerr + "<br />";
+					myTable += "DEC-ERRORS: " + decerr + " (" + decerrpre + ")<br />";
```
Jalle19, line 603: Should this logic be used for CC/BER counters too?

html/status.html
lars18th: Fixed! So please resolve

html/status.html
lars18th: Fixed! So please resolve

html/status.html
lars18th: It could be done...

Jalle19: Sure

html/status.html
```diff
@@ -489,8 +489,11 @@
 					if (ber > 0)
 						signal += "<div class='level'>BER <progress class='pber progress-1' value='" + ber + "' max='1000' />" + ber + "</div><br />";
 					var ccerr = state['ad_ccerr'][i];
+					var ccerrpre = oldstate['ad_ccerr'][i];
+					ccerrpre = oldstate['ad_ccerr'][i];
```
Jalle19, line 493: This line is superflous, the variable is already assigned this value on the line above

Jalle19 (approved):

lars18th:
Thank you @Jalle19 !

I hope someone will add more functions to the UI using this. 😉

Issue #940, opened 2022-02-11
[Req] Improve client info in UI


lars18th:
Hi,

I request if someone wants to implement this in the UI:
- At time for each used tuner a simple "client:" tag is printed when the client is HTTP and provides an identification header. Hover, this info it's insufficient to know how is using a tuner when multiple clients/tuners are in use.
- So the idea is to add more info: source address, src-port, protocol, name. Example
  - `Client: 192.168.1.100:10554/RTSP-UDP/vlc`
  - `Client: 192.168.1.101:50001/HTTP/Chrome`
  - `Client: 192.168.1.102:34567/RTSP-TCP/minisatip`
- And _optionally_, if this could be added, also the list of requested pids fron each user with a mapping of the corresponding PMTs to the name of the channel. Example:
  - `Client: 192.168.1.102:34567/RTSP-TCP/minisatip - TV1HD,TV2HD`
  - `Pids (minisatip): 0,16,18,20,21,100,101,200,201`

And also for the Tuner status of satipc modules, instead of printing only the server ip address, add the port and the protocol too. This last will be useful to know if the connection is configured for UDP RTSP or RTSP-over-TCP. And the port to distinguish different daemons (running of example in the loopback).

Thank you.

lars18th:
I forgot to mention to add a button or link for closing a specific client.

You agree?

Jalle19:
I'd like more client info as well, but personally I'd prefer it if the web interface was strictly read-only, i.e. no way to control anything. Kicking a client will only make it re-subscribe a moment later for example.

lars18th:
Hi @Jalle19 ,

I've already implemented this, without the option to disconnect clients that you prefer. Please, check it and merge.

Regards.

lars18th:
I close it because the enhacement is completed.

PR #1042, opened 2023-01-24
UI enhancement: show client protocol

lars18th:
Adds support for displaying the protocol used by each client to connect.

### DIFF
--- a/html/status.html
+++ b/html/status.html
@@ -508,7 +508,8 @@
 				if (state['has_pmt']) {
 					myTable += "<b>" + state['ad_channel'][i] + "</b><br>";
 					var st_pmt = state['ad_pmt'][i];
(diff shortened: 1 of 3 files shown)


lars18th:
Hi,

This PR implements the pending enhancement described in #940.

lars18th:
And as a bonus it shows the PMT list of channels too.

Jalle19:
Looks good, will try this myself before merging

lars18th:
As a second bonus I've added a more compact size of the columns.

Jalle19 (commented):
Can you please use CSS classes instead of inline styles?

lars18th:
> Can you please use CSS classes instead of inline styles?

No because I don't know how to do it. If you know it, then please modify it. Sorry.

PR #1055, opened 2023-02-09
Handle CWs out of cycle

lars18th:
This patch handles a very edge case: When you receive a CW with the same parity of the current one.

The standard behavior is to only receive keys for the next parity cycle. However, in some rare cases a key using the same parity can be received. In principle we do not need to discard this key because we will do a check on the next key change. However, in some special case this could generate an error. This potential case appears when during the key change the buffer does not contain any PUSI packets to validate the key AND the packets are mixing both parities. In this special case, the selection of the "new" key for the "old" parity will not work. Therefore, in this case (and only in this case) it is preferable to use the old key that actually was working.

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -934,6 +934,11 @@ int send_cw(int pmt_id, int algo, int parity, uint8_t *cw, uint8_t *iv,
     c->pmt = master_pmt;
     c->cw_len = 16;
     c->time = getTick();
+    if (parity == pmt->parity && pmt->cw && pmt->last_update_cw > 0) {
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @catalinii and @Jalle19 ,

Please, test it. I'm using it for some weeks, and warnings appear time to time. And when they appear, I don't see any glitch during the play. So IMO it's safe to merge it.

catalinii:
I think is a good idea, but the block should be after setting the expiry not before. Otherwise the new key will expire before the old key.

catalinii:
Thx @lars18th

PR #1059, opened 2023-02-18
Use the new FIFO implementation for DDCI

catalinii:
- Adds a new FIFO implementation
- Extend it for storing the data read from the DDCI device

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -50,7 +50,8 @@ SOURCES=\
 	utils/alloc.c \
 	utils/hash_table.c \
 	utils/mutex.c \
-	utils/ticks.c
+	utils/ticks.c \
+	utils/fifo.c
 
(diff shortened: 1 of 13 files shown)


lars18th:
Hi @catalinii ,

This patch seems interesting. You have the idea to update the `satipc` input module with this new FIFO implementation? And what about the processing internal buffer?

catalinii:
I was thinking about using fifo for the satipc module (for sending the messages instead of the hash table), but I think a better option is to refactor and prevent the satipc module from buffering responses.
This would be better because of a message is buffered any update to the list of pids for example, requires a new message. If there is no buffering the satipc module can just share it's current state rather than an older state.
I plan next to replace the socket buffering with FIFO, as based on current consumption it uses 2x-4x more memory that usable.

Issue #1007, opened 2022-10-14
Location in SSDP announcements use wrong IP address when -V is used


Jalle19:
If a server has two IP addresses, `10.110.4.10` and `10.110.4.13`, and minisatip is started with `-V 10.110.4.13`, it listens on 10.110.4.13 as expected, but the location of the XML in SSDP announcements is 10.110.4.10 unless `-w` is also specified.

`opts.http_host` should inherit the `-V` value.

lars18th:
Hi @Jalle19 ,

Please check my PR #1064 .

lars18th:
Hi @Jalle19 ,

Time to close this issue?

PR #1064, opened 2023-02-22
Use bind address for the services configuration

lars18th:
Without this PR the bind address configuration is not used to configure all the services. This resolves the issue #1007


Jalle19 (approved):
Thanks!

PR #1026, opened 2022-12-17
Omit unsubscribed PIDs from CA PMT

Jalle19:
CAMs generally have an upper limit to have many PIDs can be decrypted. If two services have a very large number of PIDs, the CAM limit may be reached simply by stuffing the PIDs from two PMTs inside a single CA PMT (which happens for the first two channels when the "-c" parameter is used).

Clients can filter out stream PIDs they are not interested in, e.g. with tvheadend using stream filters. However, minisatip does not take into account whether or not a particular PID for a particular service is actually "subscribed to" on the adapter level when creating the CA PMT. This changes that.

Should increase the likelyhood that a CAM can decrypt more than two channels per CA PMT (i.e. using `-c 4:3` instead of `-c 4:2`).

Here's some comparisons with a channel that happens to have a lot of PIDs (six audio tracks, teletext, and three subtitle tracks):

```
Dec 05 15:14:13 vladimir minisatip[23853]: [05/12 15:14:13.775 AD1]: Dumping pids table for adapter 7, number of unknown pids: 0
Dec 05 15:14:13 vladimir minisatip[23853]: [05/12 15:14:13.775 AD1]: pid 1011, fd 100, packets 178432, d/c errs 0/0, flags 1, pmt 6, filter -1, sock -1, sids: 257 -1 -1 -1 -1 -1 -1 -1
Dec 05 15:14:13 vladimir minisatip[23853]: [05/12 15:14:13.775 AD1]: pid 1013, fd 100, packets 10458, d/c errs 0/0, flags 1, pmt 6, filter -1, sock -1, sids: 257 -1 -1 -1 -1 -1 -1 -1
Dec 05 15:14:13 vladimir minisatip[23853]: [05/12 15:14:13.775 AD1]: pid 1016, fd 100, packets 6696, d/c errs 0/0, flags 1, pmt -1, filter -1, sock -1, sids: 257 -1 -1 -1 -1 -1 -1 -1
Dec 05 15:14:13 vladimir minisatip[23853]: [05/12 15:14:13.775 AD1]: pid 1012, fd 100, packets 5180, d/c errs 0/0, flags 1, pmt 6, filter -1, sock -1, sids: 257 -1 -1 -1 -1 -1 -1 -1
[...]

PR #1071, opened 2023-03-06
Fix omitting unsubscribed PIDs from CA PMT, web interface tweaks

Jalle19:
`get_ca_multiple_pmt()` was called on the stream adapter, not the CI adapter (introduced in https://github.com/catalinii/minisatip/pull/1026). Thus it would always return 0 and all PIDs would unconditionally be added to the DDCI PMT.

Furthermore, after fixing that bug, it turned out that the logic in `CAPMT_add_PMT()` was not necessary - the PMT given to the function will be the DD CI PMT and it will simply not contain any unwanted PIDs, so there's no filtering needed there.

I've updated the web interface to show PIDs for CI adapters too. There's still some improvements to be made there - currently it shows all PIDs, including excluded ones.

image

### DIFF
--- a/src/ddci.c
+++ b/src/ddci.c
@@ -708,8 +708,7 @@ int ddci_create_pmt(ddci_device_t *d, SPMT *pmt, uint8_t *new_pmt, int pmt_size,
     // Add Stream pids
     // Add CA IDs and CA Pids
     for (i = 0; i < pmt->stream_pids; i++) {
-
-        if (get_ca_multiple_pmt(pmt->adapter)) {
+        if (get_ca_multiple_pmt(d->id)) {
             // Do not map any pids that are not requested by the client
             SPid *p = find_pid(pmt->adapter, pmt->stream_pid[i]->pid);
             if (!p) {
--- a/html/status.html
+++ b/html/status.html
@@ -540,7 +540,10 @@
 				// STREAMS
 				myTable += "<td class='dt-left'>";
 				if (state['has_pmt']) {
-					myTable += "<b>" + state['ad_channel'][i] + "</b><br>";
+					if (state['ad_channel'][i]) {
+						myTable += "<b>" + state['ad_channel'][i] + "</b><br>";
+					}
+					
 					var st_pmt = state['ad_pmt'][i];
 					if (st_pmt != "")
 						myTable += "<small>PMT: " + st_pmt + "</small><br />";
--- a/html/status.html
+++ b/html/status.html
@@ -389,6 +389,9 @@
 
 		}
 
+		function formatPids(pids) {
+			return pids.replaceAll(',', ', ');
+		}
 
 		function getTable(state, oldstate) {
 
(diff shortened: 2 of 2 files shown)


Jalle19:
Fixed the tests

Jalle19:
@Yuri666 can you test this branch for any regressions?

src/ddci.c
```diff
@@ -707,9 +707,10 @@ int ddci_create_pmt(ddci_device_t *d, SPMT *pmt, uint8_t *new_pmt, int pmt_size,
 
     // Add Stream pids
     // Add CA IDs and CA Pids
+    adapter *ddci_adapter = get_adapter(d->id);
+    
     for (i = 0; i < pmt->stream_pids; i++) {
-
-        if (get_ca_multiple_pmt(pmt->adapter)) {
+        if (ddci_adapter && get_ca_multiple_pmt(ddci_adapter->id)) {
```
catalinii, line 713: is this not always true ?
Every DDCI adapter has a regular adapter attached to it.

src/ca.c
```diff
@@ -298,10 +298,9 @@ int create_capmt(SCAPMT *ca, int listmgmt, uint8_t *capmt, int capmt_len,
     capmt[pos++] = 0; // PI LEN 2 bytes, set 0
     capmt[pos++] = 0;
 
-    pos += CAPMT_add_PMT(capmt + pos, capmt_len - pos, pmt, cmd_id, added_only);
+    pos += CAPMT_add_PMT(capmt + pos, capmt_len - pos, pmt, cmd_id);
```
catalinii, line 301: there are 2 scenarios in which this is called:
1) DDCI adapter and as you mentioned, the unsubscribed pids are not present in the PMT (for which case the code is correct)
2) Regular CI/CAM: In that case the unsubscribed pids are also present in the PMTs, so we need to remove them.

This code was added for case 2. For 1, this code is supposed to add all the pids present in the PMT generated by the DDCI.

src/ddci.c
Jalle19: Yes it is, but Coverity complained. I can drop the commit if there's a way to mark the issue as a non-issue in Coverity?

src/ca.c
Jalle19: I forgot about the second case, I'll drop this commit then 👍

src/ddci.c
```diff
@@ -707,9 +707,10 @@ int ddci_create_pmt(ddci_device_t *d, SPMT *pmt, uint8_t *new_pmt, int pmt_size,
 
     // Add Stream pids
     // Add CA IDs and CA Pids
+    adapter *ddci_adapter = get_adapter(d->id);
+    
     for (i = 0; i < pmt->stream_pids; i++) {
-
-        if (get_ca_multiple_pmt(pmt->adapter)) {
+        if (get_ca_multiple_pmt(ddci_adapter->id)) {
```
catalinii, line 713: This is d->id

catalinii (approved):

src/ddci.c
Jalle19: Okay, so the call to get_adapter() isn't needed at all?

catalinii: Yeah

Jalle19: Could be that I'll be able to drop the changes to test_ddci.c too then

catalinii: pmt->adapter is different than d->id and I think the change to d->id is correct.

Jalle19: Compiles, but haven't tested it

PR #1084, opened 2023-03-21
Inject a virtual SDT instead of mapping PID 17 from the first PMT

Jalle19:
Should fix accidentally overwriting PID 18 (the virtual EIT) on the DDCI adapter with an SDT that has been mapped from PID 17 to PID 18.

Still not runtime tested, but the test cases pass now.

### DIFF
--- a/src/ddci.c
+++ b/src/ddci.c
@@ -583,7 +583,7 @@ int ddci_create_sdt(ddci_device_t *d, uint8_t *sdt) {
     SPMT *pmt;
     for (i = 0; i < MAX_CHANNELS_ON_CI; i++) {
         if ((pmt = get_pmt(d->pmt[i].id))) {
-            LOG("Adding PMT %d to SDT, sid %d", d->pmt[i].id, pmt->sid);
+            LOGM("Adding PMT %d to SDT, sid %d", d->pmt[i].id, pmt->sid);
             ddci_mapping_table_t *m =
                 get_pid_mapping(d, pmt->adapter, pmt->pid);
             if (!m) {
--- a/src/ddci.c
+++ b/src/ddci.c
@@ -555,31 +555,6 @@ int ddci_create_pat(ddci_device_t *d, uint8_t *b) {
     return len;
 }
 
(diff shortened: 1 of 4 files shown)


Jalle19:
Considering rewriting `ddci_create_epg` to return just the table, then we can use `buffer_to_ts` just like we do for PAT, SDT and PMTs. I also believe it's not necessary to write an EIT for the virtual CA service ID (`MAKE_SID_FOR_CA`), so we could skip that. The main point is that the EIT PID should be present, that's all the standard says.

src/ddci.c
```diff
@@ -762,12 +797,22 @@ int ddci_add_psi(ddci_device_t *d, uint8_t *dst, int len) {
...
     int psi_len;
+
+    // Add PAT
     if (ctime - d->last_pat > 500) {
         psi_len = ddci_create_pat(d, psi);
         pos += buffer_to_ts(dst + pos, len - pos, psi, psi_len, &d->pat_cc, 0);
         d->last_pat = ctime;
     }
 
+    // Add SDT
```
catalinii, line 808: Is best if this is after the PMT because the pmt->name can be populated based on the SDT.

catalinii:
I added this for my CAM, I can test it next time it needs an update.

src/ddci.c
Jalle19: The virtual SDT doesn't contain any service names at the moment, only service IDs and some status flags.

catalinii:
I am fine either way...I will test this later anyway

Yuri666:
Now my old Viaccess CI+ 1.2 CAM get URI and start decrypting after 3-4 secs instead 10-12 secs when without patch. Now decrypting starts at DDCI as fast as in embedded CI

Jalle19:
Wow, that's great!

Jalle19:
Runtime testing this too right now, so far so good

Jalle19:
https://github.com/Jalle19/minisatip/commit/5a7c363a934470ce916942751e047ccb1a27bae2

Jalle19:
Rebased

PR #1088, opened 2023-03-24
Refactor virtual EIT generation somewhat

Jalle19:
* don't write a full TS packet, use `buffer_to_ts` like for our other tables
* move CC counter to the adapter
* don't add EIT for the non-existant SID returned by `MAKE_SID_FOR_CA()`, only add for the SIDs we actually have in the transport stream


catalinii:
The reason MAKE_SID_FOR_CA was used in the first place is that is possible to decrypt the same channel from 2 separate adapters, or in case of a sid clash.

PR #1092, opened 2023-03-31
Various minor fixes, mainly to PMT names

Jalle19:
Fixes truncated PMT names on CI adapters. My theory is that `sizeof((char *)foo)` is always 8, not actually `sizeof(foo)`.


src/pmt.c
```diff
@@ -2019,7 +2020,7 @@ int process_pmt(int filter, unsigned char *b, int len, void *opaque) {
     return 0;
 }
 
-void copy_en300468_string(uint8_t *dest, int dest_len, uint8_t *src, int len) {
+void copy_en300468_string(char *dest, int dest_len, char *src, int len) {
```
catalinii, line 2023: This method should use unsigned char as arguments as chars can have values over 128 which becomes negative when assigned to charset variable.

catalinii: See line 2030.

Jalle19: Okay, I'll drop that part of the revert then

Jalle19:
I got a new baby a few days ago so it'll take some time before I get around to this

estimadarocha:
> I got a new baby a few days ago so it'll take some time before I get around to this

Congratulations... Hope you can sleep.... Eheheh

catalinii:
Big congrats @Jalle19 ...I have a 11 months old ...yeah as @estimadarocha sleep is the biggest challenge (at least for us)

src/pmt.c
Jalle19: Added a new commit which should address this

Jalle19:
Build error was unrelated and transient. @catalinii can you check the latest commit?

src/pmt.c
```diff
@@ -1331,7 +1331,8 @@ int pmt_add(int adapter, int sid, int pmt_pid) {
     pmt->opaque = NULL;
     pmt->first_active_pid = -1;
     pmt->ca_mask = pmt->disabled_ca_mask = 0;
-    pmt->name[0] = pmt->provider[0] = 0;
+    memset(pmt->name, 0, sizeof(pmt->name));
+    memset(pmt->name, 0, sizeof(pmt->provider));
```
catalinii, line 1335: pmt->provider

catalinii (approved):

catalinii:
LGTM with comment

src/pmt.c
Jalle19: Nice catch!

PR #1101, opened 2023-04-24
Allocate memory from the stack for the pmt->batch

catalinii:


Jalle19:
What does this fix? Performance?

catalinii:
Memory consumption. Now pmts are allocated for each channel and not released and pmt batch is the biggest part of that.

PR #1104, opened 2023-05-17
Don't build AXE builds with PMT=0

### LINKED ISSUE #1099: Stop building AXE builds with PMT=0
Discussed in some commit recently. The builds done by satip-axe has had PMT support enabled for some time now with non adverse side effects.

Jalle19:
Closes #1099

Note that this is for master, since the AXE binaries are only built for that branch


Jalle19:
Tested since forever here: https://github.com/Jalle19/satip-axe/blob/master/Makefile#L311

PR #1105, opened 2023-05-17
Fix bugs in "-o" parameter parsing, factor out and add tests

Jalle19:
This is against `next` but should be backported to `master` as well.

@uwe5 are you able to test this?

@catalinii I only factored out parsing of this particular parameter here, but IMO all but the most trivial parameters should get the same treatment so we can write tests and verify the parsing actually works. Especially options with multiple optional prefixes/suffixes etc. are really hard to verify just by looking at the code.

@lars18th feel free to test as well, you were the last one who touched this

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -43,6 +43,7 @@ SOURCES=\
 	stream.c \
 	adapter.c \
 	httpc.c \
+	opts.c \
 	utils.c \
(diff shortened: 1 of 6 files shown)


catalinii (approved):

Jalle19:
@uwe5 have you had a chance to test this?

PR #1110, opened 2023-05-29
Fix building with --enable-axe --disable-dvbca

Jalle19:
We must not include anything that depends on `<openssl>` since it's not available on AXE devices


Jalle19:
Correction, OpenSSL is currently available but may not be in the future. I'm trying to move the build environment away from the ancient SH4 toolkit currently used, and there are no distributions except Debian Sid that have a base system for SH4 available so everything must be compiled manually, and I'm trying to cut down on what is needed. OScam needs OpenSSL so I may get rid of that as well.

PR #1116, opened 2023-06-01
Fix incorrect BW value with some clients

lars18th:
Some SAT>IP clients send the parameter "bw" (DVB-T/C) in a incorrect range. The specifications indicate that acceptable values are: "5”, ”6”, “7”, “8", “10”, “1.712”. Therefore values like "bw=8000" or "bw=8000000" are incorrect. But we can handle them. In fact previously these requests will generate an overflow internaly in the value but the request will be served.

This patch fixes this recalculating incorrect values.


lars18th:
Hi @Jalle19 ,

Please, merge it. Finally I've discovered why my DVB-T2 tuners will show incorrect bandwith values... 🥇

Issue #1114, opened 2023-05-31
[REQUEST]: Smart "get_free_adapter()" functionality


lars18th:
Hi @catalinii ,

At time the function `adapter.c --> get_free_adapter()` is quite simple. For example in this use case it will have troubles when:
- Three tuners, each one is a different SAT>IP server. And minisatip drives the tuners using the "satipc" module. The first tuner is a capable device with sufficient CPU to stream 50Mbps. But the two others devices can stream only up to 10Mbps.
- To this central minisatip server one user requests a service 1 from MUX1.
- Then another user requests service 2 from MUX2.
- Then another user requests a service 3 from MUX2.
- What happens? In this case the adapter1 (the good one) is streaming only 1 channel, but the adapter2 (basic device) is streaming 2 channels at time, and the streaming suffers.

My suggestion is to include some smart checking to select the "free" adapter based on more information. Then in the previous example, tuner1 will streams service 1 from MUX1, tuner2 service 2 from MUX2 and tuner3 service 3 from MUX2. And no problem appears. I don't want to introduce any advanced selection, just simple restrictions easy to handle.

The idea is then:
- When one adapter is running, you store the current bitrate, the number of pids, and the number of services. All data is at time already calculated.
- In the command line the user will have a new option to control the "adapter-performance-restrictions". Like: "limit bitrarte to...", or "no more than X pids", or "up to Z services".
- And then inside the `get_free_adapter()` these values could be used to "discard" some potential tuner and then go to the next.

In fact, it's more or less to put some limitations in the adapter selection when the current tunning parameters match, but you don't want to select this adapter because other restrictions apply.

What you think?

Jalle19:
I vote no. The purpose of the function is to find a free tuner. A tuner being on a shitty device doesn't make it less free.

lars18th:
Hi @Jalle19 ,

Why not? The idea is this:
- The current behaviour will continue equal to now if the user doesn't add any "restriction".
- The command line will be something like this: `--adaptarter-limit 0:2:20:10000` that indicates: "for adapter 0, limit to 2 services, 20 pids and 10Mbps". By default the values will be `-1:-1:-1` (aka no limit).
- The implementation will be a simple _new_ function that checks the limits. And this function will be called inside the `get_free_adapter()`. Therefore nothing else will be changed.

Its *clear*, *simple* and *non-intrusive*. So, why not?

Jalle19:
[...]

PR #1117, opened 2023-06-01
Add adapter restrictions

lars18th:
A new parameter called "--adapter-restrictions" is added to enforce some adapter limitations when selecting a free tuner. The three limits are (in order of checking): number of used services, number of total pids, consumed bandwith. By default for all adapters the limits are "-1 : -1 : -1". But if you want to limit the adapter #1 to three simultaneous services only use "1:3:-1:-1".

Note: The calculation of the current bandwith is not currently implemented. Therefore it really doesn't work.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -31,7 +31,6 @@ jobs:
       run: |
        make clean
        ./configure --disable-dvbapi --enable-static --host=sh4-linux --disable-dvbaes --disable-dvbca --enable-axe
-       make PMT=0
        zip -9 -r /minisatip_axe.zip minisatip html
 
     - name: Build MIPS
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,9 +1,9 @@
 name: Binaries
 on:
   push:
-    branches: [ master ]
+    branches: [ master, next ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, next ]
 
 jobs:
   build:
@@ -16,28 +16,29 @@ jobs:
         fetch-depth: '0'
     - name: Build x64
       run: |
-       ./configure --enable-static
+       ./configure --enable-static CC="clang -static" --disable-netcv
        make DDCI=1
        zip -9 -r /minisatip_x64.zip minisatip html
 
     - name: Build ARM
       run: |
        make clean
-       ./configure --enable-static --host=arm-linux-gnueabihf
+       ./configure --disable-netcv --enable-static --host=arm-linux-gnueabihf CC="clang -target arm-linux-gnueabihf -static"
        make
        zip -9 -r /minisatip_arm.zip minisatip html
 
     - name: Build AXE
       run: |
        make clean
-       ./configure --disable-dvbapi --enable-static --host=sh4-linux --disable-dvbaes --disable-dvbca --enable-axe
+       ./configure --disable-netcv --disable-dvbapi --enable-static --host=sh4-linux --disable-dvbaes --disable-dvbca --enable-axe
+       make PMT=0
        zip -9 -r /minisatip_axe.zip minisatip html
 
     - name: Build MIPS
       run: |
        make clean
-       ./configure --enable-static --host=mipsel-tuxbox-linux-gnu
-       make
+       ./configure --disable-netcv --enable-static --host=mipsel-linux-gnu CC="zig cc -target mipsel-linux-gnueabihf.2.34 -static"
+       make EXTRA_CFLAGS="-I/sysroot/mipsel/include -L/sysroot/mipsel/lib" EMBEDDED=1
        zip -9 -r /minisatip_mips.zip minisatip html
 
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,3 +36,10 @@ jobs:
        make clean
        ./configure --enable-linuxdvb --enable-dvbca --enable-dvbcsa --enable-netcv --enable-satipc --enable-dvbapi --enable-axe
        make
+
+    - name: Build with --enable-axe, without libssl-dev
+      run: |
+       apt-get -y remove --purge libssl-dev
+       make clean
+       ./configure --enable-axe --enable-dvbapi --enable-dvbcsa --disable-dvbca --disable-dvbaes --disable-netcv
+       make
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,9 @@
 name: Build
 on:
   push:
-    branches: [ master ]
+    branches: [ master, next ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, next ]
 
 jobs:
   build:
@@ -20,7 +20,7 @@ jobs:
     - name: Install build essentials
       run: |
         apt-get -y update
-        apt-get -y install build-essential
+        DEBIAN_FRONTEND=noninteractive apt-get -y install build-essential git clang
 
     - name: Build all disabled
       run: |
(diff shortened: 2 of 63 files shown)


lars18th:
Hi @catalinii and @Jalle19 ,

This PR implements the request #1114. Please, note that the code is simple. The most large function is the parsing of the parameters. The data of the used pids and services it's already calculated. The only missing is the bandwith. All the code is here, but I don't know how to calculate it _for each individual adapter_. So I request your help to complete this.

All my tests limiting the number of services (2 or 3 for example) and the pids (10, 15, 20, etc.) work without troubles.
I feel this will useful for some users that have multiple tuners with some hardware limits. You can define the restrictions at adapter level, so it's very powerful.

Regards.

Jalle19:
Clearly we have different opinions on what "simple" means 😄

catalinii:
@lars18th there is another point that is not obvious:
the bandwidth at the adapter level is different than the bandwidth that is being sent over the network:
For example if 1 channel is requested by 2 or more clients independently:
- all pids will be set once on the adapter
- the stream.c code will send the same pids across the network to 2 independent clients

So even if the BW at adapter level is <15Mbits, you will still send > 15Mbits over the network.

lars18th:
> @lars18th there is another point that is not obvious:

Hi @catalinii ,

Thank you for the response. Regarding your question, what I want to implement regarding the limit of the bandwith is the "adapter bandwith" and not the network bandwith. The problem is that not all adapters/drivers/devices can handle the full bandwith of a transponder. And then if you put a limit, then you have the option to not serve more requests over this limit. For sure the limit is reliable for CBR and not true for VBR. However, with a single 10Mbps limit you will not get add more services if the current bandwith is going from 10-15Mbps for example. That's the simple idea. And the limit have more sense if you are using the `satipc` module because in this case the "network bandwith" will be the same as the "adapter bandwith". So you can help me to calculate this value?

Futhermore, I'll try to simplify the code a bit more.
Regards.

lars18th:
Hi @Jalle19 ,

> Clearly we have different opinions on what "simple" means 😄

If you remove the parsing of the command line parameter the code is very _simple_. Just a few lines of code.

lars18th:
Hi @catalinii ,

I've implemented the bandwith adapter calculation. Therfore the check is complete.

Please, check it and merge.
Regards.

lars18th:
Hi @Jalle19 and @catalinii ,

If you aren't testing this PR because the erros in CI, then the problem is not with my code. It's because this commit https://github.com/catalinii/minisatip/commit/8d33e939b820fef824c84a92dd6affcd455bbf55 that has broken the binaries build.

So please fix this, and check my PR. I'm waiting for merging it.
Regards.

catalinii:
@lars18th this PR takes the current get_adapter_code to a place where is not really too complex. The motivation for that as I said is to support devices that are not supported by their vendor. As I said is not worth merging it.

lars18th:
Hi @catalinii ,

> @lars18th this PR takes the current get_adapter_code to a place where is really too complex and hard to reason about. The motivation for that as I said is to support devices that are not supported by their vendor. As I said is not worth merging it.

Sorry, but I don't agree with that.

Let me to comment these key points about devices:
- Some SAT>IP servers with _certification_ (https://www.satip.info/resources/products/) can't handle more than N programs at the same time.
- A lot of SAT>IP _hardware_ devices have closed source firmwares with last dates of more than five years. And these devices continue in production. So we can't consider them as not supported by the vendor.

Let me to comment other points regarding the PR too:
- The code inside this patch is very simple and clean. Futhermore it provides functionality to "debug" the current state of an adapter related to number of programs, bandwith and number of pids in use. Did you notice?
- The target of this code when I've created it isn't with the idea of solving some troubles with low cost Enigma boxes. The idea is to limit the number of channels processed by one tuner. Based on three areas: bandwith, pids and programs. A know a lot of different devices that has one of these limitations. I've asked/proposed this several years ago if you remember it.

So please, don't be wrong. This patch has sense for several users. And because it doesn't change anything by default I feel it's preferable to merge it.

Anyway, please comment if you want about point per point of the discussed points. Perhaps I'm really missing something.
Regards.

Jalle19:
> Some SAT>IP servers with certification (https://www.satip.info/resources/products/) can't handle more than N programs at the same time.

I feel it would be better to document which these devices are so people don't buy them, instead of working around their shortcomings.

lars18th:
> > Some SAT>IP servers with certification (https://www.satip.info/resources/products/) can't handle more than N programs at the same time.
> 
> I feel it would be better to document which these devices are so people don't buy them, instead of working around their shortcomings.

Sorry? You can recommend whatever you want. However, manufacturers will continue to provide what they want. That's the current situation. And minisatip **can** handle any type of devices. And this PR is proof that this is true.

Sorry @Jalle19 and @catalinii , but I'm very confused about your reluctance to merge a simple/clear/useful/non-intrusive patch that provides very interesting functionality. Just because you don't seem to want to boost "underpowered and poorly supported hardware". My intention is to enhance this software, and in my environment the _physical_ tuner is not very relevant.... I'm using from Digital Devices to low cost ALi based STB (very, very, very bad and closed source). But with a central minisatip I can manage everything together.

And in case you are still thinking about it, I don't recommend buying certain low-cost Enigma2-based products with outdated, poorly implemented, unsupported, absurdly limited and closed-source drivers. But what's wrong with adding this functionality?

Regards.

Jalle19:
As a compromise, could we consider just having a limiter for the number of services per adapter?

lars18th:
Hi @catalinii and @Jalle19 ,

I've resolved the conflicts with this patch, but now the tests fail because the out-of-sync with the current master branch. I feel I need to return to create a PR to provide this change. However, BEFORE THAT I need to know your current point of view. Please, take note of the effort to contribute to the project and be as open-minded as you can:

- I've created this enhancement to improve the user experience and make the tool more flexible. Please, belive me that the root cause is not to add support for buggy devices. At the beginnig my objective was to improve this debug line that prints the current state of the used tuners:
https://github.com/catalinii/minisatip/blob/80578a2031b14a63e3a2204be064134fda7fa1a8/src/adapter.c#L562-L567
- The reason is because I need to know the number of services and used pids when a new tuner will be selected.
- Therefore I've implementing this in the code without much effort. But after I see that this could be enhanced with more data and can be used to tweak the selection of a new tuner. So finally, I've added three new data to the print: services, bandwith and pids. And I've added also code to check these values when a new tuner will be select. And at the end I've added the command line parameters to use different values for each adapter.

Said that, let me to explain the use cases of this new functionality:
1. In the debug now you can see in the dump of the current used adapters new relevant information: number of services, total input bandwith and the number of pids enabled. This info is provided without any overhead. And it's very useful to detect some interferencies that could be created by other changes or unsupported hardware.
2. When you have more than one remote SAT>IP servers providing the same frequencies now you can limit the total used bandwith. Imagine that you have a central minisatip, connected to TWO different SAT>IP servers in a remote location. Each one can connect to the same transponders, and the network to reach each of them has around 8Mbps. Then you can stream 1 service of around ~6Mbps without problems. OK, but what happens if you request 2 services of the same transponder? In this case, only one remote SAT>IP will be used. But then the output bandwith could exceed the available bandwith (12 versus 8). And therefore you'll in troubles. One solution is to limit each adapter with this PR to 8Mbps. Then you could request anything on them until the bandwith will be less. For example, one HD video service and 4 radios perhaps could be handled by one SAT>IP server, or two SD services, etc. But if you request two HD services _of the same transponder_ then the two adapter will be used. Nice, right?
3. Another different use case: you have multiple SAT>IP servers that handle CA in hardware. But you can decrypt only 2 services with one CA. In this case, you want to request 1 or 2 services only for each server. And this is true also in case of multiple services in the same transponder. Therefore with this patch you can now set the maximun number of services that you want to request to an individual adapter. Futhermore with a different value for each one. Nice too, right?
4. Finally the limit of pids could be another different use case. For this the only limit is the hardware pid filtering of the server. In this case we need to remember that a lot of hardware SAT>IP servers have a limit of the number of pids. It's common a large number like 32, 64 or more. But the limit exist in several devices. This is not the case with enigma2 boxes. As we can run minisatip on them, and overcome any hardware limit with a software pid filtering if the tuner driver supports the full transport stream. Anyway, the functionality is here and we can limit the pids without any effort.

After this explanation about the sense of this enhancement, please review the code and check the real simplicity of it:
- Overall the code are simple lines or blocks of few lines. Except the function `set_restrictions_adapters()`. This is a new 50 lines of a parsing function. And it only handles the new parameter `--adapter-restrictions`. Nothing more, nothing less.
- This patch follows the KISS principle because it only adds a new functionality in a simple way.
- It also follows the main principles pointed by @catalinii bacause: a) It doesn't changes the current functionality. b) It's a non-intrussive new paramter that could be used or not. c) It doesn't breaks other modules. d) It doesn't depends on any other functionality.

Therefore, I feel this could be added without any problem. But I need your confirmation. My idea is that if you accept this then I'll remake the PR for the current master branch to be ready to merge it.

You agree?

Jalle19:
From the use cases you listed, numbers 2 and 3 are IMO reasonable scenarios. Perhaps we could add just support for "max services" and skip "max pids" and "max bandwidth" to keep things simple?

Jalle19:
Closing this for now, we can re-open it when you have time to work on it again :pray:

lars18th:
> Closing this for now, we can re-open it when you have time to work on it again 🙏

Hi @Jalle19 ,

No problem. I'm using it for a lot of time, but I've more "smart" options for the tuner selection. So I'll try to implement all-in-one in a new P.R. in the future when I'll have sufficient time.

Regards.

Issue #1078, opened 2023-03-13
UUID not generated if SSDP discovery is disabled


Jalle19:
With discovery disabled, it should still be possible to use the SAT>IP server by providing the URL to `/desc.xml` manually. This currently only kind of works, because the XML is out of spec:

```
<UDN>uuid:</UDN>
```

This at least causes some older tvheadend versions to crash

lars18th:
Interesting, so we need to generate the UUID outside of the SSD function.

Jalle19:
Yep, looks like it. I'm thinking it could also make sense to store the UUID in a file so it could be decoupled from the MAC address completely. But first step would be to separate it from SSDP discovery.

lars18th:
> Yep, looks like it. I'm thinking it could also make sense to store the UUID in a file so it could be decoupled from the MAC address completely. But first step would be to separate it from SSDP discovery.

I don't think so. Storing the UUID in a file will generate troubles. It's preferable to assign it with the command line if the user *needs* it. Remember it: minisatip is _file config free_ !!

Jalle19:
I'm thinking it would auto-generate a random UUID and store it in a file if the file does not exist.

lars18th:
Hi @Jalle19 ,

I implemented the fix in #1119 . 😉

lars18th:
Hi @Jalle19 ,

Please remember to close this issue when PR #1140 will be merged. 👍 

Regards.

PR #1119, opened 2023-06-12
Fix UUID generation with SSDP disabled

lars18th:
This fixes the empty UUID when the SSDP is disabled with the command line parameter `-G`, as described in #1078.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -979,29 +979,7 @@ void set_options(int argc, char *argv[]) {
                 app_name);
             exit(0);
 #else
-
-            char buf[100];
-            memset(buf, 0, sizeof(buf));
-            safe_strncpy(buf, optarg);
-            if (buf[0] == '~') {
-                opts.pids_all_no_dec = 1;
-                memmove(&buf[0], &buf[1], sizeof(buf) - 1);
-            }
-            char *sep2 = strchr(buf, ',');
-            if (sep2 != NULL) {
-                *sep2 = 0;
-                opts.dvbapi_offset = map_int(sep2 + 1, NULL);
-                _strncpy(buf, optarg, sizeof(optarg) - 1 - strlen(sep2));
-            }
-            char *sep1 = strchr(buf, ':');
-            if (sep1 != NULL) {
-                *sep1 = 0;
-                safe_strncpy(opts.dvbapi_host, buf);
-                opts.dvbapi_port = map_int(sep1 + 1, NULL);
-            } else {
-                safe_strncpy(opts.dvbapi_host, buf);
-                opts.dvbapi_port = 9000;
-            }
+            parse_dvbapi_opt(optarg, &opts);
 #endif
             break;
         }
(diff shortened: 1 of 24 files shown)


lars18th:
Hi @Jalle19 ,

Check it and merge.
Regards.

src/minisatip.c
```diff
@@ -1939,7 +1948,8 @@ int main(int argc, char *argv[]) {
                           (socket_action)ssdp_discovery);
         if (si < 0 || si1 < 0)
             FAIL("sockets_add failed for ssdp");
-    }
+    } else
+        generate_uuid();
```
Jalle19, line 1952: Is this needed? Would it be possible to call `generate_uuid()` just in one place?

lars18th: Yes, it's necessary because the **global** var `uuidi` that controls the UUID generation. So we need to be sure that _in all cases_ the UUID is created. So for reason and in order to not generate side effects errors if the code changes in the future, then it's preferable to call to the new `generate_uuid()` function every time a SSDP call is done and overcomed by the `opts.disable_ssdp` option. Therefore it's more safe to add this line here.

Please check it and merge! 😉

Jalle19: I still think it would be preferable to all `generate_uuid()` just once sometime during startup. Then we could even get rid of `uuidi` completely.

lars18th: Hi @Jalle19 ,

That's my original idea. However, I'm not the one that has introduced the `uuidi` var. This var has _specifically_ the objective of caching the creation of the UUID. I feel this was introduced (by @catalinii) to overcome potential errors of different edge cases or to permit to recalculate it. Anyway the fix is clean, simple and it works, so I suggest to merge it.

Regards.

lars18th: Hi @Jalle19 ,

I undertand you suggestion. And I agree with the idea of some refactor. However, now I don't have much time. So because the code is clean and it fixes the bug, why not merge it now and in the future someone will refactor it?

Jalle19: I'll see if I can find the time to do the necessary changes

lars18th: Hi @catalinii ,

Can you comment please?

catalinii: A lot of things have been done historically for reasons probably they are no longer valid.
Cleaning the code is the best way forward IMHO, so I think that stripping the code out from there and putting it in the initialization part would make the code more clean and it would be called just once.

lars18th: Hi @catalinii ,

Thank you for the response. I agree with you and @Jalle19 . I prefer it the code will be cleaned. However, I don't have time now to do it. So I hope @Jalle19 could enhance the PR. In fact it's sufficient to check _all_ calls to `generate_uuid()`. The problem in my case it's the time necessary to do an extensible check of all edge cases. And for this reason I've suggested to merge the PR and _after_ refactor the code. Perhaps the most simple way is to encapsulate all initialization tasks inside a new function and execute the call to `generate_uuid()` inside it.

Regards.

Jalle19:
Replaced by https://github.com/catalinii/minisatip/pull/1140

Issue #1107, opened 2023-05-22
minisatip and James Donkey HD DUO


lars18th:
Hi @catalinii and others,

I've now one unit of the cheap James Donkey HD DUO. And after a lot of testing over several weeks I can finally run minisatip on this device, albeit with fairly limited capabilities. So first of all I want to share the facts I have discovered:
- The unit can run minisatip if you install the last version of the OpenATV firmware version **6.4**. One relevant aspect is to use the _driver version_: **20210607** (kernel 4.4.35).
- To run minisatip you need to use the option `-1 1` because the stream can only be readed from the `/dev/dvb/adapter0/demuxX` and not the common `/dev/dvb/adapter0/dvrX`.
- The "streaming" mode requires to be enabled using `echo 1 > /proc/stb/parameters/ts_tap` (already set by the script `/etc/init.d/dinobot-pvr.sh`).
- The firmware has some bug regarding the background splash image. Therefore you need to kill the `showiframe` process that is consuming a lot of CPU resources.
- Current minisatip ARM binaries can't be run because a mismatch with the GLIBC version. You need to recompile it with an older toolchain with `./configure --enable-static --host=arm-linux-gnueabihf --disable-dvbca --disable-dvbcsa` and `make EXTRA_CFLAGS="-DNEEDS_SENDMMSG_SHIM"`.
- The pid filtering implemented inside the dvb driver is very buggy. With the simple `dmesg` command you can see a lot of errors. And now I discovered that the driver only supports up to 10 "service" pids plus another 10 "psi" pids. So you have to request the opening of the pids in this order: first you can request up to 10 pids from the range 32-8190, and then continue with another 10 pids from the range 0-31. This is very crazy. But if you request the opening of the pids not using this technique, then you can open only a small number of pids. If you request for example: `100,0,101,1,102,16` then you will recive only pids 100 and 0, and nothing else! So the golden strategy is to open **(0-10)service_pids, (0-10)psi_pids**. 
- But anything else seems to work very well. I can use the two tuners independently, and stream multiple pids with a bandwith of 30mbps from each tuner without troubles.

Therefore, I request if you can help to improve the support for this device:
1. The first "simple" question is to provide a compatible binary. I suggest to add to the releases another ARM target compiled with an older toolchain, and with `EXTRA_CFLAGS="-DNEEDS_SENDMMSG_SHIM"`. I feel several low cost devices with ARM cores will agree. You accept this @catalinii ?
[...]

PR #1120, opened 2023-06-20
Use clang for arm/x86 with static binary and zig cc for mips #1107

catalinii:
This also fixes the building with the new libnetceiver

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,9 +1,9 @@
 name: Binaries
 on:
   push:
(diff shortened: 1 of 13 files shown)


Jalle19:
libdvbcsa-dev is missing from the builder image

Jalle19:
@catalinii would it make sense to move the Dockerfile used to build the image to this repository?

Jalle19:
Seems like you we have to add DEBIAN_FRONTEND=noninteractive to the apt commands

PR #1136, opened 2024-01-27
fix typo and cosmetic readme changes

docgalaxyblock:
Hi, I've found a little typo (that triggered me) and made the redme a bit prettier on github.


catalinii:
The changes are fine, but they need to be made in readme_* files as the README.md is autogenerated by tools/gen_readme 
The other part of the readme is in minisatip.c

docgalaxyblock:
Updated.
Since for me it looks like your gen_readme is not used by an action and the readmes list version 1.2 ...
How about adding an action that triggers an every release, runs that script and then commits the change to the READDME.md ?

docgalaxyblock:
Hey @catalinii could this be merged now?

PR #1140, opened 2024-03-22
Decouple UUID generation from SSDP discovery

### LINKED ISSUE #1078: UUID not generated if SSDP discovery is disabled

Jalle19:
Fixes #1078, replaces #1119

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -1437,8 +1437,6 @@ int read_rtsp(sockets *s) {
 
 #define JSON_STATE_MAXLEN (256 * 1024)
 
-USockAddr ssdp_sa;
-
 int read_http(sockets *s) {
     char *arg[50];
     char buf[2000]; // the XML should not be larger than 1400 as it will create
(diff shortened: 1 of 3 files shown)


catalinii (approved):

Jalle19:
Needs runtime testing, seems to work but 🤷

src/minisatip.c
```diff
@@ -1669,23 +1670,18 @@ int close_http(sockets *s) {
...
-    if (uuidi == 0) {
-        uuidi = 1;
-        get_mac_address(mac);
-        sprintf(uuid, "%s-%s", uuid1, mac);
-        // use IPv4 only as disc_host is multicast IPv4
-        fill_sockaddr(&ssdp_sa, opts.disc_host, 1900, 1);
-    }
+
+    // use IPv4 only as disc_host is multicast IPv4
+    fill_sockaddr(&ssdp_sa, opts.disc_host, 1900, 1);
```
Jalle19, line 1682: I have a feeling that this is no longer correct (this should not be called all the time)

catalinii: The multicast address needs to be set somewhere in the ssdp_sa. Maybe is not the right place but it should be done, or am I missing something?

Jalle19: The fact that the variable `ssdp_sa` was global made me think there were some other uses for it, but it's apparently only used in this function so it should be fine to call `fill_sockaddr` many times.

Jalle19:
Works as expected, UUID is generated correctly with `-G` applied.

PR #1141, opened 2024-04-17
AXE build changes, attempt to store build artifacts

Jalle19:
Was looking for an easy way to test #1140 on my AXE devices, this should help with that. The idea is to be able to download the built binaries from any GitHub workflow run.


Jalle19:
Works the way it should, will merge this and rebase #1140 so I can test that one locally

Issue #586, opened 2019-05-14
Sundtek dvb-s2 powersave


riegaz:
I have a sundtek dvb-s2 USB connected to a linux machine(A) with minisatip installt. For watching live TV I use another box(B) to access the dvb-s2 via minisatip. After shutting down B where I use tvheadend. I would expect the dvb-s2 to go to sleep. However, the A driver detects that minisatip is still accessing the dvb-s2 and thus doesn't send the dvb-s2 to powersave.

Any ideas on this one?

Ntemis:
This is not a minisatip issue, as there is no mechanism to check if the usage of the adapters are nil then go to sleep and when rtsp is requested to wake up :)

riegaz:
I see your point. However, at the Sundtek forum they tell me otherwise. Honestly, I like the approach that minisatip should release /dev/dvb/adapter0/frontend0 if there is no rtsp.

What do you think?

The post is in german: https://support.sundtek.com/index.php?topic=2399.0

Ntemis:
And then there was /dev/dvb/adapter1 /dev/dvb/adapter2 etc

riegaz:
Exactly, my question is why minisatip shouldn't release them when not used?

Ntemis:
You have to kill the app

riegaz:
I know that killing minisatip would help it but this isn't the goal. Minisatip should keep running. However, release the tuner when not used.

Ntemis:
Then go back to my initial answer.

catalinii:
Minisatip should release any adapter after 30s after the last client
disconnected. If you do not have any client connected and see minisatip
holding the adapter open, please upload logs.

You can use lsof | grep adapter0 to see who keeps the adapter open

catalinii:
If this is still an issue please feel free to re-open

PR #1144, opened 2024-05-08
Release adapter if there are no listening clients.

andrewjw:
Hi,

I have Sundtek DVB-S/S2 USB device, which works reasonably well. It works most of the time, but is a bit unreliable and sometimes fails to tune correctly. I've been trying to debug where the problem is - it could be MythTV, minisatip, the drivers, device or my cables... Anyway, as part of my research I came across [this forum post](https://minisatip.org/forum/viewtopic.php?f=5&t=840&p=4368&hilit=sundtek&sid=9cb93658b17be3a90bdc2259c22973ab#p4368) which contains a patch to get minisatip to correctly release a tuner when no-one is listening, and I can't see it having been submitted. I think this might also relate to #586.

I'm aware that in the forum post there is a comment that this might be better done with a timer, but at least initially I wanted to submit the patch as-is, for completion and for attribution to the original poster, Rainer.

This patch seems to work well for me, and the devices do go offline when not in use, which previously they didn't. It also seems to improve the instability I see a bit, but doesn't fix it completely though. I suspect that's a Sundtek driver issue though.

Thanks,
Andrew

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -309,6 +309,7 @@ int init_hw(int i) {
         ad->delsys(i, ad->fe, ad->sys);
     ad->master_sid = -1;
     ad->sid_cnt = 0;
+    ad->nosid_cnt = 0;
     ad->pid_err = ad->dec_err = 0;
     ad->new_gs = 0;
     ad->force_close = 0;
--- a/src/adapter.h
+++ b/src/adapter.h
@@ -31,6 +31,7 @@ typedef struct ca_device ca_device_t;
 
 #define MAX_DELSYS 10
 #define MAX_PMT_FOR_ADAPTER 255
+#define MAX_NOSID 3000
 
 typedef struct struct_adapter adapter;
 struct struct_adapter {
@@ -54,6 +55,7 @@ struct struct_adapter {
     int ca_mask;
     int master_sid; // first SID, the one that controls the tuning
     int sid_cnt;    // number of streams
+    int nosid_cnt;
     int sock, fe_sock;
     int do_tune;
     int force_close;
--- a/src/stream.c
+++ b/src/stream.c
@@ -1090,6 +1090,19 @@ int read_dmx(sockets *s) {
         return 0;
     }
 
+    LOG("adapter %d sid_cnt %d nosid_cnt %d", ad->id, ad->sid_cnt, ad->nosid_cnt);
+    if (ad->sid_cnt < 1) {
+        ad->nosid_cnt++;
+        if (ad->nosid_cnt > MAX_NOSID){
+            ad->nosid_cnt = 0;
+            request_adapter_close(ad);
+        }
+        s->rlen = 0;
+        return 0;
+    }
+    else
+        ad->nosid_cnt = 0;
+
     threshold = ad->threshold;
 
     if (rtime - ad->rtime > threshold)


catalinii:
I think there is a feature to close the adapter after 30s by default, if that does not work, please upload the logs of

./minisatip -f -l http

lars18th:
Hi @catalinii and @andrewjw ,

The current behaviour is that minisatip maintains any tuner "open" until the close delay (30'') is achieved. This resolves a lot of troubles in different scenarios and don't generate side effects. This works because the tuner is left open but without reading any pid. And if the tuner is to be used, then the correct commands are sent to it. So in principle everything works as expected.

However, I think some obscure bug is inside minisatip at the moment. The problem, perhaps the root of your problems, is that in some cases the tuner is in "open" state, there are no users connected, and a new connection requests a new tuning but minisatip doesn't select the "open" tuner. This problem has more effect on servers with a single tuner, because the new request cannot be completed. But it has less effect with multiple tuners. But in any case, in this situation I can see this problem as well. I'm not sure but maybe the problem is related to the internal stream management. And I add that usually the problem happens more frequently when the tuner can't find good signal. In that case it stays "semi-open" (in open state but not selected) until the 30 seconds.

That said, I reject to merge this patch because it only implements a hack: every time a tuner stops having clients, then a counter is incremented. Until a maximum is reached and then it's forced to close. This doesn't solve the problem. And the counter in this patch has a very high value (3000). Therefore, I don't recommend it (but you can compile your own version... of course you can! You're free to do so, and if it works for you, then go for it. But in that case I recommend using a value no larger than 10 for `MAX_NOSID`).

Finally, I do think there is a problem and it needs to be solved. But in a different way.
Regards.

andrewjw:
Hi,

Sorry for the delay in getting the requested logs, I plan to get those tonight.

I agree that the patch as it stands is a bit of a hack. I found it on the minisatip forums when trying to debug a problem I'm having, and I couldn't find a PR on GitHub about it. I wanted to share it as-is to start a discussion about it. The forum thread was about a Sundtek tuner (which I also use), but I don't know if it also affects other brands.

I am running the patch and it does what I wanted, which is reliably switch off the tuners when my MythTV box goes to sleep.

In the forum thread there is a comment about using a timer rather than a counter, but it's essentially the same strategy. It sounds like you suspect there is a deeper bug that needs fixing? I'm happy to help change the patch to use a timer, but fixing deeper bugs might be beyond my abilities.

Cheers,
Andrew

lars18th:
Hi @andrewjw ,

I suggest to wait until you provide the log to start the discussion. Because, if you change the delay of 30'' to 0 then it will be the same (more or less) that enforcing the close of the tuner by a timer. And I repeat: without this delayed close a lot of other troubles appear. So it's necessary.

Regards.

andrewjw:
[minisatip.log](https://github.com/catalinii/minisatip/files/15326092/minisatip.log)

Here are the logs as promised. I used VLC to view a channel for a few seconds, then disconnected. After three minutes the frontend was still saying 92% signal strength, although the frequency showed 0.

Cheers,
Andrew

lars18th:
I think this is not normal:
```
[15/05 20:25:53.500 AD0]: leak detected 1 0!!! 
```

@catalinii , any idea why this message appears?

catalinii:
Can u upload a log with -v dmx  and do the same test ?

Jalle19:
Closing this due to inactivity. It's probably better to open an issue about this instead of continuing with this PR (since it's unlikely to get merged).

Issue #1091, opened 2023-03-25
Services pinning to specific CAM or do-not-descramble option


Delitants:
I think this feature was requested earlier, but I can't find a related ticket.
This is essential, please make possible pinning services to a specific CAM and do-not-descramble option per service (similar to existing flag -E but I want to set manually what service to pass though).

Situation is next. Tuner TBS 6590 with dual CI slot, TP has 3 services, one SHOULD be descrambled by CAM-1, second with CAM-2 and not otherwise, and third must be passed AS IS without even trying to process it, because it will be descrambled by softcam in different software.

Note: flag -E does not work in this scenario, because it still tries to send channel to an already busy CAM, it passes through only if cam refuses to descramble, but this is not the case.

Jalle19:
Pinning services to a specific CAM is supported, but "ignoring" a service is not.

Delitants:
> Pinning services to a specific CAM is supported, but "ignoring" a service is not.

How do you pin them without physically using secondary tuner of the card?

Jalle19:
I'm not sure how the TBS 6590 works, it could be that the CAM assignment cannot be controlled if the card doesn't work as "standalone" CI cards, like the Digital Devices models.

catalinii:
CAM assignment works at the moment just for standalone CI cards (tbs 6900)

Yuri666:
> > Pinning services to a specific CAM is supported, but "ignoring" a service is not.
> 
> How do you pin them without physically using secondary tuner of the card?

No way. In the tbs6590 CI are physically connected to the tuner and accordingly if you wish to use second CI you must use a second tuner.

Delitants:
> > > Pinning services to a specific CAM is supported, but "ignoring" a service is not.
> > 
> > 
> > How do you pin them without physically using secondary tuner of the card?
> 
> No way. In the tbs6590 CI are physically connected to the tuner and accordingly if you wish to use second CI you must use a second tuner.

Okay, I'm more interested in skipping services for descrambling in external software, second part of this ticket.

Yuri666:
Theoretically it can be done - just do not send CA_PMT when switching to this channel.
I don't know how difficult it is to implement. It's easier to use a dvb card without CI for these purposes.

Jalle19:
I think "service skipping" could be reasonably implemented for standalone cards (using the same ddci.conf file), but for "bonded" CI slots it's a bit clumsy like Yuri explained.

Yuri666:
[...]

PR #1148, opened 2024-06-24
Add support for service blacklisting for CAMs

Jalle19:
Brought up in https://github.com/catalinii/minisatip/issues/1091

Basically this allows the user to prevent a CAM from being used for a particular service, e.g. like this:

```
# ddci.conf
7000: # not assigned anywhere
```


Jalle19:
Had this code lying around, figured let's try to upstream it

catalinii (approved):

PR #1149, opened 2024-06-24
Fix intermediate frequency and tone for C-band LNBs

Jalle19:
Factored out the logic to separate functions and added tests to verify the functionality.

@j3dd4h can you test this branch with `-L *:5150-0-0`?


Jalle19:
User confirmed this works. @catalinii would be nice if you could review!

catalinii (approved):

PR #1150, opened 2024-06-24
Update various GitHub workflow versions

Jalle19:
GitHub Actions shows warnings for these


Jalle19:
Let's hope the master branch goes through now

PR #1160, opened 2024-07-07
Fix tuning with Unicable LNBs

### LINKED ISSUE #1158: v1.3.10 - 14 no signal on GigaBlue UHD X3 4K
hi,

i1 have since v1.3.10 no signal on GigaBlue UHD X3 4K.
v1.3.9 works fine.

Jalle19:
Fixes #1158, regression from 083c6475d9f110c7031507a6ff75b591ce0b809c

@misushot @MartB can you try this?

### DIFF
--- a/src/dvb.c
+++ b/src/dvb.c
@@ -1030,6 +1030,9 @@ int setup_switch(adapter *ad) {
 
     frontend_fd = master->fe;
 
+    freq = get_lnb_int_freq(tp, &tp->diseqc_param);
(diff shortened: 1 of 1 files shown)


Jalle19:
I have Unicable myself but I use axe devices which use a separate `setup_switch()` implementation that isn't affected by this, that's why I never noticed any issues myself. Universal LNBs should be unaffected, could be the reason why more people haven't noticed anything wrong.

catalinii (approved):

PR #1162, opened 2024-07-16
Mark current directory as "safe" for git so revision number is parsed correctly

Jalle19:
Currently, official builds done by GitHub get "1.3.~" as version


Jalle19:
Totally not sure if this change will actually do the trick

Jalle19:
Probably won't, we tag the release after the binaries are built...

PR #1167, opened 2024-08-05
Generate a random UUID on startup, store and read from bootid

### LINKED ISSUE #1166: UUID / UDN generation non-deterministic
Hi,

I was having a weird problem with TVHeadend losing its SAT-IP tuner configuration from time to time.

This was caused by Minisatip generating a different UUID when it is starts during boot. Probably because the network is not yet fully initialized.
`<UDN>uuid:11223344-9999-0000-b7ae-f2b605000000</UDN>`

After restarting Minisatip the UUID has changed to the MAC address of the network interface:
`<UDN>uuid:11223344-9999-0000-b7ae-001dec187062</UDN>`

Overwriting the MAC address using -m is a workaround.

Hardware: DUO4KSE running a stripped OpenATV (no enigma)

Jalle19:
Fixes #1166


catalinii:
Will this generate a new UUID every time?

I think the point is to have it be persisted across runs.

Jalle19:
@catalinii no, it will be persisted in the bootid file

catalinii (approved):

Jalle19:
@jellekaufmann can you try this before I merge it?

Jalle19:
Tried this on one of my installations, works as expected:
> UUID: 4fdd49d8-eb86-6c30-b211-4339746c3398 -- BOOTID: 129 -- DEVICE-ID: 414

PR #1170, opened 2024-08-07
Properly support different DVB charsets

Jalle19:
Finally a fix for those pesky service names.

Before:

After:

@Yuri666 can you try this?


catalinii:
Nice, was thinking of doing the same a while back

Jalle19:
Yeah, turns out it wasn't as difficult as I thought.

PR #1171, opened 2024-08-12
Improve RTSP return code parsing

Jalle19:
@retrofan1979 can you try this branch?

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -236,7 +236,7 @@ int satipc_reply(sockets *s) {
     satipc *sip;
     char *arg[100], *sep;
     int is_transport = 0;
-    int rc = 999;
+    int rc = 0;
     __attribute__((unused)) int rv;
     get_ad_and_sipr(s->sid, 1);
 
@@ -246,7 +246,7 @@ int satipc_reply(sockets *s) {
 
     sep = strchr((char *)s->buf, ' ');
     if (sep)
-        rc = map_intd(sep + 1, NULL, 998);
+        rc = map_intd(sep + 1, NULL, 0);
 
     LOGM("MSG process << server :\n%s", s->buf);
     LOG("satipc_reply (adapter %d, state %d): receiving from handle %d, sock "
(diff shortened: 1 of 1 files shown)


retrofan1979:
Hi, thanks, it looks far better, but still some errors in the logfile?

Channelscan does work now with TransEdit, DVBViewer and Tvheadend. Bravo!

However, I noticed --satip-xml http://192.168.178.1:49000/satipdesc.xml gives 0 devices back instead of 4.
I now manually set the tuners with -s dvbc:192.168.178.1 -s dvbc:192.168.178.1 -s dvbc:192.168.178.1 -s dvbc:192.168.178.1

Also when changing channels in Kodi with TVH client, it looks like the used tuner doesn't respond and after a time-out of 8 seconds, it switches to the next free tuner and then it streams again.

Overal, I'm very happy with this fix, but maybe it needs some more adjustments :-)

I also saw a quirck for Aurora client, but this one is very old and not working anymore and I don't know if it still needs to be in the minisatip source code (missing mtype fix).

[minisatip.zip](https://github.com/user-attachments/files/16581721/minisatip.zip)

retrofan1979:
Hmmm, when streaming a channel after 2-3 minutes the streams stops and then it continues again. See logfile attached.

[minisatip.zip](https://github.com/user-attachments/files/16587906/minisatip.zip)

retrofan1979:
Hmmm, the stream sometimes stops after 1 minute and restarts again when using 4 x -s dvbc:192.168.178.1

Jalle19:
> However, I noticed --satip-xml http://192.168.178.1:49000/satipdesc.xml gives 0 devices back instead of 4. Maybe a bug? I now manually set the tuners with -s dvbc:192.168.178.1 -s dvbc:192.168.178.1 -s dvbc:192.168.178.1 -s dvbc:192.168.178.1
I think this is a known quirk needed for Fritzbox.

The log looks fine to me in general

PR #1172, opened 2024-08-14
Download the latest Coverity Scan client before performing the scan

Jalle19:
Let's see if this works

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -3,8 +3,6 @@ name: Coverity Scan
 on:
   push:
     branches: [ master ]
-  pull_request:
-    branches: [ master ]
 
 jobs:
   build:
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -9,25 +9,10 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: catalinii/minisatip-build-image:latest
-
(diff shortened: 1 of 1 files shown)


Jalle19:
@catalinii can you double-check that there's a secret named `COVERITY_SCAN_TOKEN`?

catalinii:
Yes COVERITY_SCAN_TOKEN exists

Jalle19:
@catalinii is it the same value as in your minisatip-build-image repository? I can't find a reason why the downloading of the client doesn't work

Jalle19:
@catalinii can you send me the token via e-mail or WhatsApp or soemthing so I can try this locally?

Jalle19:
Okay so the command works fine, but the token is not being read by GitHub for some reason

9000h:
not sure but could this help
```
curl \
            --form token=$TOKEN \
```

Jalle19:
It simply doesn't work. @catalinii are you really sure the secret is defined in this project?

Jalle19:
It could be that it won't work for pull requests, otherwise anyone could make a PR that logs the token. We should try changing the action to only run against the master branch (makes more sense anyway) and merge this, then deal with the eventual fallout.

Jalle19:
It passed! https://github.com/catalinii/minisatip/actions/runs/10573195509/job/29292218299

PR #1175, opened 2024-08-27
Run tests in default ubuntu-latest container

Jalle19:


Jalle19:
Only things left using catalinii/minisatip-build-image is the binaries.yml and coverity.yml. I'm working on fixing Coverity while @lars18th is working on fixing the image builds.

Issue #1173, opened 2024-08-20
[REQUEST] Update build of MIPS targets


lars18th:
Hi @catalinii and @Jalle19 ,

I've created a new (manual) workflow to create MIPS builds. You can check it:
https://github.com/lars18th/minisatip/blob/master/.github/workflows/binaries-manual-mips.yml

Some comments with the objective of reusing some parts to update the current master `binaries.yml` workflow of the project:
- The extra flag `-Wno-error=date-time` is necessary to compile with recent compilers. The current use of the `__DATE__` and `__TIME__` macros breaks any reproducible compilation. And new compilers refuse to compile (by default) non-reproducible binaries. So, I suggest to target this issue in the future. But in the meantime it will be useful to set this flag. You agree?
- The second point is to update the ZIG CC compiler. At time the binary is installed directly in the container `catalinii/minisatip-build-image`. But a recent (5 days ago) master version of zig has updated support for MIPS targets. Therefore, using the master release you can compile for MIPS GNU and MUSL targets without troubles. But the best point is that you can use the external Action `goto-bus-stop/setup-zig@v2` to setup the ZIG CC compiler. And it has cache support. So I suggest to change the container to move this ancient compiler to `/opt/catalinii/zig` and remove it from the PATH. Please, check my dirty workaround command `mv /opt/zig /opt/zig.disabled` in my workflow to overcome this compiler. My suggestion is to not remove the compiler, but not add it to the PATH. So you can still use it if you want it.
- And the last is to enable MUSL and GNU mips targets at the same time. And for the GNU version if you use `-target mipsel-linux-gnueabihf.2.30` then you will support Linux 4.x kernels and OpenATV 6.x devices. Because these releases are `static` then it has sense to support ancient devices. For more optimized versions, a cross-compiler for the specific target using dynamic linking will be preferable. But for these releases I recommend in this way.

I hope you want to update the releases.
Regards.

Jalle19:
IMO we should move away from minisatip-build-image, it only makes things more complicated right now. I need to figure out how to get Coverity Scan working first, then I'll have time to look at this.

Jalle19:
@lars18th can you make a pull request with your new workflow, but remove `container: catalinii/minisatip-build-image:latest`? Then we'll fix whatever needs fixing for the build to succeed.

lars18th:
Hi @Jalle19 

[...]

PR #1177, opened 2024-08-29
New workflow: binaries-manual.mips.yml

lars18th:
This workflow is standalone and it compiles MIPS releases (static) with the ZIG CC compiler.

(answer to https://github.com/catalinii/minisatip/issues/1173#issuecomment-2311648555)


catalinii:
The issue here is that is still missing DVBCSA and libssl. Basically oscam and CI will be disabled for enigma.

catalinii:
I re-built a new minisatip-build-image that is updated and that works.

PR #1182, opened 2024-09-04
Update links in README.md

lars18th:
Link to the local copy of the PDF document specification.


lars18th:
Hi,

Because I need to read the document sometimes, I prefer to have the full link in the README. 😉

Issue #1180, opened 2024-08-31
select_and_execute: select() error 4: Interrupted system call


Delitants:
Keeps randomly crashing when working as sat ip client. Systemd set to restart it automatically.

[untitled text 7.txt](https://github.com/user-attachments/files/16822925/untitled.text.7.txt)

[Unit]
Description=Multi-threaded satipserver
After=network-online.target

[Service]
User=root
Group=root
#WorkingDirectory=/root/minisatip
Type=forking
ExecStart=/usr/local/bin/minisatip -I minisatip -O -s dvbs2:192.168.20.230:554 -y 554 -l general -F /var/log/minisatip.log
TimeoutSec=120
Restart=on-failure
RestartSec=5
StartLimitInterval=350
StartLimitBurst=20

[Install]
WantedBy=multi-user.target

catalinii:
After crashing, is there something in syslog / dmeag

Delitants:
> After crashing, is there something in syslog / dmeag

Aug 30 16:08:57 neolo-KBN-I kernel: [ 5195.551698] dvb_ca_en50221: dvb_ca adapter 0: DVB CAM detected and initialised successfully
Aug 30 16:08:57 neolo-KBN-I systemd[1]: minisatip.service: Deactivated successfully.
Aug 30 16:09:19 neolo-KBN-I systemd[1]: Reloading.

That's it. But service is already enabled.

catalinii:
Can you build with “make debug” and try to repro?

lars18th:
Hi @Delitants and @catalinii ,

If this trouble is _really_ related to the `satipc` functionality I need to comment that I've detected a big problem with this module. So I'm working on a fix. Perhaps it will solve (or not) your problem. So I suggest waiting for it to be shared (hopefully within a week).

Regards.

Jalle19:
It doesn't crash, it exits cleanly due to:

```
[30/08 14:32:26.045 main]: select_and_execute: select() error 4: Interrupted system call
[30/08 14:32:26.046 main]: The main loop ended, run_loop = 0
```

No idea why that happens though.

Delitants:
> Hi @Delitants and @catalinii ,
> 
> If this trouble is _really_ related to the `satipc` functionality I need to comment that I've detected a big problem with this module. So I'm working on a fix. Perhaps it will solve (or not) your problem. So I suggest waiting for it to be shared (hopefully within a week).
> 
> Regards.

What satipc is responsible for?



[...]

PR #1183, opened 2024-09-05
LOG the running mode

lars18th:
Print in the log the running mode: background/foreground


lars18th:
This PR is to add some "missing" info.
I detected this from #1180, because the message `running in background` in `becomeDaemon()` is printed to the console and not to the log. Therefore, it has sense to store this info in the log.
Ready to merge.

PR #1184, opened 2024-09-09
Fix incorrect device_id initialization

lars18th:
This fix solves the problem of messages `Device ID conflict, changing our device id to...` at the start caused by an incorrect initialization. Without this fix the values from `-D --device-id` parameters and `bootid` file are not used.

### DIFF
--- a/src/minisatip.c
+++ b/src/minisatip.c
@@ -654,7 +654,7 @@ void set_options(int argc, char *argv[]) {
     opts.timeout_sec = 30000;
     opts.adapter_timeout = 30000;
     opts.daemon = 1;
-    opts.device_id = 1;
+    opts.device_id = 0;
     opts.dvr_buffer = DVR_BUFFER;
     opts.adapter_buffer = ADAPTER_BUFFER;
     opts.udp_threshold = 25;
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @Jalle19 ,

Please merge it. When running with multiple SAT>IP servers in the same network you will see the trouble.

Jalle19:
I'm not entirely sure what the problem is here? Device ID initialization from bootid file seems to work fine at least?

lars18th:
> I'm not entirely sure what the problem is here? Device ID initialization from bootid file seems to work fine at least?

NO. With the last changes of the initialization a BIG BUG is present: The code _allways_ starts to count from 1. Therefore the stored value and the enforced value of the parameter is _never_ used.

This patch is necessary. Please, merge it.
Thank you.

Jalle19:
Okay, I think I get it now. This is mainly about fixing the scenario where a conflict is detected, `opts.device_id` is incremented and `readBootID()` is called again?

lars18th:
> Okay, I think I get it now. This is mainly about fixing the scenario where a conflict is detected, `opts.device_id` is incremented and `readBootID()` is called again?

Not really. With the "previous" code the "initial" value is allways 1, and the paramters and bootid data is overwrited.

And thank you for merging it! 😉

PR #1186, opened 2024-09-09
Fix target device_id when other device is detected

lars18th:
When the enforced/stored device_id is already used, use another one free.


lars18th:
Hi @Jalle19 ,

One edge case requires to be targeted: when the enforced/stored value is already used. In this case, the next free value requires to be used. This simple patch solves this case.

Please, merge it. And sorry for the error.

PR #1187, opened 2024-09-10
Fix satipc module when TEARDOWN is not correctly handled

lars18th:
This fixes the edge case when a new tuning call is received in the tuner when the TEARDOWN message is sended to the remote SAT>IP server but the confirmation (OK 200) is not already received. Futhermore improves a bit the logging.


lars18th:
Hi @catalinii ,

After a lot of time, I've fixed the `satipc` module. The problem occurs most often when the clients are HTTP with fast channel changes. In this case the close of the adpater and the reopening of it occurs while the TEARDOWN is sended and waiting to receive the response. This finally fixes the problem.

As a note, the three `//LOGM->LOG` comments are not an error. I prefer to leave them to quickly edit the source code and then enable printing of full client and satipc messages without enabling verbose debug (more difficult to read).

Please test it and merge.
Regards.

Jalle19 (approved):

catalinii (approved):

Issue #1118, opened 2023-06-07
[BUG]: "satipc" does not work properly when the tuner is in use.


lars18th:
Hi @catalinii ,

When using the `satipc` module if you "share" the source adapter then you will be in troubles when the tuner is in use.

Here a simple tcpdump log with the problem:
```
18:48:29.137741 IP 192.168.1.51.56302 > 192.168.1.82.rtsp: Flags [P.], seq 3586096098:3586096289, ack 2841071189, win 501, length 191: RTSP: PLAY rtsp://192.168.1.82:554/?freq=... RTSP/1.0
18:48:29.138591 IP 192.168.1.82.rtsp > 192.168.1.51.56302: Flags [P.], seq 1:127, ack 191, win 473, length 126: RTSP: RTSP/1.0 404 Not Found
18:48:29.138756 IP 192.168.1.37.54071 > 192.168.1.82.rtsp: Flags [.], ack 2298112, win 21, length 0
18:48:29.138823 IP 192.168.1.82.rtsp > 192.168.1.37.54071: Flags [P.], seq 2298112:2299324, ack 1, win 490, length 1212: RTSP
18:48:29.140357 IP 192.168.1.51.56302 > 192.168.1.82.rtsp: Flags [.], ack 127, win 501, length 0
18:48:29.140519 IP 192.168.1.51.56302 > 192.168.1.82.rtsp: Flags [P.], seq 191:274, ack 127, win 501, length 83: RTSP: DESCRIBE rtsp://192.168.1.82:554/ RTSP/1.0
```

As you can see the problem is this:
- the minisatip satipc requests a SETUP with new tunning parameters.
- The SAT>IP server (that is another minisatip process) has the adapter in use, therefore it responds with 404 error.
- The minisatip satipc receives the response and starts to process a restart of the adapter.
- And this loop will continue until the adapter will be free.

However, in the minisatip server using the satipc module I've more adapters. And the problem is that instead of searching for another (next) free adapter it will continue reopening (aka restarting) the same adapter.

Any idea to fix this?

lars18th:
Hi @catalinii ,

The minisatip LOG when the problem appears:
```
[07/06 19:28:42.648]: satipc: commit for adapter 3 freq 11347000, pids to add 15, pids to delete 0, expect_reply 0, force_commit 0 want_tune 1 do_tune 1, force_pids 0, sent_transport 0, sleep 0, closed_rtsp 0
[...]

PR #1190, opened 2024-09-13
Handle occupied remote SAT>IP server

lars18th:
Within the satipc module, when a remote SAT>IP server sends a 404 "Not Found" response instead of retrying the Request it stays quiet.

### DIFF
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -272,9 +272,8 @@ int satipc_reply(sockets *s) {
             ad->tp.mtype = QAM_256;
     }
 
-    if (rc == 404) {
-        LOG("satipc %d: Ocuppied server, can't use it", s->sid);
-        sip->state = SATIP_STATE_INACTIVE;
+    if (rc == 404 && (strstr((char *)s->buf, "minisatip") != NULL)) {
+        rc = 503; // Hack the response from minisatip!
     }
 
     // quirk for Geniatech EyeTV Netstream 4C when fe=x is in the URL
@@ -283,7 +282,8 @@ int satipc_reply(sockets *s) {
             sip->satip_fe = 0;
 
     if (rc == 454 || rc == 503 || rc == 405) {
-        sip->state = SATIP_STATE_OPTIONS;
+        LOG("satipc %d: Ocuppied server, can't use it", s->sid);
+        sip->state = SATIP_STATE_INACTIVE;
         sip->last_setup = -10000;
     } else if (rc != 200) {
         if (rc != 0) // AVM Fritz!Box workaround sdp reply without header
--- a/src/satipc.c
+++ b/src/satipc.c
@@ -273,7 +273,8 @@ int satipc_reply(sockets *s) {
     }
 
     if (rc == 404) {
-        sip->state = SATIP_STATE_OPTIONS;
+        LOG("satipc %d: Ocuppied server, can't use it", s->sid);
+        sip->state = SATIP_STATE_INACTIVE;
     }
 
     // quirk for Geniatech EyeTV Netstream 4C when fe=x is in the URL


lars18th:
This fixes #1118

lars18th:
Hi, I've updated the patch to be more generic. Instead of retry forever in cases of 454, 503 or 405 (and now 404 from minisatip as well) the adapter goes to the inactive state because the request can't be served.

Pending: How to change to other free adapters.

catalinii:
Why not returning 503 (on the rtsp code path) when there are no adapters?

lars18th:
Hi @catalinii ,

1) I'll improve a bit this PR with a simple enhancement: "configure the behaviour when the SAT>IP server is occupied". The idea is simple: add the parameter `--satipc-busy-timer [msec]`. So don't merge it now.

2) I don't know what signifies "Why not returning 503 (on the rtsp code path) when there are no adapters". At time minisatip returns 404 when no tuner is available. And this return code is multiple files/lines. I don't think this is a problem. We can handle it as it's implemented in this PR. Or you want to say something different?

lars18th:
Hi @catalinii ,

I need you help to implement the `--satipc-busy-timer [msec]` option. The user functionality is quite simple:
- If "msec=0" and we receive any "busy response", then state changes to SATIP_STATE_INACTIVE.
- In any other case we wait for "msec" and then we retry the request from the SATIP_STATE_OPTIONS.

I feel this has sense and it's easy to understand. However, to implement I don't know something: the current `satipc->last_setup` member is not really used. I can use `sip->last_setup = getTick();` to set it when we need to set the timer for retrying the request. But I don't know how I can use the timeouts.

Any idea to set another timer inside the `satipc` module to trigger some action?

catalinii:
Hi,

why not adding a new structure member with this specific name (or rename last_setup).
About timeout, you can add a new sock like this and set the timeout in the next line:
https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L586

Also do u have some logs on what happens ?
Basically if get 404/503 from the server when you tune first time, you should return basically the same code to the upstream client and this should be the default.
This would ensure the correct behavior from other clients (like TVH) which could use another tuner to process the request

lars18th:
> why not adding a new structure member with this specific name (or rename last_setup). 

My idea is to reuse `last_setup` because in the current code is not used.

> About timeout, you can add a new sock like this and set the timeout in the next line: https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L586

Thank you. That's the point that I need. 😉 

> Also do u have some logs on what happens ? Basically if get 404/503 from the server when you tune first time, you should return basically the same code to the upstream client and this should be the default. This would ensure the correct behavior from other clients (like TVH) which could use another tuner to process the request

With the current implementation of the `satipc` module the "tuner" (aka adapter) status is handled _locally_ by de module. Therefore any request from the upper layers is not waiting for the response. And then when you receive the 404/503 response from the remote SAT>IP you've already sended to the upper layer (the client) an OK to the request that has originated the "busy tuner" response (actually a `PLAY/SETUP` command or a simple `HTTP GET`).

With this architecture what the _current_ code is doing is this (checked with logs): when the 404/503 response is received the local state is changed to send a new SETUP/PLAY request and _retry_ the last command. This is more or less useful _if you not overload the remote SAT>IP server_. Because when the tuner will be free the request will be served. But for the problem of the efficiency this PR is necessary. Without it an infinite loop appears... until the client closes the connection. And this infinite loop could generate troubles in the network and/or the remote server. So, the _easy_ solution is to almost "disable any retry" and not serve the request.

But my further proposal is to leave the user the option to fine-tune this behaviour. With the parameter `--satipc-busy-timer [msec]` we enable the option to disable all retries or control the time between them. I feel it's sufficient flexible and easy to add to the current code. You agree with that?

But in any case, before the implementation of this new parameter I recommend to merge this PR. Without it it would be very easy to overload the server. So please merge it now. And I'll provide another one later witht he `--satipc-busy-timer [msec]` option.

Jalle19:
Let's close this for now, it's been stale for almost a year

PR #1212, opened 2025-03-15
Fix 0% strength/SNR due to DVB API legacy fallback

### LINKED ISSUE #1211: DVB API: 0% strength/SNR due to DVBAPI legacy fallback
When using a DVB device, after tuning/locking the signal strength and SNR are sometimes not yet available when they are retrieved the first time. [In that case](https://linuxtv.org/downloads/v4l-dvb-apis-new/userspace-api/dvb/frontend-stat-properties.html#dtv-stat-signal-strength), the scale `FE_SCALE_NOT_AVAILABLE` is returned instead.

At the moment this results in an immediate fallback to legacy DVBAPI calls for further retrivals of the status, which will never succeed though. As a consequence, signal strength and SNR will always stay at 0% then. I was able to observe this behaviour with a [Hauppauge WinTV-soloHD](https://www.hauppauge.co.uk/site/webstore/webstore_solohd.html).

Note that [another German user already observed this behavour a while ago](https://www.vdr-portal.de/forum/thread/135858-minisatip-vdr-plugin-satip-probleme/?postID=1362775#post1362775) and apparently posted it to the official board (I wasn'n able to find the corresponding thread though).

basicmaster:
If strength/SNR are not immediately available after tuning/locking, the scale `FE_SCALE_NOT_AVAILABLE` (= 0) is returned. Don't fall back any longer to legacy DVB API for further status requests, as they lead to 0% strength/SNR then, due to failing legacy calls.

Fixes #1211.

### DIFF
--- a/src/dvb.c
+++ b/src/dvb.c
@@ -1853,8 +1853,9 @@ int get_signal_new(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
         strength_s = "dBm";
         init_strength = enum_cmdargs[0].u.st.stat[0].svalue; // dBm / 1000
         strengthd = get_strength_decibels(init_strength);
-    } else if (enum_cmdargs[0].u.st.stat[0].scale == 0)
-        err |= 1;
+    }
+    //  else if (enum_cmdargs[0].u.st.stat[0].scale == FE_SCALE_NOT_AVAILABLE)
+    //      err |= 1;
 
     if (enum_cmdargs[1].u.st.stat[0].scale == FE_SCALE_RELATIVE) {
         snr_s = "%";
@@ -1864,7 +1865,7 @@ int get_signal_new(adapter *ad, int *status, uint32_t *ber, uint16_t *strength,
         init_snr = enum_cmdargs[1].u.st.stat[0].svalue; // dB * 1000
         *db = (int)get_snr_decibels(init_snr, &snrd, ad->db_snr_map);
     }
-    //	else if (enum_cmdargs[1].u.st.stat[0].scale == 0)
+    //	else if (enum_cmdargs[1].u.st.stat[0].scale == FE_SCALE_NOT_AVAILABLE)
     //		err |= 2;
 
     *ber = enum_cmdargs[2].u.st.stat[0].uvalue & 0x7FFF;


Jalle19:
Can you make the actual changes instead of writing the new code as comments?

basicmaster:
Actually I don't add new code here - I just disable the `if` clause that sets `err` to a non-zero value in case of `FE_SCALE_NOT_AVAILABLE`. I wanted to leave the two clauses as comment in the code for easier debugging, replacing the actual constant value `0` with the particular constant name.

Jalle19:
I see. @catalinii what do you think?

Jalle19 (approved):

catalinii:
This patch will work for devices that supports the new ioctl but not for old devices that does not.

The issue happens in dvb_get_signal, if you do not return err=1, it will always use get_signal_new.

the other option is to see rely on get_signal (old ioctls for getting signal) only when the ioctl for getting new signal fails, rather than detecting values

Jalle19:
IIRC in tvheadend there's a checkbox to use the old API, maybe we should do the same?

basicmaster:
To support both old and new devices, we could invoke the new ioctl more than once (e.g. 5 to 10 times) in order to see whether we get valid values sometime. If we don't, we would finally fall back to the old ioctl then.
This way, older devices would get a plausible status at least after some calls. I could adjust the MR accordingly...

> IIRC in tvheadend there's a checkbox to use the old API, maybe we should do the same?

Hmm, this would require manual user interaction, and (assuming it is a general setting) would not allow to mix older and newer devices.

catalinii:
Enigma seems to use a config value to force old API too. I think it should be good considering the oldest kernel should be on enigma devices.

Jalle19:
While @basicmaster has a good point, I think it can be considered rare that someone would have a setup with a mix of tuners that require both DVBAPI versions. Nevertheless, the option could be per tuner instead of global. Polling a few times then falling back is a nice solution but perhaps needlessly complex and thus errorprone.

basicmaster:
> Polling a few times then falling back is a nice solution but perhaps needlessly complex and thus errorprone.

IMHO this should be relatively simple, adding a counter variable on adapter level and (carefully!) adding some lines to `dvb_get_signal`. Before that I would have to do a few tests, to determine a reasonable value for the necessary number of tries.

> Enigma seems to use a config value to force old API too. I think it should be good considering the oldest kernel should be on enigma devices.

This would be a behaviour change, at least for users that need to use the old API i.e. they would have to explicitely specify that they want to use the old API, instead of relying on the current "autodetection". To address this one could keep the current autodetect approach as default, but also add a parameter that allows to explicitely specify the to-be-used API version (i.e. the desired value for `new_gs`).

That new parameter could be set per tuner and/or in general, similar to the `--diseqc` parameter.

catalinii:
In general you are right. But in this case the new_signal drivers were introduced like 14 years ago, so it should be really really rare when someone needs it.

i think the current logic was flaky but rather than making the code more complex simplifying in this case may be best.

if you want to implement it the way you are proposing, I am not against it, but try to add some unit tests to ensure the autodetection works properly.

basicmaster:
> In general you are right. But in this case the new_signal drivers were introduced like 14 years ago, so it should be really really rare when someone needs it.
> 
> i think the current logic was flaky but rather than making the code more complex simplifying in this case may be best.

Thinking about it, I agree with that. I added a possible solution in #1213.

basicmaster:
(closing in favour of #1213)

PR #1220, opened 2025-04-13
Fix assemble packets with adaptation

catalinii:


Jalle19 (approved):
Can't comment on the specifics, but there is now both less code and less magic numbers, so that's good

PR #1226, opened 2025-06-10
Fix slave adapter assignments

### LINKED ISSUE #1225: dreambox dm920 ultrahd, 2 tuner cards, 16 DVBS2X tuners problem.
Hello,

I have a Dreambox 920 with two DVBS2X FBC tuner cards, so a total of 16 tuners. 
When I turn off Enigma2 and start minisatip with all Jess options, only 8 tuners [0-7] from slot 1 work. 
Tuner 0 delivers 4 satellites, and tuner 1 also delivers 4 satellites via src[1,2,3,4] from Slot 1.

Problem 1: With the -s options, I'm unable to bind tuners 5-7 to tuner 1 in Slot 1.

Problem 2: Slot 2 [tuners 8-15] don't deliver any streams. Here, tuners 8 and 9 would be the master.

I tested this with openvix and GP4.2 Nobody Image. everything runs smoothly in enigma2, but I want to run it via satip.

Have you ever heard of this? Or is it a hardware problem that can't be solved.

Jalle19:
Might fix #1225 , fixes certain assignment configurations on AXE devices at least.

Without this patch, `0-1:0,2-3:2` results in the following, which is wrong:

> Setting master adapter 0 for adapter 0
> Setting master adapter 3 for adapter 2

With this patch, the output is correct:

> Setting master adapter 0 for adapter 0
> Setting master adapter 0 for adapter 1
> Setting master adapter 2 for adapter 2
> Setting master adapter 2 for adapter 3

Verified at runtime as well:

### DIFF
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -1887,7 +1887,7 @@ void set_slave_adapters(char *o) {
                 continue;
 
             if (master >= 0 && master < MAX_ADAPTERS) {
-                if (!a[master] && !is_adapter_disabled(master))
+                if (!a[master])
                     a[master] = adapter_alloc();
 
                 if (a[master] && a[master]->master_source < 0) {
--- a/src/adapter.c
+++ b/src/adapter.c
@@ -1890,9 +1890,7 @@ void set_slave_adapters(char *o) {
                 if (!a[master])
                     a[master] = adapter_alloc();
 
-                if (a[master] && a[master]->master_source < 0) {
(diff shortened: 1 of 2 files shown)


Jalle19:
Actually this just breaks something else - one of the adapters gets closed after a while and the stream stops.

I think fundamentally the concept of `master_source` is incompatible with what AXE uses to assign inputs. The option should never have been overloaded to have multiple meanings.

Issue #1227, opened 2025-06-12
Stream doesn't start with VLC (minisatip with Fritz!Box)


retrofan1979:
Hello,

When trying to start a SAT>IP url from minisatip with VLC, the stream doesn't start even though the url is correct.
I tried both rtsp with port 554 and http with port 8080.

I'm using the Fritz!Box 6690 Cable with this command line to install a tuner from the Fritz!Box:

sudo ./minisatip --satip-xml http://192.168.178.1:49000/satipdesc.xml -f

It shows the 4 DVB-C tuners, but I cannot use them in VLC.

Alternative commandline to use 2 tuners:
sudo ./minisatip -s dvbc:192.168.178.1 -s dvbc:192.168.178.1 -f

Attached is the extensive logfile with all possible options for debugging.
Can you help me out?

VLC is version 3.0.21, the latest one.

[minisatip.zip](https://github.com/user-attachments/files/20713703/minisatip.zip)

retrofan1979:
If the logfile is too big, please let me know and instruct me which data is needed to investigate this issue.

retrofan1979:
Maybe someone else with a Fritz!Box Cable can test this as well? Thanks!

retrofan1979:
Looks like port 554 is already in use, but I cannot see it.

sudo lsof -i :554
COMMAND   PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
minisatip 779 root    3u  IPv4   5511      0t0  TCP *:rtsp (LISTEN)

retrofan1979:
Please check with latest Fritz!OS version 8.03 as well.

retrofan1979:
I also tried DVBViewer Media Server and this one works fine. Could this be a bug in minisatip?

catalinii:
This is generally a VLC bug. The stream needs to arrive very fast after starting. Maybe using another satip server (fritzbox) increases the latency

retrofan1979:
Okay, I see, thanks.
Is is possible for you to make a quirck for VLC?
I would like to use http protocol with port 8080 from minisatip.

I'm using the correct parameters for Fritz!Box?

catalinii:
Try to replace https://github.com/catalinii/minisatip/blob/master/src/satipc.c#L163 minisatip with FRITZ and lets see if it works

retrofan1979:
I'm sorry, but it doesn't work.
[...]

Issue #1164, opened 2024-07-17
Fritzbox response quirk broken due, response code parsed incorrectly


retrofan1979:
Hello,

I noticed when I try to do a channelscan with DVBViewer Pro and TransEdit with minisatip as SAT>IP server, it rejects the connection.
However, when I play an SAT>IP URL it streams just fine.
Also when analyzing a mux with TransEdit it works fine.
Only a channelscan fails. Hope you can fix it.
I'm using a Fritz!Box 6690 cable.
See logfile attached.

retrofan1979:
![Schermafbeelding 2024-07-17 175358](https://github.com/user-attachments/assets/a2d1ab9f-871e-443f-999d-2ef482365962)
[minisatip.log](https://github.com/user-attachments/files/16269179/minisatip.log)

retrofan1979:
No clues? The logfile is attached in previous comment. Thanks!

xxj737:
pid=all  
Not Supported

retrofan1979:
Thanks, but with TransEdit I can scan a mux with over 100 pids, so no problems here. When I want to do a normal channelscan I get this error.

xxj737:
for example
![微信截图_20240727145624](https://github.com/user-attachments/assets/19b97800-5229-40ac-8e8f-f8c3e80f943b)

retrofan1979:
with SAT>IP server of Fritz!Box 6690 Cable I have no problems at all, but with minisatip the channelscan fails, so it seems a bug in minisatip? Hope you can fix it.

retrofan1979:
Fritzbox:

![fritzbox 1](https://github.com/user-attachments/assets/db4a1cb7-f538-4861-8703-3f2d0ae351ea)

![fritzbox 2](https://github.com/user-attachments/assets/dfc2eb49-3979-4254-a40a-2f50744f9d7c)

retrofan1979:
minisatip:

![minisatip 1](https://github.com/user-attachments/assets/38434cd1-9b52-438b-817d-0de21eec6bcf)

![minisatip 2](https://github.com/user-attachments/assets/ff382d2a-0bc4-4fe1-8563-f49c0658deff)

retrofan1979:
I found out when selecting option "Open whole transponder" it will scan one mux at the time. With Fritzbox this option isn't needed.

retrofan1979:
[...]

PR #1230, opened 2025-07-01
Wait for the entire RTSP message #1227

catalinii:


retrofan1979:
thanks, but I think the waitstate is still not long enough. see log attached. The stream isn't opnened by VLC.

[minisatip.log](https://github.com/user-attachments/files/21003771/minisatip.log)

Let's try again  @catalinii :-)

retrofan1979:
When paring --satip-xml http://192.168.178.1:49000/satipdesc.xml to minisatip command-line, I get this:

[01/07 17:49:22.248 main]: minisatip version 1.3.~, compiled on Jul  1 2025 17:29:07, with s2api version: 050C
[01/07 17:49:22.276 signal]: Initializing with 0 devices

0 devices? It should be 4 in my case.

However, after a few seconds the status page shows 4 tuners.

image

retrofan1979:
Maybe you also can check the satip server part of Tvheadend:
https://github.com/tvheadend/tvheadend/blob/master/src/satip/server.c

catalinii:
Can you provide remote access? catalinii@gmail.com

retrofan1979:
No, I can't, sorry. But you can compare it with version 1.3.22 when it worked or I can give you a extensive log where needed.
Maybe the sources of the satip server part of Tvheadend are useful as well:
https://github.com/tvheadend/tvheadend/blob/master/src/satip/server.c
This one also works with fritzbox.

catalinii:
Can you try now?

retrofan1979:
Hi @catalinii thanks for trying, but still the stream isn't opened by VLC, however it get the correct pids. 

BUT.... I did another approach. I copied the zipfile of version 1.3.52 from minisatip to my RPI5 and replaced the satipc.c file with the one of version 1.3.22 and SUCCESS, the streams starts in all possible modes as it should in VLC (HTTP, RTP/UDP and RTP/TCP, nice!). So, the bug or invalid change for fritzbox is probably only in the satipc.c file. And as far as I can see, there are not much differences between those two. :-) Will do some more tests, but maybe you'll see it earlier :-)

You can compare those files and the log files over here.

[satipc.c compare.zip](https://github.com/user-attachments/files/21058358/satipc.c.compare.zip)

[minisatip #1230.log](https://github.com/user-attachments/files/21058273/minisatip.1230.log)
[minisatip 1.3.52 with satipc.c from 1.3.22.log](https://github.com/user-attachments/files/21058315/minisatip.1.3.52.with.satipc.c.from.1.3.22.log)

P.S.: how to compile a commit directly from source instead of the master?

catalinii:
git clone https://github.com/catalinii/minisatip
git checkout origin/issue_1227

sudo ./configure
sudo make

catalinii:
Can u provode a new log with this commit, the one attached is not conclusive

retrofan1979:
> Can u provode a new log with this commit, the one attached is not conclusive

Sure, here it is. ;-)

[minisatip.log](https://github.com/user-attachments/files/21076088/minisatip.log)

retrofan1979:
I found the problem! When I undo this commit, the stream will open in VLC https://github.com/catalinii/minisatip/commit/cdf1749f6606010d764dbbb247451e91234402ec
Ofcourse I also used the latest release of minisatip: https://github.com/catalinii/minisatip/releases/tag/v2.0.0-beta.6

I don't know why the fritzbox doesn't like this change (RTSP return code).
What does this commit exactly do? What's the purpose of it?

Jalle19:
The commit you mentioned was to fix another issue you yourself reported: https://github.com/catalinii/minisatip/issues/1164

retrofan1979:
> The commit you mentioned was to fix another issue you yourself reported: #1164

O, that's funny, I didn't realize that, thanks :-) That fix is still needed :-)
Hmm, then it's indeed probably difficult to get both VLC and DVBViewer working at this point.
Fritzbox doesn't like RTSP return code fix and the RTSP header is splitted in 2 parts. Maybe it also needs more buffer to prevent discontinuities.

catalinii:
Can you add -l http,satipc
The issue seems the fact that no pids are passed to fritzbox but need to confirm that

catalinii:
This diff is working as expected and will merge it, will continue the fix of this issue in a separate diff

retrofan1979:
> Can you add -l http,satipc The issue seems the fact that no pids are passed to fritzbox but need to confirm that

O, I hope you're right. That would be nice! Here's the log.

[minisatip.log](https://github.com/user-attachments/files/21077963/minisatip.log)

PR #1235, opened 2025-07-07
switch http to check header is complete

catalinii:
Refactor the http/rtsp request buffering only of the header is complete

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -832,14 +832,20 @@ void _strncpy(char *a, char *b, int n) {
 }
 
 /**
- * Checks if the given buffer contains a valid RTSP header.
+ * Checks if the given buffer contains a valid RTSP/HTTP header.
  *
  * @param buf The buffer containing the data to be checked.
  * @param len The length of the buffer.
  * @return 1 if the buffer contains a valid RTSP header, 0 otherwise.
  */
-int is_rtsp_http_header(char *buf, int len) {
-    if (strncmp(buf, "RTSP/1.", 7) && strncmp(buf, "HTTP/1.", 7))
+int is_rtsp_http_header(char *buf, int len, const char *start[], int lstart) {
+    int has_start = 0;
+    for (int i = 0; i < lstart; i++) {
+        if (!strncmp(buf, start[i], strlen(start[i])))
+            has_start = 1;
+    }
+
+    if (!has_start)
         return 0;
 
     char *nlnl = strstr(buf, "\r\n\r\n");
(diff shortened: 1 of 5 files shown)


src/utils.cpp
```diff
@@ -842,8 +842,8 @@ void _strncpy(char *a, char *b, int n) {
  * @param len The length of the buffer.
  * @return 1 if the buffer contains a valid RTSP header, 0 otherwise.
  */
-int is_rtsp_header(char *buf, int len) {
-    if (strncmp(buf, "RTSP/1.", 7))
+int is_rtsp_http_header(char *buf, int len) {
+    if (strncmp(buf, "RTSP/1.", 7) && strncmp(buf, "HTTP/1.", 7))
```
Jalle19: This new logic is not covered by the existing tests

retrofan1979:
I'm sorry, but the stream doesn't open in vlc.
The previous version was better but stream stops after 1 minute and it's stuttering.

[minisatip.log](https://github.com/user-attachments/files/21126427/minisatip.log)

catalinii:
Generally build with debug and run ./minisatip -f -l http,satipc -s …. I do not see the http/rtsp traffic
This is painful for both of us, but if you change your mind about the remote access is sufficient to forward port 554 to fritzbox.

Also, you do not need to pass the signal multipliers anymore

catalinii:
Please re-test the latest commit from this branch

retrofan1979:
This is all I get with:

make clean && make debug && ./minisatip -f -s dvbc:192.168.178.1:554 -l http,satipc 2>&1 | tee /tmp/log
[minisatip.log](https://github.com/user-attachments/files/21129746/minisatip.log)

retrofan1979:
Anyway, thanks for your help so far, much appreciated

retrofan1979:
The sources of tvheadend are not useable for you or octonet? Maybe I also can capture a stream from DVBViewer for you? It streams a channel from minisatip as long as I don't use SAT>IP urls.

catalinii:
Can you try now?

catalinii:
Generally logs are sufficient but there is a lot of back and forth

retrofan1979:
New try, still no stream, but it's there as the logs says (attached).
But there are no pids! pids=none!
rtsp://192.168.178.1:554/?freq=274.0&sr=6900&msys=dvbc&mtype=256qam&pids=none

[minisatip.log](https://github.com/user-attachments/files/21134914/minisatip.log)

retrofan1979:
Hmmm, don't know, played around a bit more with urls in vlc and now suddenly the stream plays and the log shows activity as well. Looks like the initializing doesn't go well in the first place? However, the stream still stops after 1 minute.

[minisatip.log](https://github.com/user-attachments/files/21146970/minisatip.log)

retrofan1979:
don't know if this is relevant what I read on the internet:

_** It would be better to start using satip:// in the urls instead of rtsp://. SAT>IP boxes don't use proper RTSP, so satip:// url makes it more clear to the client that it should use the specific SAT>IP version of RTSP protocol. VLC does work with satip:// urls.**_

Also see over here:
https://ffmpeg.org/pipermail/ffmpeg-devel/2020-December/273910.html

Issue #1239, opened 2025-07-12
MIPS pre-built binaries are missing dvbcsa support


FritLP:
Such to some problems like unstable run e.g. with an older version of minisatip, i downloaded the newest version. Now i currently have the problem, that normal channels works fine, but private channels, like sky are down. 

Im using it on my VuSolo2 and started it with the following command:
./minisatip -b 50000:8192000 -B 8000 -R /usr/share/minisatip/html -x 8080 -o 192.168.178.40:9000 -G

This is the command i used already in the older version. The client connect successfully to oscam and when start a channel like sky, i can see in the oscam live log, that it should get up, but nothing. 

Can u help?

catalinii:
1) does a different provider (except sky) work?
2) feel free to upload the logs of ./minisatip -f -l http,pmt 
And try to tune to an encrypted channel

FritLP:
1. No, other provider doesnt work to. So its a general problem.
2. I will give u an log of minisatip, oscam, and a screen of the minisatip web ui. (attached files)

[LOG Minisatip.txt](https://github.com/user-attachments/files/21204856/LOG.Minisatip.txt)
<img width="1906" height="291" alt="Image" src="https://github.com/user-attachments/assets/2946dd51-feee-4b45-8a5c-79d8f19485d0" />
<img width="1104" height="207" alt="Image" src="https://github.com/user-attachments/assets/a6469b36-8cae-47e7-81f2-ce607aa05148" />

catalinii:
Use one of the binaries from github. 

You are missing libdvbcsa from here: https://github.com/catalinii/libdvbcsa 

@Jalle19 should we just fail minisatip and put a meaningful error message?

Jalle19:
Not really sure where a meaningful error message would be shown?

catalinii:
To compile dvbcsa when using oscam?

Jalle19:
The -o switch should not be available if dvbcsa support is not compiled in

FritLP:
Ok. I was using the latest mips release from the github. I will try to cross compile it.

Jalle19:
@catalinii maybe we should include dvbcsa support in the pre-built binaries?

catalinii:
The pre-build binaries have dvbcsa included and it should work, this is just for folks compiling

Jalle19:
He said he used the latest mips release from GitHub 🤷

FritLP:
[...]

PR #1246, opened 2025-07-17
Fix missing features in the MIPS builds

Jalle19:
Reference: https://github.com/catalinii/minisatip/issues/1239#issuecomment-3080711913

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -28,10 +28,11 @@ jobs:
        zip -9 -r /minisatip_arm.zip minisatip html
 
     - name: Build MIPS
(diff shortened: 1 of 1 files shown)


Jalle19:
Needs more work.

@catalinii why are we using zig for MIPS builds? I can get this to work locally with the standard mipsel-linux-gnu-g++ but zig complains about "error: undefined symbol: __fdelt_chk" etc.

catalinii:
I did not manage to compile the code using mips gcc (nor clang for that matter) ... if you can make it work I am ok removing zig

Jalle19:
Closing this for now

Issue #1240, opened 2025-07-15
TCP mode not working


retrofan1979:
When using -O parameter on my Fritzbox, it returns RTSP error 461 unsupported transport. However, it shows TCP mode on webpage.

[minisatip.log](https://github.com/user-attachments/files/21226294/minisatip.log)

<img width="1381" height="182" alt="Image" src="https://github.com/user-attachments/assets/d6c28fe5-c419-41ea-a555-2fee30ed8db1" />

catalinii:
Can you use tcpdump -A -s 0 port 554 when using tvh or VLC in tcp mode?

Jalle19:
When you tell tvheadend to use TCP, it uses that when communicating with minisatip. Minisatip still uses UDP when conmunicating with your Fritzbox. I don't think there's a bug here.

Jalle19:
The web interface could show whether UDP or TCP is used to connect to the SAT>IP server, would help with confusion.

retrofan1979:
> Can you use tcpdump -A -s 0 port 554 when using tvh or VLC in tcp mode?

I did a test with my favourite program DVBViewer.

[tcpdump_dvbviewer.log](https://github.com/user-attachments/files/21237053/tcpdump_dvbviewer.log)

retrofan1979:
> The web interface could show whether UDP or TCP is used to connect to the SAT>IP server, would help with confusion.

Yes, see picture above. It shows RTP/TCP.

catalinii:
Basically what I am asking is if VLC can pull the stream directly from fritzbox using TCP. Here, VLC is pulling the stream from minisatip

retrofan1979:
No, the fritzbox cannot use TCP directly AFAIK. That's why I'm using minisatip :-) tcpdump doesn't capture packets as it seems, only when I'm pointing to minisatip with fritzbox indirectly. No more dropouts as it seems as well.

catalinii:
There is not much we can do then

Jalle19:
> > The web interface could show whether UDP or TCP is used to connect to the SAT>IP server, would help with confusion.
> 
> Yes, see picture above. It shows RTP/TCP.

No, it only shows that the client is using TCP to talk to minisatip.

retrofan1979:
Okay, I see, thank you all. Then I close this issue.

Jalle19:
[...]

PR #1251, opened 2025-07-19
Fix is_rtsp_header() when using TCP

Jalle19:
For some reason there's an extra 4 bytes of padding in the buffer. Relax the check to simply ensure the buffer has at least the Content-Length available in it.

Fixes https://github.com/catalinii/minisatip/issues/1240


catalinii (approved):
@Jalle19 can u add some comments here ?

Jalle19:
@catalinii added

Issue #1231, opened 2025-07-01
xupnpd2 with minisatip


retrofan1979:
Hi,

I tried xupnpd2 to get it running with minisatip, but no luck so far. However, I get it running with tvheadend.

xupnpd2 is located over here: https://github.com/jacklul/xupnpd2

xupnpd2 is a DLNA server which can be handy to stream TV to DLNA certified devices like Playstation 4, XBOX360 etc.

catalinii:
Logs from playing 1 channel?

retrofan1979:
Here it is. Please search for text Panasonic in log. (192.168.178.31)

[minisatip.log](https://github.com/user-attachments/files/21087680/minisatip.log)

retrofan1979:
this syntax of tvheadend works:

#EXTM3U
#EXTINF:0,1
http://192.168.178.105:9981/stream/channelnumber/1

maybe you can do the same for minisatip? converting play url for SAT>IP into alias for client?
channelnumber can be extracted from LCN.

Example:
http://192.168.178.105:8080/channel/1

Jalle19:
minisatip uses correct URLs, if xupnpd2 can't handle them then the bug is in xupnpd2.

retrofan1979:
I guess so, but it would be nice if minisatip also supplies an alternative syntax like tvh.

Jalle19:
Don't really know what's wrong based on the logs, probably unrelated to xupnpd2 itself. I did notice this at the end though:

```
munmap_chunk(): invalid pointer
RECEIVED SIGNAL 6 (-6) - SP=0 IP=0
Obtained 0 stack frames.
```

catalinii:
I see 2 streams playing one without CC errors, one with CC errors.

The bandwidth seems is high for both those channels.

Before further investigations can you ensure both those URLs play with VLC (from minisatip directly)
Then try to use VLC to play thru xupnpd by looking for DLNA servers in the local network.

If yes, feel free to open an issue with xupnpd to see what are the errors on the xupnpd side.

I may be wrong but xupnpd is the one making the requests to minisatip, then we need to understand what breaks.

retrofan1979:
Yes, VLC does play the channel through DLNA with xupnpd2.
Maybe my TV doesn't like the comma sign?

Anyhow, no chance to add alternative syntax for streaming a channel as mentioned in this thread?
If not possible, you may close this topic. Otherwise it's a feature request :-)

catalinii:
[...]

PR #1255, opened 2025-07-19
Retry poll() on EINTR

Jalle19:
Fixes https://github.com/catalinii/minisatip/issues/1180, fixes https://github.com/catalinii/minisatip/issues/1231

### DIFF
--- a/src/socketworks.cpp
+++ b/src/socketworks.cpp
@@ -752,14 +752,17 @@ void *select_and_execute(void *arg) {
             break;
         }
         les = es;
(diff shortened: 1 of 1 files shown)


Jalle19:
Changed to use https://www.gnu.org/software/libc/manual/html_node/Interrupted-Primitives.html

Jalle19:
Macro not available on MIPS: `error: use of undeclared identifier 'TEMP_FAILURE_RETRY'`

catalinii (approved):

PR #1257, opened 2025-07-20
Minor build tweaks

Jalle19:
Let's see if we can get the MIPS build to work this way


Jalle19:
Can someone with MIPS and ARM test the binaries generated by https://github.com/catalinii/minisatip/actions/runs/16397142054/artifacts/3571616189 ? @FritLP?

FritLP:
I tried the mips version on my vusolo2 but cant execute it. So i think maybe a wrong toolchain?

image

FritLP:
I think its because of the MIPS32 Release 2 Version. The latest release, which actually working, is Mips32 Release 1. 
image

Jalle19:
@catalinii was this the reason you switched to zig? It doesn't seem to be possible to compile a "MIPS32 version 1" binary with clang

Jalle19:
Could be this too: https://stackoverflow.com/questions/58222225/gcc-cross-compile-mips1-not-generating-mips-i

Jalle19:
https://stackoverflow.com/questions/73130956/gnu-mipsel-gcc-does-not-seem-to-respect-the-architecture-option :shrug:

catalinii:
I stopped when I tried to build minisatip with gcc from enigma but the toolchain (libc + headers) from ubuntu and that still broke, which means the libs are not compiled correctly.

enigma gcc + libc(++) + headers works

so one option would be to just pack enigma toolchain and use that instead of zig

Jalle19:
Can't believe it's this hard to compile something for MIPS...

Jalle19:
Can't for the life of me get past this:

```
ld.lld: error: unable to find library -l:libdvbcsa.a
ld.lld: error: unable to find library -l:libcrypto.a
```

The files are right there, yet it doesn't work.

Jalle19:
Needs https://github.com/catalinii/minisatip-build-image/pull/4

catalinii:
Seems to have worked, but I see it still uses zig

Jalle19:
Yes, will have to keep using it for now

Issue #1265, opened 2025-07-28
Release binaries are missing version information


Jalle19:
`Minisatip Version: ~ -- Build date/time: Jul 27 2025 | 17:26:23`

I think this can be fixed by fixing those "unsafe git directory" errors in GitHub actions.

Jalle19:
Still need some fixing, now the binaries get the previous release. We should probably bump the tag before building the binaries and push it when we make the release and upload the binaries to the release. To be continued.

Jalle19:
All good now!

PR #1266, opened 2025-07-28
Mark GitHub workspace as "safe" for git

Jalle19:
Should fix https://github.com/catalinii/minisatip/issues/1265


Jalle19:
Before
```
-DTAG=\"\" -DREVISION=\"\"
```
Now
```
-DTAG=\"v2.0.6\" -DREVISION=\"5cfb905\"
```

Jalle19:
So the problem with this is that it uses the previous tag, not the new that gets created later by the action :/ Need to think some more

Jalle19:
I think we should add https://www.npmjs.com/package/semver to the Docker image (it's the same tool the GitHub action that bumps the version for us uses), use it to determine the bumped version, then pass that manually to the build

catalinii (approved):

Jalle19:
@catalinii ready to go now, I added semver and some tricks to be able to pass a version manually. Seems to work as intended.

Jalle19:
No idea where it's getting `55db457` from

Jalle19:
Reason for the mysterious ref is https://github.com/actions/checkout/issues/426

Jalle19:
I'm starting to think we should use `v2.0.6-PR1234-revision` instead

Jalle19:
Highly improved. Pull request builds now get e.g. `v2.0.6+PR1266-3ff7369`, while main branch builds remain unchanged functionally

PR #1269, opened 2025-07-30
gcc++ fixes

pbiering:
- add several missing casts for compiling with gcc-c++ 15
- add TODO for a leftover warning, potential use-after-free


pbiering:
can reproduce build error if all features are disabled, will check how to circumvent.

Jalle19:
What operating system ships with GCC 15? We could/should add it to the CI pipeline.

pbiering:
> What operating system ships with GCC 15? We could/should add it to the CI pipeline.

Al least Fedora 42, which I'm using for VDR.

catalinii:
Feel free to ignore use-after-free issue

PR #1270, opened 2025-07-30
WIP: MIPS DVBCSA build

Jalle19:


Jalle19:
The `--enable-foo` flags don't do anything anymore

Jalle19:
And we're back to `ld.lld: error: undefined symbol: __getsockopt64` etc.

Jalle19:
Well, this has never worked so no surprise we're running into issues trying to enable it now. I thought I had introduced a regression when I updated the build system somewhat but turns out the MIPS builds have always been without DVBCSA.

Jalle19:
I think the solution is to build OpenSSL for MIPS with the same compiler as we're building the binaries.

Jalle19:
Closing this, I don't have time/energy to improve the MIPS build

catalinii:
the mips build with libcsa and libssl does not work ?

Jalle19:
There is no MIPS build with libdvbcsa enabled

catalinii:
hmm, weird that zig works in the image, but is not included in the minisatip-build-image

Jalle19:
It's added by a GitHub action

catalinii:
now the compilation works, just some missing libraries from libcrypto for static build.

Jalle19:
Closing this for now

PR #1271, opened 2025-07-30
Automatically create debug builds too

Jalle19:
Let's see if `-g` instead of `-ggdb` does the trick

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -35,7 +35,7 @@ jobs:
 
     - name: Build x64
       run: |
-        ./configure --disable-netcv --enable-static CXX="clang++ -static"
+        ./configure --disable-netcv --enable-static
         make DDCI=1
         file minisatip
         zip -9 -r /minisatip_x64.zip minisatip html
@@ -43,7 +43,7 @@ jobs:
     - name: Build ARM
       run: |
         make clean
-        ./configure --disable-netcv --enable-static --host=arm-linux-gnueabihf CXX="clang++ -target arm-linux-gnueabihf -static"
+        ./configure --disable-netcv --enable-static --host=arm-linux-gnueabihf
         make
         file minisatip
         zip -9 -r /minisatip_arm.zip minisatip html
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -40,6 +40,12 @@ jobs:
         file minisatip
         zip -9 -r /minisatip_x64.zip minisatip html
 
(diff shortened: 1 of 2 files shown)


Jalle19:
Testing v2.0.10+PR1271~e416f1a locally now, crashes about once a day so I'll get some results sooner or later

catalinii:
The release build code to generate stack traces is completely broken, was planning to add < stacktrace > and migrate to c++ 23 code.

the other options involve taking another dependency on other libs. I tried libunwind but is not so easy

see this: https://github.com/catalinii/minisatip/blob/master/src/utils.cpp#L242

Jalle19:
Where does backtrace() come from?

catalinii:
It’s a glibc function but not part of musl & co

Jalle19:
Okay, so it's possible these current changes may end up producing proper traces in the x64 build at least?

catalinii:
It depends on addr2line which may or may not be installed

catalinii:
I think the best option is < stacktrace > feature introduced in c++ and we can do it conditionally (if c++ 23 is available add stacktrace functionality).

or just enforce c++ 23

Jalle19:
Yes, `<stacktrace>` sounds like a good solution

Jalle19:
```
Aug 02 23:25:48 vladimir minisatip[5755]: RECEIVED SIGNAL 8 (1) - SP=0 IP=0
Aug 02 23:25:48 vladimir minisatip[5755]: Obtained 0 stack frames.
```

Jalle19:
@catalinii can you check that the Coverity secret is valid in this repo? There are many similar issues reported and they all boil down to wrong/missing token.

Jalle19:
@catalinii 
image
Please check

Jalle19:
I haven't had a crash for days so I changed the scope of this to include debug builds as release artifacts so people who want/need to debug something can easily pick the same version but with debug symbols.

Jalle19:
```
Aug 06 20:55:05 vladimir minisatip[122000]: RECEIVED SIGNAL 11 (1) - SP=0 IP=0
Aug 06 20:55:05 vladimir minisatip[122000]: Obtained 0 stack frames.
Aug 06 20:55:05 vladimir systemd[1]: minisatip.service: Main process exited, code=exited, status=1/FAILURE
```

Seems nothing except actually using `make debug` will produce a stack trace

Jalle19:
Debug builds fail with GLIBC 2.38 something something, because we can't build statically with the debug options...

Issue #1274, opened 2025-08-04
update_pids() race condition leading to endless "pids already associated with pmt X"


Jalle19:
[update_pids_bug.txt](https://github.com/user-attachments/files/21579802/update_pids_bug.txt)

Stumbled on a tricky bug last night that caused DDCI to stop working until restarting minisatip.

The problem seems to originate with:

```
Aug 03 22:32:55 vladimir minisatip[27683]: [03/08 22:32:55.031 main]: ad 14 [14], another update_pids is in progress skipping
```

This leads to a ton of this:

```
Aug 03 22:32:55 vladimir minisatip[27683]: [03/08 22:32:55.348 AD5]: del_pmt_mapping_table: ad 5 pid 1025 ddci_pid 1025, pid also associated with pmt 279
Aug 03 22:32:55 vladimir minisatip[27683]: [03/08 22:32:55.348 AD5]: del_pmt_mapping_table: ad 5 pid 1023 ddci_pid 1023, pid also associated with pmt 279
Aug 03 22:32:55 vladimir minisatip[27683]: [03/08 22:32:55.348 AD5]: del_pmt_mapping_table: ad 5 pid 1028 ddci_pid 1028, pid also associated with pmt 279
Aug 03 22:32:55 vladimir minisatip[27683]: [03/08 22:32:55.348 AD5]: del_pmt_mapping_table: ad 5 pid 1027 ddci_pid 1027, pid also associated with pmt 279
```

I'm not sure where to start looking, @catalinii any ideas?

Jalle19:
A quick glance shows that `update_pids()` is called without adapter mutex lock here: https://github.com/catalinii/minisatip/blob/b3831a79bba864d815c7cd49cba7c1bbd4c9d29b/src/ddci.cpp#L488, looks suspicious to me

Jalle19:
Since updating to 2.0 I have another issue where free-to-air channels are sometimes sent via the CAM, not sure why. The logs indiciate that the services have a CA PID which is absolutely not the case. Some race condition probably.

Jalle19:
I have a feeling this got fixed recently by either https://github.com/catalinii/minisatip/pull/1285 or https://github.com/catalinii/minisatip/pull/1275, will close this for now

Issue #1264, opened 2025-07-28
Read overflow in EVP_DecryptUpdate()


Jalle19:
Happens after 10-300 minutes of runtime while using CI.

```
Jul 28 00:00:22 vladimir minisatip[260348]: =================================================================
Jul 28 00:00:22 vladimir minisatip[260348]: ==260348==ERROR: AddressSanitizer: dynamic-stack-buffer-overflow on address 0x7f945a4e3ee2 at pc 0x55869dd49fdb bp 0x7f945a4e3700 sp 0x7f94>
Jul 28 00:00:22 vladimir minisatip[260348]: READ of size 16 at 0x7f945a4e3ee2 thread T17
Jul 28 00:00:23 vladimir minisatip[260348]:     #0 0x55869dd49fda in memcpy (/usr/local/bin/minisatip+0x318fda)
Jul 28 00:00:23 vladimir minisatip[260348]:     #1 0x7f9464bdd721  (/lib/x86_64-linux-gnu/libcrypto.so.3+0x33d721)
Jul 28 00:00:23 vladimir minisatip[260348]:     #2 0x7f9464bdc5ee  (/lib/x86_64-linux-gnu/libcrypto.so.3+0x33c5ee)
Jul 28 00:00:23 vladimir minisatip[260348]:     #3 0x7f9464aa85e5 in EVP_DecryptUpdate (/lib/x86_64-linux-gnu/libcrypto.so.3+0x2085e5)
Jul 28 00:00:23 vladimir minisatip[260348]:     #4 0x55869df484eb in decrypt /home/negge/minisatip/src/aes.c:104
Jul 28 00:00:23 vladimir minisatip[260348]:     #5 0x55869df48d51 in dvbaes_cbc_decrypt_stream /home/negge/minisatip/src/aes.c:126
Jul 28 00:00:23 vladimir minisatip[260348]:     #6 0x55869df6460c in cw_decrypt_stream /home/negge/minisatip/src/pmt.c:659
Jul 28 00:00:23 vladimir minisatip[260348]:     #7 0x55869df65310 in test_decrypt_packet /home/negge/minisatip/src/pmt.c:678
Jul 28 00:00:23 vladimir minisatip[260348]:     #8 0x55869df67bf0 in update_cw /home/negge/minisatip/src/pmt.c:744
Jul 28 00:00:23 vladimir minisatip[260348]:     #9 0x55869df7026c in decrypt_batch /home/negge/minisatip/src/pmt.c:939
[...]

PR #1275, opened 2025-08-04
Fix various race conditions

Jalle19:
May be related to https://github.com/catalinii/minisatip/issues/1264 and https://github.com/catalinii/minisatip/issues/1274


Jalle19:
Testing these commits locally on top of #1271

PR #1278, opened 2025-08-05
Add Ubuntu 25.04 to the build matrix (for evaluating GCC 15)

Jalle19:


Jalle19:
Herp, neither Ubuntu 25.04 nor Debian Trixie ships with GCC 15, let's revisit this at a later stage

catalinii:
25.10 then? There may be a docker image

PR #1283, opened 2025-08-07
Netceiver fixes

pbiering:
related to https://github.com/catalinii/minisatip/discussions/1282

### DIFF
--- a/src/netceiver.cpp
+++ b/src/netceiver.cpp
@@ -150,7 +150,11 @@ int netcv_commit(adapter *ad) {
         if (!SN->ncv_rec)
             SN->err = 1;
 
-        // SN->want_tune = 0; // wait until netcv_tune triggers the tuning
+        /* looks like worked in the past but found need to be
+         * disabled 2025-08-08 - in case issues will be reported this needs to
+         * be toggled by option to cover both use cases
+        SN->want_tune = 0; // wait until netcv_tune triggers the tuning
+         */
     }
 
     /* tune receiver to a new frequency / tranponder */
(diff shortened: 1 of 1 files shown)


M-Reimer:
Probably it would be a good idea to also make minisatip detect libnetceiver without the "legacy library symlink":
https://github.com/vdr-projects/libnetceiver/blob/master/lib/Makefile#L179-L183
This is only there to make minisatip build without modification. All other projects already build with the new library name.

src/netceiver.cpp
```diff
@@ -295,6 +297,10 @@ int netcv_commit(adapter *ad) {
                 tp->tmode, tp->gi, tp->bw, tp->sm, tp->t2id);
 
             break;
+
+	default:
```
Jalle19: Can you run clang-format on this thing?

src/netceiver.cpp
```diff
@@ -149,12 +150,13 @@ int netcv_commit(adapter *ad) {
         if (!SN->ncv_rec)
             SN->err = 1;
 
-        SN->want_tune = 0; // wait until netcv_tune triggers the tuning
+        // SN->want_tune = 0; // wait until netcv_tune triggers the tuning
```
Jalle19: Remove the line completely if this is the correct behavior

src/netceiver.cpp
pbiering: tab replaced by spaces by 3ac4c8e29b2d5de96294f96d387f468441fe3da4

src/netceiver.cpp
pbiering: I added a surrounding comment now, can be that later it need to be toggled, let see once more feedback is received.
3013b53b91e8348e7576dfa08e8d31182c425868

Jalle19: Okay, that's fine

Jalle19 (approved):

PR #1287, opened 2025-08-08
Try to fix wrong tag being added when backporting fixes

Jalle19:


Jalle19:
> Bumping tag v1.3.55 - New tag v1.3.56

PR #1301, opened 2025-08-21
Remove locking and allocation wrappers to be able to call new on creating objects

catalinii:
- in C++ new needs to be called instead of malloc to be able to use C++ types.
- remove PMT, SKey, SCW, SFilter mutex and leave just global mutex (for each individual struct) to make locking easier
- Remove almost all mutex_lock/unlock, technically they can be replaced with unique_lock

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -264,11 +264,8 @@ int add_filter_mask(int aid, int pid, void *callback, void *opaque, int flags,
         LOG_AND_RETURN(-1, "%s failed", __FUNCTION__);
     if (!filters[fid])
         filters[fid] = new SFilter();
-    if (!filters[fid])
-        LOG_AND_RETURN(-1, "%s failed for id %d", __FUNCTION__, fid);
 
     f = filters[fid];
-    std::lock_guard<SMutex> lock2(f->mutex);
     f->id = fid;
     f->opaque = opaque;
     f->pid = pid;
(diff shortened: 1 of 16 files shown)


src/pmt.cpp
```diff
@@ -1402,8 +1408,8 @@ int pmt_del(int id) {
         if (pmts[i] && pmts[i]->enabled)
             break;
     npmts = i + 1;
+    pmt->enabled = 0;
```
Jalle19: This and the `pmt->batch = NULL;` change above seems to be unrelated to the mutex/alloc changes, is it an attempt to fix something?

Jalle19:
The `find_new_id` function is pretty bare-bones now and we could probably replace it with a standard container of some sort at some point.

catalinii:
> The `find_new_id` function is pretty bare-bones now and we could probably replace it with a standard container of some sort at some point.

was thinking about that but re-using the ids seems to be not supported by the container types.
For example std::vector has insert and delete which changes the index of the elements.

Jalle19:
Is there any point in reusing?

catalinii:
The other option would be to switch to use pointers rather than indexes, but this would be a bigger change, however would eliminate all get_xyz from the code.

for example is socket 1 points to stream 1 and sockets 2 points to stream 2, and then delete the stream 1, stream 2 becomes stream 1.

we can also add always streams at the end but then all for loops iterating over streams will have increased number of iterations (+ costs on allocating/freeing memory)

PR #1303, opened 2025-08-29
WIP: Add support for descrambling data streams with constant code-words

### LINKED ISSUE #1223: Error  "op not found for algo 1"
Playing 2 concurrent scrambler channels , only the first channel is working.

> [26/05 15:12:58.375 main]: dvbapi read from socket 8 the following data (21 bytes), pos = 0, op 40106F86, key 0
> [26/05 15:12:58.375 main]: dvbapi: received DVBAPI_CA_SET_DESCR, key 0 parity 0, index 1, CW OK: 01 23 45 69 89 FF DD 65
> [26/05 15:12:58.375 main]: op not found for algo 1
> [26/05 15:12:58.375 main]: dvbapi read from socket 8 the following data (17 bytes), pos = 0, op 400C6F88, key 0
> [26/05 15:12:58.375 main]: Key 0, Algo set to 1, Mode set to 0
> [26/05 15:12:58.375 main]: dvbapi read from socket 8 the following data (21 bytes), pos = 0, op 40106F86, key 0
> [26/05 15:12:58.376 main]: dvbapi: received DVBAPI_CA_SET_DESCR, key 0 parity 1, index 1, CW OK: 01 23 45 69 89 FF DD 65
> [26/05 15:12:58.376 main]: op not found for algo 1

Thanks,

Jalle19:
This set of changes enable successful reception of Abertis services on Hispasat 30W. Essentially each service contains just a PMT PID, a data PID and a CA PID. The data PID is BISS encrypted and contains an MPEG transport stream in its payload (basic encapsulation). Despite what the internet says, this doesn't seem to be T2-MI encapsulated, but instead uses a custom simpler encapsulation.

Once the data PID is decrypted, the output from minisatip can be piped through a simple de-encapsulator and then fed as an IPTV service to tvheadend.

Changes required:
* remove some "is video/audio" checks so we can successfully start a PMT that only contains a data PID. These particular PIDs use type 144 which is completely bogus (Blu-ray subtitling)
* since null packets are used as CA PID we need to discard them in `assemble_packet()`. The PID doens't contain anything required for descrambling, but I guess for technical reasons the service must have a CA PID.
* the stream parity never changes, so the initial constant code word we receive from DVBAPI will expire after some time. Work around this by detecting whether the PMT in question uses constant code-word scrambling and use a much longer life time.

@catalinii ready for review

Fixes https://github.com/catalinii/minisatip/issues/1223

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -312,8 +312,7 @@ int CAPMT_add_PMT(uint8_t *capmt, int len, SPMT *pmt, int cmd_id,
                  pmt->id, pmt->adapter, pmt->stream_pid[i]->pid);
             continue;
         }
-        if (!pmt->stream_pid[i]->is_audio && !pmt->stream_pid[i]->is_video)
-            continue;
+
         capmt[pos++] = pmt->stream_pid[i]->type;
         copy16(capmt, pos, pmt->stream_pid[i]->pid);
         pos += 2;
--- a/src/dvbapi.cpp
+++ b/src/dvbapi.cpp
@@ -332,6 +332,9 @@ int dvbapi_reply(sockets *s) {
         case DVBAPI_ECM_INFO: {
             int pos1 = s->rlen - pos;
             SKey *k = get_key(b[4] - opts.dvbapi_offset);
+            if (!k) {
+                break;
+            }
             unsigned char cardsystem[255];
             unsigned char reader[255];
             unsigned char from[255];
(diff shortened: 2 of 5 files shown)


Jalle19:
Need a way to indicate which CAIDs use constant keys so we can set a longer lifetime for those

Jalle19:
Waiting for a data cable for a spare DDCI to arrive so I can test these and some other changes propely before merging this. I may split out the genuine fixes from here to a separate PR.

src/ca.cpp
```diff
@@ -312,8 +312,7 @@ int CAPMT_add_PMT(uint8_t *capmt, int len, SPMT *pmt, int cmd_id,
                  pmt->id, pmt->adapter, pmt->stream_pid[i]->pid);
             continue;
         }
-        if (!pmt->stream_pid[i]->is_audio && !pmt->stream_pid[i]->is_video)
-            continue;
+
```
catalinii, line 315: can u comment on why this is needed exactly ? Maybe add some comments around it...
This keeps the CAPMT small for bundling multiple PMTs into 1

Jalle19: In my particular case I needed to have a particular PID that is neither video nor audio sent to the CAM. I did it this way to avoid having to use arbitrary logic for determining whether the PID should be included or not.

I tried a different approach where we mark each stream PID as scrambled based on whether it has a 0x09 descriptor, but it turned out not to work correctly because those particular descriptors get stripped when the PMT is processed on the CA device adapter. Not sure if that's a bug or what's causing it though.

Jalle19: Figured out why that's the case! We don't store CA descriptors so when we reconstruct the PMT in the DDCI device that descriptor gets lost.

I'll fix that and move all these PMT-related changes to a separate PR that we can get in before we do this one. I'll also factor out the DVBAPI fixes from here.

PR #1311, opened 2025-09-17
Prevent new objects being created while shutting down

catalinii:


Jalle19:
Seems okay but I'm wondering if it's really necessary?

catalinii:
Got a deadlock while a stream was added during shut down. The stream was likely destroyed at that time and the lock operation blocked the thread

PR #1315, opened 2025-09-19
Skip only new fake CW

lars18th:
When a new CW is received with parity in use, it should not be penalized when searching for a new CW. This solves the cases where a new CW arrives after the parity change.

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -861,9 +861,13 @@ int send_cw(int pmt_id, int algo, int parity, uint8_t *cw, uint8_t *iv,
         c->expiry = c->time + expiry * 1000;
 
     if (parity == pmt->parity && pmt->cw && pmt->last_update_cw > 0) {
-        LOG("CW %d for PMT %d (%s) Warning! New CW using the current parity",
-            c->id, pmt_id, pmt->name);
-        c->time = pmt->cw->time - 1000; // We set the time before the active CW
+        int res = 0;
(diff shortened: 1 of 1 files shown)


Jalle19:
What exactly does this fix?

lars18th:
Hi @Jalle19 ,

> What exactly does this fix?

I've added time ago this code in #1055. However, now I've discovered one case not coverred by my original code:
- When a _new_ KEY is received _after_ the parity change.

In this is particular case without this patch the new key will never be checked because the _ancient_ key with the same parity will have greater pritority. This new PR solves this case.

Please, merge it!

Jalle19:
I guess I just don't understand the comment about a "fake" one?

lars18th:
> I guess I just don't understand the comment about a "fake" one?

That's the use case that my original code wants to solve:
- Imagine that you're decoding with a valid CW. And during the time that the parity of this CW is active you receive a new CW for the _same_ parity. This breaks the standards, so the _new_ CW is fake. That's the reason for the comment in the LOG and to penalize that key.
- But the _updated_ code targets one specific case: when the _real_ CW arrives _after_ the parity change. In this case, and _only_ in this case the new received key needs to be trusted (almost in the sense of not mark it as fake).

The code of this PR is tested for a long time too. So ready to merge.
You agree?

Jalle19:
I don't really understand who's sending fake code words, but okay

PR #1316, opened 2025-09-19
Don't reject CWs with a failed check

lars18th:
When searching for a valid CW if you don't have any other alternative then accept any.

This makes sense when decoding with one CA system that can't be checked using the standard method: initial PUSI packet with a bitstream content of 0x0 0x0 0x1. Sometimes the check fails, but the CW is valid. Therefore, the check will be used now to distinguish between more than one CW. If you have several and one of them passes the check, that's the correct one. But don't assume that one that fails is invalid.

This fixes the decoding with some recent CA systems.

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -738,7 +738,7 @@ void update_cw(SPMT *pmt) {
                 break;
             }
             // if we can verify if the CW is return the latest CW
-            if (len)
+            if (len && cw) // but don't reject failed decrypt checks
                 continue;
 
             int change = 0;


Jalle19:
Some tests would be nice, a bit hard to follow the logic.

catalinii:
Do you have logs for when the check fails but the CW is valid?

lars18th:
> Do you have logs for when the check fails but the CW is valid?

Hi @catalinii ,

I don't want to share any log with a CA cryto. 😜 

But it's not necessary at all! Please consider these facts:
- When a check fails you can't be 100% sure that a CW is bad because the CA system could implement some mesures to send a PUSI packet with _fake_ data (that's a packet without starting with valid ES data).
- What you can be sure of is that if a CW passes the check, then the CW is fine, but not the opposite.
- Therefore, if you get _only_ one CW you can only: use it to decrypt or not decrypt at all.  And in _this_ case if the CW don't pass the check the best option it's to assume the same as if the check can't be done. That's to select this CW and retry in the future (re-check). This solves the problem of countermeasures.
- And if you receive _more_ than one CW you can use the check the select the correct one. And if _all_ of them fail, we can try again in the future and use the last one until that time.

So, based on all possible cases, the best available strategy is now used... which consists of using verification _only_ to ACCEPT a key, but _not_ to REJECT it.

Please, merge it as this code is running ok for several months.

lars18th:
> As long as it works for standard use cases it's fine by me

For sure! It only _improves_ a bit the functionality.

So, ready to merge?

lars18th:
Thank you @Jalle19 !

catalinii:
@lars18th can u clarify how this does not work for you ?

The debugging of CW issues seems to be very difficult after this diff (for example troubleshooting the current ICAM issues).

Also it is unclear if using dvbapi + ddci at the same time works after this diff. @Jalle19  can u test?

Jalle19:
I use DDCI and DVBAPI on separate instances so cannot test unfortunately

lars18th:
> @lars18th can u clarify how this does not work for you ?

Hi @catalinii ,

This patch is to resolve this particular problem: Some systems are sending over some streams _phantom_ packets with PUSI indicator enabled that CAN'T be decrypted with the CURRENT key. This benviour generates a "false negative" with the test that check if the PES data starts with `0x00 0x00 0x01`. Please take into account that you can only be sure about TRUE positives, never about negatives (true or false). Without this patch it's posible that you lost an entire key interval only because you've rejected a _valid_ key. So the patch IS necessary.

> The debugging of CW issues seems to be very difficult after this diff (for example troubleshooting the current ICAM issues).

I don't understand why it can dificult that. But in any case you can add one user parameter to enable/disable this if you be more confortable with it. However, IMHO the default mode needs to be OFF (the _current_ behaviour provided by this patch).

> Also it is unclear if using dvbapi + ddci at the same time works after this diff. @Jalle19 can u test?

In this case, the only potential trouble is that the DDCI module could receive a duplicated ECM. Nothing more. Remember: it only overcomes false negatives!

catalinii:
Before we add another option, do you have a oog with phantom packets?

the whole point of decrypting validation was to not second guess which CW to be used

lars18th:
> Before we add another option, do you have a oog with phantom packets?

Sorry, no. And I'll _never_ share that.

> the whole point of decrypting validation was to not second guess which CW to be used

Yes. But I _repeat_ the same: the CW check is only true to validate the correct key, and _never_ to unvalidate it. That's the point that you need to assume.

As a resume, the four cases related to the check:

- Valid CW && Check OK --> That's the correcty key.
- Valid CW && Check FAIL --> You don't know anything real about this key, so retry to search for the correct key (that could be the same CW if you don't invalidate it). That's the change introduced with this patch.
- Invalid CW && Check OK --> It's impossible for this to happen.
- Invalid CW && Check FAIL --> You don't have the correct key, so you need to continue searching for it.

Jalle19:
So this actually solves a real problem (channel can't be decrypted at all without this fix) and not just a hypothetical?

lars18th:
If you read the initial post: _This fixes the decoding with some **recent** CA systems._

Jalle19:
Right

Jalle19:
The issue @catalinii was debugging is solved so I guess we can leave this as is.

catalinii:
I am trying to focus here on the valid CW and failed check scenario, and my ask here is not just debugging related.

There are for example providers on Astra that send the new CW for odd for example, while the odd key is in use. If you start the channel at the end of the current CW is hard to figure out which CW is active now without actually validating.

If you look at the entire code here https://github.com/catalinii/minisatip/pull/1316/files#diff-d62717f6a7119c9045dab91d7f9629d8848c049e7ff89b0350d5079d6a498990L741 before the change, the logic was:
If we have a packet with PUSI:
 1) use a validated CW if it exists
 2) if not, select the latest CW for that PMT
if no packets with PUSI exist, force a recheck of the CW at the next batch of packets

now the logic is:
If we have a packet with PUSI
1) use a validated CW if it exists
2) use the first CW from the list (which can go quite wrong for CI+ where multiple active CWs are within expiry period at the same time), or the scenario I've explained.

Probably what happens for your use case is that you have 1 CW that is not expired at one time so it is a lucky case that you are getting a CW.

If the phantom packets are the issue here, you can set pmt->update_cw = 1 if the validation of CW fails and leave the logic as it is. The next batch will have most likely few instances of packets with PUSI and then you can validate the CW easily. 

if you do not want to provide logs, can u highlight what goes wrong before the patch ?
The goal here is not to cover the issue but to make it work right as other people will see it too.

lars18th:
Hi @catalinii ,

I'm not sure about the problem that you describe. This patch _only_ changes this: if a KEY is _NOT_ valited, then we can't invalidate it forever because a false negative can be true. There are many different reasons why the validation process may not be completed: the key is wrong (or fake), no PUSI header, PUSI countermesures, etc. In any case the current test can't prove that a key is incorrect, it can only prove that is valid. And nothing else.

So, if your problem is with a CI+ module with multiple active CWs, then the problem is not with this patch... the problem is how you're handling this specific scenario. Perhaps the current behaviour of "expired" keys needs to be revised to handle CI+ modules with multi-channel decoding.

catalinii:
There are 2 issues here:
1) allowing invalided CWs
2) which CW we pick if the CW is not validated

this patch solves 1) which is known to be a problem for your use case.
This patch breaks 2) because it picks the first CW that is not expired (not the last CW as before).

I will add a command line option to allow a flag (by default disabled) that allows picking a CW when none is validated and when someone has the same issue and can provide logs we can fix the behavior.

lars18th:
Hi @catalinii ,

> This patch breaks 2) because it picks the first CW that is not expired (not the last CW as before).

Why you say that? With this patch the _current_ (or _last_) CW continues in use if you can't validate the CW.
Take note about this: if a new CW appears and you can't validate it... this doesn't breaks anything in the _current_ CW interval, because the new appeared CW is for _other_ interval (even/odd). I can't understand the use case that the current implementation with this patch can't work.

> I will add a command line option to allow a flag (by default disabled) that allows picking a CW when none is validated and when someone has the same issue and can provide logs we can fix the behavior.

My suggestions is that the _current_ behaviour with this patch will be the default mode.

lars18th:
Hi @catalinii ,

After more tests, perhaps we need to rollback this PR or make it optional for specific CAIDs. The problem seems to be correlated to some _delay_ in the reception of the KEY when the CW parity is changed. So the question is: you can implement a _buffer_ to delay the decryption?

Jalle19:
Feel free to revert until we have a better solution 👍

lars18th:
> Feel free to revert until we have a better solution 👍

Hi @Jalle19 . Temporally disabled. Please merge it until an update is carried out.

PR #1317, opened 2025-09-23
Add PMT pid too when emulating pids all

lars18th:
Until now, the PMT pid was missing from the emulation of the pids all. This fix resolves this bug by adding the PMT pid first, before the other pids from the service.

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -1148,6 +1148,11 @@ void emulate_add_all_pids(adapter *ad) {
                 if (!pmt)
                     continue;
 
+                LOG("%s: adding PMT pid %d to emulate all pids",
+                    __FUNCTION__, pmt->pid);
+                mark_pid_add(p_all->sid[i], ad->id, pmt->pid);
+                updated = 1;
+
                 for (k = 0; k < pmt->stream_pids; k++)
                     if (!pids[pmt->stream_pid[k]->pid]) {
                         LOG("%s: adding pid %d to emulate all pids",
@@ -1155,7 +1160,6 @@ void emulate_add_all_pids(adapter *ad) {
                         mark_pid_add(p_all->sid[i], ad->id,
                                      pmt->stream_pid[k]->pid);
                         pids[pmt->stream_pid[k]->pid] = 1;
-                        updated = 1;
                     }
             }
 


src/pmt.cpp
```diff
@@ -1148,14 +1148,18 @@ void emulate_add_all_pids(adapter *ad) {
                 if (!pmt)
                     continue;
 
+                LOG("%s: adding PMT pid %d to emulate all pids",
+                    __FUNCTION__, pmt->pid);
+                mark_pid_add(p_all->sid[i], ad->id, pmt->pid);
+                updated = 1;
```
Jalle19, line 1154: Now this is called unconditionally, maybe we should check for `if (!pids[pmt->pid])` first?

Jalle19: Same thing below when it adds the enforced PIDs - it should check first whether they're added.

Jalle19: `mark_pid_add` already checks for existing PIDs, but checking here at call sites would prevent the logging from happening

lars18th: Hi @Jalle19 ,

Plase, don't mix the _optimization_ of this function (described in #1318) with the _bug_ of the missing PMT pid solved here.

This PR must now be merged. After that, we can start looking for solutions for the un-optimized code.
You agree?

Jalle19: I don't, my comments are are about your currently proposed change.

Jalle19: You need to change this so that `updated` is not always set to 1, since then `update_pids()` will be called every time.

src/pmt.cpp
```diff
@@ -1148,14 +1148,18 @@ void emulate_add_all_pids(adapter *ad) {
...
+
                 for (k = 0; k < pmt->stream_pids; k++)
                     if (!pids[pmt->stream_pid[k]->pid]) {
                         LOG("%s: adding pid %d to emulate all pids",
                             __FUNCTION__, pmt->stream_pid[k]->pid)
                         mark_pid_add(p_all->sid[i], ad->id,
                                      pmt->stream_pid[k]->pid);
                         pids[pmt->stream_pid[k]->pid] = 1;
-                        updated = 1;
```
Jalle19, line 1158: This needs to be kept too

src/pmt.cpp
lars18th: Sorry?

`updated` is a local var that is used to trigger the call to `update_pids(ad->id);` at the end. And when you add the PMT pid you're adding a new pid! Therefore the `updated = 1;` is needed.

In the _current_ code the full list of pids for each service is processed to add the pids... except the PMT pid because this pid is not in the list. For this reason when you process the list of a service first your need to add the PMT pid. Remember: any service has a PMT. No Service implies no PMT. That's the bug that this PR solves.

src/pmt.cpp
lars18th: No, no! Before processing in the loop the list of pids of a service you're adding the PMT pid. That's done before, and it flags the `updated = 1`.

lars18th:
Hi @Jalle19 ,

This code is _fully_ tested and ready to merge. 😉

src/pmt.cpp
Jalle19: You need to add `if (!pids[pmt->pid]) {` around the call to `mark_pid_add()` and `updated = 1`.

Jalle19:
I think the reason you think this function consumes too much CPU is due to the changes I've suggested and you're unwilling to make.

src/pmt.cpp
Jalle19: Nevermind, I mistakenly thought `pids` was a global thing, but it's local so it doesn't matter.

Jalle19:
Sorry for the confusion, I mistakenly thought the `pids` array we're checking when we add stream PIDs was global or at least persistent, but it's local so we don't need to check if the PMT PID exists in it already (it won't).

Issue #793, opened 2021-01-12
Request: "pids=all" enhancement 


lars18th:
Hi @catalinii ,

This request is mainly for Enigma2 STBs that doesn't support (directly) the "pids=all" parameter. As the support in this case is based on a fixed list of pids (PSI/SI tables) + PMTs + pids_of_each_service; I want to request a "simple" enhancement:

- The idea is to support a "new" non-standard version of the "pids=all" parameter. To maintain the compatibility the extension should have this syntax: `pids=all&no-av-pids-all=1`. Here this new "no-av-pids-all" parameter indicates to not include in the list of the pids any of the video or audio streams.

As you can imagine, this has only sense in Enigma based boxes, as in this case the code parses all PMT tables do identify the pids. So in this case, the code only needs to check the parameter and "not add" a pid from the PMT if the stream is identified as audio or video.

What you think about this idea?

lars18th:
Hi @catalinii ,

What you think about this idea? It has sense?

catalinii:
Hey @lars18th ,

I am struggling to understand how is this useful?

Can you explain more?

Also which clients will use this feature?

lars18th:
> I am struggling to understand how is this useful?

When you want to receive the full transport stream, that's all pids, but the hardware doesn't support huge bitstreams.

> Can you explain more?

It's simple: You want to parse all data, or do a scaning, or whatever, but without "filtering" control (so using a simple HTTP request). In this case, you want _all_ pids, but the audio and video aren't needed.

> Also which clients will use this feature?

That's only for user custom scripts. I repeat: You _only_ do **one** HTTP request, so it's not possible to add pids after parsing the PAT table. The, a "pids=all without A/V" will do what you want. And the current case is that the code to "add" pids is included. So the idea is a simple if to not add a pid if it's audio or video. Nothing more.

You think you can implement this, please?

xxj737:
Hi!
I have a ZGEMMA H7S (openatv 7.3 Enigma2)   running Minisatip standalone.
for Enigma2 STBs that doesn't support (directly) the "pids=all" parameter

PR #1322, opened 2025-09-30
Emulate pids all locally when using satipc

lars18th:
When using the satipc module with "no_pids_all" enabled, it has sense that with "emulate_pids_all" enabled the pids emulation processing will be done locally. So instead of sending "pids=all" the request contains the full pids list.

This patch fixes this issue. Therefore, you don't need that the target SAT>IP box have real "pids=all" support. Using `-k -s ~satip_server` will be sufficient.

### DIFF
--- a/src/satipc.cpp
+++ b/src/satipc.cpp
@@ -1338,7 +1338,7 @@ void satipc_get_pids(adapter *ad, satipc *sip, char *url, int size,
     if (!sip->lap && !sip->ldp)
         send_pids = 1;
 
-    get_adapter_pids(ad->id, tmp_url, sizeof(tmp_url));
+    get_adapter_pids(ad->id, tmp_url, sizeof(tmp_url), opts.emulate_pids_all);
     if ((!strcmp(tmp_url, "all") || !strcmp(tmp_url, "none"))) {
         send_pids = 1;
     }
--- a/src/adapter.cpp
+++ b/src/adapter.cpp
@@ -2074,7 +2074,7 @@ int signal_thread(sockets *s __attribute__((unused))) {
     return 0;
 }
 
-char *get_adapter_pids(int aid, char *dest, int max_size) {
+char *get_adapter_pids(int aid, char *dest, int max_size, int emu_all) {
     int len = 0;
     int i;
     adapter *ad = get_adapter_nw(aid);
@@ -2086,9 +2086,13 @@ char *get_adapter_pids(int aid, char *dest, int max_size) {
     for (i = 0; i < MAX_PIDS; i++)
         if (ad->pids[i].flags == 1 || ad->pids[i].flags == 2) {
             int pid = ad->pids[i].pid;
-            if (pid == 8192) {
-                strlcatf(dest, max_size, len, "all,");
-                break;
+			if (pid == 8192) {
+                if (!emu_all) {
+                    strlcatf(dest, max_size, len, "all,");
+                    break;
+                } else
+                    strlcatf(dest, max_size, len, "0,1,");
+                    continue;
             } else
                 strlcatf(dest, max_size, len, "%d,", pid);
         }
--- a/src/adapter.h
+++ b/src/adapter.h
@@ -174,7 +174,7 @@ int get_lnb_int_freq(transponder *tp, diseqc *diseqc_param);
 int delsys_match(adapter *ad, int del_sys);
 int get_enabled_pids(adapter *ad, int *pids, int lpids);
 int get_all_pids(adapter *ad, int *pids, int lpids);
-char *get_adapter_pids(int aid, char *dest, int max_size);
+char *get_adapter_pids(int aid, char *dest, int max_size, int emu_all);
 int adapter_timeout(sockets *s);
 void adapter_set_dvr(adapter *ad);
 char is_adapter_disabled(int i);


lars18th:
Please do not merge yet. I'll improve it a bit.

Jalle19:
Not sure if it's worth emulating this on both ends, couldn't you just run minisatip on the target device and use `-k` there? Then the client (minisatip) could send pids=all normally.

lars18th:
Hi @Jalle19 ,

Don't worry about that! My next improvement is to use `-k` for local pids emulation and `-s ~` for remote SAT>IP servers. But in both cases it will be independent. Te reason is because it not have sense to enforce both and also when using `-s ~` we need to do _real_ emulation (not like now that be do a fake emulation). Using this behaviour the user could select to use local emulation or send `pids=all` and the emulation will be done remotely (if the remote server needs it).

Regards.

Jalle19:
I don't understand why any of this is necessary? If you have a device that doesn't support pids=all, run minisatip on it with `-k`. What else is needed?

lars18th:
> I don't understand why any of this is necessary? If you have a device that doesn't support pids=all, run minisatip on it with `-k`. What else is needed?

🤣 🤣 🤣 🤣 🤣 

I'm using minisatip on one X86 powerful server receiving the streams from several E2 boxes running minisatip too. However, when emulating pids all... why want you to execute it on the low end hardware tuners instead of a powerful CPU in the central minisatip instance?

Seriously, when using the `satipc` module you don't need to use only _good_ SAT>IP tuners. And emulate the `pids=all` feature could be interesting _if the user wants to use it_. By default this is not enabled and it's only an enhanced feature. Futhermore, in the future I'll reuse it to support someting requested years ago: #793.

Jalle19:
I have yet to see any evidence whatsoever that emulating pids all is an expensive operation that your average Enigma box cannot handle on its own.

Jalle19:
In general it's a good idea to keep hacks to a minimum, and emulating pids=all is a hack

lars18th:
> In general it's a good idea to keep hacks to a minimum, and emulating pids=all is a hack

The emulation of pids=all it's **a feature**. It's a hack in Enigma2 boxes when it's enabled by default.

Please note that “minisatip” is a powerful SAT>IP server for professionals and researchers. Any “optional” feature can be very useful for engineers. The only important thing is to not activate features by default that may be annoying to a regular user.

Jalle19:
I'm thinking more about the maintainability of the code

Jalle19:
Professionals don't use hacky hardware, hobbyists do.

lars18th:
> Professionals don't use hacky hardware, hobbyists do.

Why do you think that emulating pids=all is related to hacky hardware? This patch is to provide a _functionality_.  It's true that originally it was evolved from a limitation of some E2 boxes. But I'm a professional and engineer, and I use SAT>IP for several use cases. So, please stop to think that minisatip is only a simple server inside a small STB. When using the `satipc` module it evolves to a full _central server_ processing streams from multiple tuners. And this PR (when enhanced) will be the base to provide the implementation of the idea described in #793.

So please wait until the patch is complete.

Jalle19:
I don't think #793 is a good idea either. It's non-standard and none of this is required if you use proper DVB hardware. The limitation of E2 boxes is the original and only remaining reason for this functionality, and thus it should be limited in scope and only used on such devices.

Jalle19:
Do you have any examples of hardware that can't do pids=all, except for Enigma boxes?

lars18th:
> I don't think #793 is a good idea either. It's non-standard and none of this is required if you use proper DVB hardware. The limitation of E2 boxes is the original and only remaining reason for this functionality, and thus it should be limited in scope and only used on such devices.

My idea is to implement this _new_ functionality `pids=all,nav` (or `allnov` I'm not sure) that indicates: send me all pids but block every Audio/Video packet.

And this will not breaks anything. It only adds a _new_ word (`nav`) that has this sense: enable pids all emulation and don't pass any audio or video pid. Why this? Imagine that you want to receive _ALL_ other pids, mainly tables plus something else. This reduces the bitrate to a minium (around 1~500Kbps).

This will have no impact on you (an advanced user). And for me and other engineers, it will be a very valuable feature.

Anyway, please wait until I complete this PR.

Jalle19:
How exactly is it valuable? What does any of this actually fix or enable?

lars18th:
Hi @Jalle19 ,

One year after, I withdraw the idea of implementing #793. You're right that is quite complex and low useful.

However, this PR fixes a real problem (and doesn't hurt any functionality):
- If the user enables (because hi wants it and not because it's necessary) the local pids-all emulation (with `-k`) and it uses the `satipc` module, without this patch the request includes `&pids=all,0,1,16,...`. That is the "all" option and the full list of pids. Instead of this, it has sense to send only the full list of pids. Furthermore, the problem is not that this request could be considered erroneous, the problem is that you're receiving the full TS to send only part of it filtered. Then it has sense to pass the list of pids, like is done with a local adapter, and reduce the necessary bandwith.

So please, review and merge this PR.
Thank you.

lars18th:
Hi @catalinii ,

Please could you review this _simple_ PR?

catalinii:
@lars18th can we close this PR and open a new one with a proper description?

my question is more around why doing this in get_adapter_pids after that

lars18th:
Hi @catalinii ,

Thank you for the response.

> @lars18th can we close this PR and open a new one with a proper description?
> 
> my question is more around why doing this in get_adapter_pids after that

Regarding the place to do the change: the function `get_adapter_pids()` is used _only_ by the `satipc` module. And it requests the internal list of pids from the adapter to use this list with the remote SAT>IP server. The problem, is that when you enable all-pids-emulation, you need to do this locally. And the code really does it. However, the current implementation is doing the emulation, but adding at the start one `pids=all,...`. This is received by the target SAT>IP server, that process `all` and discards the rest of the pids. However, in minisatip when also doing emulation of pids-all on the remote, then this `pids=all` received triggers another time the search of the pids. So in fact, a waste of resources and one behavior not expected. The correct way is: if then pids-all is enabled then emulate it _locally_, and then you only send a pid list... like you do with local tuners.

So, IMHO the description is correct. Why reopen another one PR? Or the problem is that you want a "correct" description for the commit?

src/satipc.cpp
```diff
@@ -1537,7 +1338,7 @@ void satipc_get_pids(adapter *ad, satipc *sip, char *url, int size,
     if (!sip->lap && !sip->ldp)
         send_pids = 1;
 
-    get_adapter_pids(ad->id, tmp_url, sizeof(tmp_url));
+    get_adapter_pids(ad->id, tmp_url, sizeof(tmp_url), opts.emulate_pids_all);
     if ((!strcmp(tmp_url, "all") || !strcmp(tmp_url, "none"))) {
```
Jalle19, line 1342: This condition will become false after your additions

Jalle19:
The client (minisatip in this case) should send pids=all regardless of whether itself emulates it - it's up to the server to decide what to do with it. So to me it seems sending "pids=all,1,2,3..n" is a bug, but the fix would be to send just "pids=all".

lars18th:
Hi @Jalle19 ,

> The client (minisatip in this case) should send pids=all regardless of whether itself emulates it [...] but the fix would be to send just "pids=all".

This is absolutely wrong! Emulate pids-all implies that the device is not using "all" or 8912 _at any case_. Instead of it, it requests only a list of pids. And to do that it search inside the PSI tables. This is the behaviour with any internal tuner, and needs to be the behaviour when using `emulate pids=all` and `satipc` at the same time. The current implementation is wrong not because is sending something wrong like `&pids=all,0,1,16,...`, the error is to send the `&pids=all` when using `emulate pids=all`.

src/satipc.cpp
lars18th: Hi @Jalle19 ,

The condition checking requires being here. Why? Because the request to `get_adapter_pids()` could change all or not (the `strlcatf()` is conditional. So the code is right. However I need to check the missing `send_pids = 1;`. I'll check it.

Jalle19:
No, it's a hack for physical adapters that can't pass the full mux.

PR #1327, opened 2025-11-24
Revert "Add missing null check when handling DVBAPI_ECM_INFO"

### LINKED ISSUE #1309: Spurious deadlocks after switching to C++11 mutexes/locks
Hi
Issue almost identical to #1302
I followed advice from there trying to solve it with driver updating but didn't helped.
Since I keep all my git clones archived so that I can go back if something goes wrong, this minisatip15 is OK. 16 and 17 not ok

Image
If I can see exact version in some file from cloned git, let me know.

As I wrote in #1302 OS starts acting weird, it takes alot of time to reboot and after reboot it goes to recovery mode. I just reboot it again and it starts up normally.
It made sense that it's driver issue but as it turns out, it is not.

I just started it with logging on latest version and when it hangs I will attach logs.

Jalle19:
This reverts commit 160cda3732632d3037c4e20e7d8566a60a711fae.

Fixes #1309

### DIFF
--- a/src/dvbapi.cpp
+++ b/src/dvbapi.cpp
@@ -331,9 +331,6 @@ int dvbapi_reply(sockets *s) {
         case DVBAPI_ECM_INFO: {
             int pos1 = s->rlen - pos;
             SKey *k = get_key(b[4] - opts.dvbapi_offset);
-            if (!k) {
-                break;
-            }
             unsigned char cardsystem[255];
             unsigned char reader[255];
             unsigned char from[255];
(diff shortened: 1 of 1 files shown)


Jalle19:
The real issue though is - why are we getting ECM_INFO for a key we no longer have? Why have we deleted the key?

Jalle19:
Another way to fix this would be to keep track of previous position as well as current position within the message parsing loop. If the position hasn't advanced we break out with an appropriate error message. This would simplify handlers like this one and the one for DVBAPI_DMX_SET_FILTER, which is now littered with null checks because we can't afford to incorrectly parse the message or risk not advancing the position correctly.

Alternatively we could parse into temporary structures in DVBAPI_ECM_INFO and copy over to the key only if it's found - that way position would be advanced correctly regardless of whether a key was found or not.

Issue #1323, opened 2025-10-19
no libdvbcsa patch in v2.0.39?


misushot:
with v1.3.57 icam works
https://github.com/catalinii/minisatip/issues/1003

Unfortunately it doesn't work with v2.0.39 anymore

could I compile this myself?

Jalle19:
I don't have a way to verify this, maybe @catalinii can help out?

localhosthack0r:
Hi,
for me it is still running fine (Sky Germany).
Minisatip Version: v2.0.39~856996b

misushot:
x64?

localhosthack0r:
Yes, x86_64.

misushot:
ok i use it as satip client on my x86. it is connected to my dreambox minisatip server. i think here is the failure. and you use physical dvb card

catalinii:
On the minisatip where oscam is connected you need a patched dvbcsa. 

https://github.com/catalinii/libdvbcsa

misushot:
i have compiled https://github.com/catalinii/libdvbcsa then minisatip. no success with minisatip client. with v1.3.57~828235a binary from you no problems,

catalinii:
Can you provide both minisatip and oscam logs?

misushot:
Hi catalinii,

i have send my logs to your @yahoo.com mailaddress.

okmikel:
Same problem, v1.3.57 works and  v2.0.39 won't. Both binaries build myself with the same libdvbcsa. x86_64 on EL10.

Jalle19:
Can you try some older 2.0 versions?

okmikel:
Strange, I tested 2.0.19 and it worked, then I did a fresh git pull and now  v2.0.39 also works. Sorry for the false alarm.

Jalle19:
So is there an actual problem here or not? 😄

misushot:
catalinii have now a remote access to my system and can check this.

catalinii:
@Jalle19  I bisected this to v2.0.28: https://github.com/catalinii/minisatip/pull/1305

catalinii:
I commented on https://github.com/catalinii/minisatip/pull/1305 seems to be related to algo being 3 (as opposed to 0)

Jalle19:
@catalinii any update on this?

misushot:
@Jalle19
#define CA_ALGO_AES128_CBC 3 -> #define CA_ALGO_AES128_CBC 0 ?
[...]

PR #1334, opened 2025-11-26
Fix wrong algorithm used for ICAM ECMs

Jalle19:
Fixes https://github.com/catalinii/minisatip/issues/1323

The problem is once we started correctly parsing the algorithm indicated in DVBAPI_SET_DESCR_MODE responses, we end up getting a value of `3`, which in OSCam is overloaded to mean "CW_ALGO_CSA_ALT": https://git.streamboard.tv/search?search=CW_ALGO_CSA_ALT&nav_source=navbar&project_id=11&group_id=2231&search_code=true&repository_ref=master

We need to ignore this value and treat it as normal CSA for our own purposes.

### DIFF
--- a/src/dvbapi.cpp
+++ b/src/dvbapi.cpp
@@ -678,18 +678,23 @@ int send_ecm(int filter_id, unsigned char *b, int len, void *opaque) {
     return 0;
 }
 
-int set_algo(SKey *k, int algo, int mode) {
-    // Use CBC when mode indicates it
-    if (algo == CW_ALGO_AES128 && mode == CW_ALGO_MODE_CBC)
-        algo = CA_ALGO_AES128_CBC;
-
-    // Treat ICAM ECM the same as normal ECM
-    if (algo == CW_ALGO_CSA_ALT)
-        algo = CA_ALGO_DVBCSA;
-
-    k->algo = algo;
-
-    return 0;
+void set_algo(SKey *k, int algo, int mode) {
+    switch (algo) {
+    case CW_ALGO_CSA:
+    case CW_ALGO_CSA_ALT: {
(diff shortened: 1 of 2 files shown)


src/dvbapi.h
```diff
@@ -24,6 +24,13 @@
 #define DVBAPI_ECM_INFO 0xFFFF0003
 #define CA_SET_DESCR_MODE 0x400c6f88
 
+#define CW_ALGO_CSA 0
```
catalinii, line 27: We should use the same values as oscam otherwise there will be issues with other algos

Jalle19: These are copied from oscam sources

src/dvbapi.cpp
```diff
@@ -678,8 +679,14 @@ int send_ecm(int filter_id, unsigned char *b, int len, void *opaque) {
...
 int set_algo(SKey *k, int algo, int mode) {
-    if (algo == CA_ALGO_AES128 && mode == CA_MODE_CBC)
+    // Use CBC when mode indicates it
+    if (algo == CW_ALGO_AES128 && mode == CW_ALGO_MODE_CBC)
         algo = CA_ALGO_AES128_CBC;
+
+    // Treat ICAM ECM the same as normal ECM
+    if (algo == CW_ALGO_CSA_ALT)
+        algo = CA_ALGO_DVBCSA;
+
```
Jalle19: @catalinii currently we're only changing the value of `algo` here if it differs between the value indicated by DVBAPI and the value we expect in our descramblers. If you want to I can make it explicit, that way we could also show a proper error message if we receive a DES key instead of `op not found for algo 1`.

Jalle19:
@catalinii okay to merge?

Issue #1267, opened 2025-07-28
Expose adapter names in web interface


Jalle19:
```
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.438 signal]: Detected adapter 0 handle 7 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.441 signal]: Detected adapter 1 handle 9 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.446 signal]: Detected adapter 2 handle 11 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.449 signal]: Detected adapter 3 handle 13 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.451 signal]: Detected adapter 4 handle 15 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.454 signal]: Detected adapter 5 handle 17 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.456 signal]: Detected adapter 6 handle 19 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.459 signal]: Detected adapter 7 handle 21 DVB Card Name: CXD2843 DVB-C/C2 DVB-T/T2
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.462 signal]: Detected adapter 8 handle 23 DVB Card Name: STV0367 DVB-C DVB-T
Jul 28 00:00:28 vladimir minisatip[276450]: [28/07 00:00:28.465 signal]: Detected adapter 9 handle 25 DVB Card Name: STV0367 DVB-C DVB-T
[...]

PR #1335, opened 2025-11-29
Expose adapter/frontend name via API, show in web interface

Jalle19:
We've displayed the SAT>IP server address and protocol for SAT>IP devices, but other adapter types have not had any information shown.

This introduces a `name` function for adapters that can be implemented to return a name. DVB adapters use the frontend name, SAT>IP devices use the name we previously constructed in the web interface directly.

`ad_adapter_names` have been added to the API.

Untested on DVB devices still.

Closes https://github.com/catalinii/minisatip/issues/1267

### DIFF
--- a/src/adapter.cpp
+++ b/src/adapter.cpp
@@ -2124,7 +2124,8 @@ char *get_adapter_name(int aid, char *dest, int max_size) {
     if (!ad)
         return dest;
 
-    _strncpy(dest, ad->adapter_name.data(), ad->adapter_name.length() + 1);
+    _strncpy(dest, const_cast<char *>(ad->adapter_name.data()),
+             ad->adapter_name.length() + 1);
 
     return dest;
 }


src/satipc.cpp
```diff
@@ -1557,6 +1566,18 @@ fe_delivery_system_t satipc_delsys(int aid, int fd, fe_delivery_system_t *sys) {
...
 }
 
+std::string satipc_name(int aid, int fd) {
+    satipc *sip = get_configured_satip(aid);
+
+    if (!sip)
+        return "";
+
+    std::ostringstream ss;
+    ss << sip->sip << " @ " << (sip->use_tcp ? "RTP/TCP" : "RTP/UDP");
```
Jalle19, line 1576: Wanted to use `std::format` here but my compiler (g++ 11) doesn't support it.

src/adapter.cpp
```diff
@@ -2214,4 +2226,6 @@ _symbols adapters_sym[] = {
     {"tuner_ac", VAR_INT, &tuner_ac, 1, 0, 0},
     {"tuner_at", VAR_INT, &tuner_at, 1, 0, 0},
     {"ad_disabled", VAR_ARRAY_INT8, disabled, 1, MAX_ADAPTERS, 0},
+    {"ad_adapter_names", VAR_FUNCTION_STRING, (void *)&get_adapter_name, 0,
```
catalinii, line 2229: VAR_STRING does not require a function definition

catalinii: Or if you are using std::string just add a VAR_STD_STRING ?

Jalle19: Yeah I thought about that but took the easy way out 😄 The JSON handler thing is quite hairy already. But it would indeed be a better solution to add something like VAR_STD_STRING.

Jalle19:
@catalinii I can't figure out how to add support for `std::string` to `get_json_state()`. It seems to be so focused around strings being pointers that it's not easy to change. I guess I could use `char *` for `adapter_name` but it would be nice to start using `std::string` more widely.

catalinii:
merge it as it is, maybe we can refactor later the entire get_json_state to use c++ types to prevent statically sized arrays.

Jalle19:
image

Works

Jalle19:
Need to refactor `get_configured_adapter()` so it doesn't log if adapter is NULL, it spams the logs at the moment.

PR #1304, opened 2025-09-02
WIP: Improve how PIDs are selected for descrambling

Jalle19:
Fixes two scenarios:

* a user wants to descramble a PID that is neither video nor audio. This wasn't possible because we skipped processing the PMT completely if it contained no video/audio stream PIDs
* simply removing the video/audio check would potentially cause CA PMT to balloon due to all DVBSUB and TELETEXT PIDs etc. potentially being added to it. This has been improved now with the addition of marking each stream PID as scrambled or not based on whether it has a CA descriptor (`0x09`). Due to a bug this particular descriptor was never stored in the stream PID, so when the PMT was reconstructed on a DDCI device the information was no longer there are filtering out scrambled PIDs would cause all PIDs to be left out.

Before:

```
[02/09 21:04:30.427 AD3]: new PMT 0 AD 3, pid: 05D3 (1491), filter 6, len 106, pi_len 0, ver 1, pcr 1535, sid 4330 (17200)  
[02/09 21:04:30.427 AD3]: PMT pid 1491 - stream pid 05FF (1535), type 27, es_len 11, pos 9, caids 0
[02/09 21:04:30.427 AD3]: PMT 0 PI pos 1 caid 0B00 => pid 1016 (4118), index 0
[02/09 21:04:30.427 AD3]: PMT pid 1491 - stream pid 0D25 (3365), type 3, es_len 20, pos 25, caids 1
[02/09 21:04:30.427 AD3]: PMT pid 1491 - stream pid 0FA9 (4009), type 3, es_len 20, pos 50, caids 1
[02/09 21:04:30.427 AD3]: PMT pid 1491 - stream pid 181F (6175), type 6, es_len 22, pos 75, caids 1
...
[02/09 21:04:30.552 AD0]: new PMT 4 AD 0, pid: 05D3 (1491), filter 9, len 72, pi_len 6, ver 1, pcr 1535, sid 4330 (17200)  
[02/09 21:04:30.552 AD0]: PMT 4 PI pos 1 caid 0B00 => pid 1016 (4118), index 0
[02/09 21:04:30.553 AD0]: PMT pid 1491 - stream pid 05FF (1535), type 27, es_len 5, pos 15, caids 1
[02/09 21:04:30.553 AD0]: PMT pid 1491 - stream pid 0D25 (3365), type 3, es_len 14, pos 25, caids 1
[...]

PR #1339, opened 2025-12-10
Use a vector of structs to store stream PID descriptors

Jalle19:
Previously we stored all descriptors as raw data in a single pointer. This is inflexible since we can't easily access individual descriptors if we want to. We also had to allocate a large chunk of memory for the descriptors to make sure we were able to fit all of them.

I've added tests around the functionality in `ddci_create_pmt()` since we didn't have any.

The next step would be to use the same mechanism for storing PMT descriptors too, instead of what we do now where we parse CA descriptors and store the CAIDs and CA PIDs separately. Since we can now loop through descriptors, we no longer need to store information from them separately. In essence, we'd replace `SPMTCA` with the newly introduced `descriptor_t`.

Prerequisite for https://github.com/catalinii/minisatip/pull/1304

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -1781,10 +1781,8 @@ void pmt_add_descriptor(SPMT *pmt, int stream_id, unsigned char *desc) {
     descriptor_t d = create_descriptor(desc);
 
     // do not add an already existing descriptor
-    if (std::find_if(sp->descriptors.cbegin(), sp->descriptors.cend(),
-                     [&d](const descriptor_t &v) {
-                         return v.type == d.type;
-                     }) != sp->descriptors.cend()) {
+    if (std::find(sp->descriptors.cbegin(), sp->descriptors.cend(), d) !=
+        sp->descriptors.cend()) {
         LOGM("PMT %d pid %d descriptor already added %d", pmt->pid, sp->pid,
              d.type);
         return;
--- a/src/pmt.h
+++ b/src/pmt.h
@@ -98,6 +98,11 @@ typedef struct descriptor {
     uint8_t type;
     uint8_t len;
     std::vector<uint8_t> data;
+
+    bool operator==(const struct descriptor &other) const {
+        return this->type == other.type && this->len == other.len &&
+               this->data == other.data;
+    };
 } descriptor_t;
 
 typedef struct struct_stream_pid {
--- a/tests/test_pmt.cpp
+++ b/tests/test_pmt.cpp
@@ -53,6 +53,9 @@ extern adapter *a[MAX_ADAPTERS];
 extern SFilter *filters[MAX_FILTERS];
 extern SPMT *pmts[MAX_PMT];
 
+// Forward declarations
+descriptor_t create_descriptor(const uint8_t *data);
+
 uint8_t packet[188] = {
     0x47, 0x40, 0xff, 0x99, 0x14, 0x4c, 0x83, 0x7f, 0x46, 0xba, 0xb8, 0x12,
     0xfb, 0x83, 0xf7, 0x50, 0x9c, 0x73, 0x55, 0xe1, 0x8a, 0x1a, 0x54, 0x66,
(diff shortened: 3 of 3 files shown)


src/pmt.cpp
```diff
@@ -1760,25 +1762,34 @@ void pmt_add_caid(SPMT *pmt, uint16_t caid, uint16_t capid, uint8_t *data,
...
-        es_len = sp->desc[i + 1];
-        int desc_id = sp->desc[i];
-        if (desc_id == new_desc_id) {
-            LOGM("PMT %d pid %d descriptor already added %d", pmt->pid, sp->pid,
-                 desc_id);
-            return;
-        }
+    if (std::find_if(sp->descriptors.cbegin(), sp->descriptors.cend(),
+                     [&d](const descriptor_t &v) {
+                         return v.type == d.type;
```
Jalle19: This is taken from the old code, but it's wrong - there can be multiple descriptors with the same type. I think the safest would be to remove the check completely and just rely on not adding the same descriptor twice (should never happen anyway). The alternative is to compare the whole descriptor which seems a bit excessive. @catalinii what do you think?

Jalle19:
Runtime tested for a while with DDCI - works fine. Haven't tested with DVBAPI, wouldn't expect any issues.

Jalle19:
@catalinii do you have a .ts sample that contains CA descriptors that have private data?

src/pmt.cpp
catalinii: process_pmt can be called for the same pmt multiple times especially when the version changes, that’s why the check is here.

So if you want to make it right, then compare the whole descriptor.

catalinii (approved):
Great! Thanks for doing this, this never really worked well in C with static arrays.

Jalle19:
Managed to get a crash, and I don't have stack trace support on my desktop...

Jalle19:
Upgraded my desktop to Ubuntu 24.04, now I get stack traces:

```
[11/12 13:42:27.754 AD4]: SDT PMT 14: name FRANCE 2 provider Eutelsat, sid: 23000 (59D8)
[11/12 13:42:27.754 AD4]: SDT PMT 15: name 3sat HD provider ZDFvision, sid: 23100 (5A3C)
[11/12 13:42:27.754 AD4]: SDT PMT 11: name Viasat History HD provider Telenor, sid: 23200 (5AA0)
[11/12 13:42:27.754 AD4]: SDT PMT 13: name Kunskapskanalen HD provider Sveriges Television, sid: 23300 (5B04)
[11/12 13:42:27.754 AD4]: SDT PMT 12: name FilmBox Arthouse provider A1 Broadcasting, sid: 23400 (5B68)
[11/12 13:42:27.790 AD0]: Stack trace:
   0# print_trace() at /home/negge/Projects/minisatip/src/utils.cpp:230
   1# posix_signal_handler(int, siginfo_t*, ucontext_t*) at /home/negge/Projects/minisatip/src/utils.cpp:258
   2#      at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
   3# __pthread_kill_implementation at ./nptl/pthread_kill.c:44
   4# __GI_raise at ../sysdeps/posix/raise.c:26
   5# __GI_abort at ./stdlib/abort.c:79
   6# __libc_message_impl at ../sysdeps/posix/libc_fatal.c:134
   7# malloc_printerr at ./malloc/malloc.c:5772
   8# _int_free at ./malloc/malloc.c:4541
   9# __GI___libc_free at ./malloc/malloc.c:3398
  10# std::__new_allocator<descriptor>::deallocate(descriptor*, unsigned long) at /usr/include/c++/13/bits/new_allocator.h:172
  11# std::vector<descriptor, std::allocator<descriptor> >::push_back(descriptor const&) at /usr/include/c++/13/bits/stl_vector.h:1292
  12# pmt_add_descriptors(struct_pmt*, int, unsigned char*, int) at /home/negge/Projects/minisatip/src/pmt.cpp:1808
  13# process_pmt(int, unsigned char*, int, void*) at /home/negge/Projects/minisatip/src/pmt.cpp:1978
  14# process_filters(struct_adapter*, unsigned char*, struct_pid*) at /home/negge/Projects/minisatip/src/pmt.cpp:539
  15# pmt_process_stream(struct_adapter*) at /home/negge/Projects/minisatip/src/pmt.cpp:1266
  16# process_dmx(struct_sockets*) at /home/negge/Projects/minisatip/src/stream.cpp:1020
  17# read_dmx(struct_sockets*) at /home/negge/Projects/minisatip/src/stream.cpp:1137
  18# ddci_read_sec_data(struct_sockets*) at /home/negge/Projects/minisatip/src/ddci.cpp:1033
  19# select_and_execute(void*) at /home/negge/Projects/minisatip/src/socketworks.cpp:831
  20# start_thread at ./nptl/pthread_create.c:447
  21# clone3 at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
  22#
  ```

Jalle19:
Fixed the crash, had accidentally removed the calls to set deleted pointers to NULL. Since we check for NULL here and there we ended up reusing deleted memory.

src/pmt.cpp
Jalle19: Implemented `operator==`

Jalle19:
@catalinii I added equality comparison support to `descriptor_t` and switched to using it instead of just checking the type. Would have liked to use a set instead of a vector to get de-duplication for free, but it got complicated with hashing etc.

PR #1340, opened 2025-12-10
Fix log spam after introducing get_configured_adapter()

Jalle19:
Since it's used when looping through all adapters it must not warn about NULL values.


Jalle19:
Would be nice if the build would pass... At some point we may have to stop supporting older distributions.

Jalle19:
Dropped the `<source_location>` stuff for now

PR #1341, opened 2025-12-11
Remove unused PMT "grace time" concept

Jalle19:
The tracking of start time and grace time has ultimately been done in order to call tables_update_encrypted_status(), which calls ca_encrypted() or ca_decrypted() if they're implemented by a registered CA - but nothing implements those functions so it's all dead code.

### DIFF
--- a/src/ddci.cpp
+++ b/src/ddci.cpp
@@ -332,9 +332,7 @@ int ddci_process_pmt(adapter *ad, SPMT *pmt) {
 
     if ((d = get_ddci(ad->id))) {
         LOG("Skip processing pmt for ddci adapter %d", ad->id);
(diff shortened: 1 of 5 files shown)


catalinii:
@Jalle19 I need to bring start_pmt back because there are still references in pmt.cpp

catalinii:
https://github.com/catalinii/minisatip/pull/1351

Jalle19:
Are you sure? VS code found no references and the thing compiles.

catalinii:
Is hidden behind ifdef: https://github.com/catalinii/minisatip/blob/master/src/pmt.cpp#L1213

Jalle19:
Of course

PR #1342, opened 2025-12-18
Store PMT descriptors in the PMT, not the first stream PID

Jalle19:
Follow-up from https://github.com/catalinii/minisatip/pull/1339

PMTs can have descriptors defined in the program info, while stream PIDs have descriptors defined in their elementary stream info. Now we actually keep these separate so we know where they came from.

The long term goal is to not have to call `pmt_add_caid()` but instead just operate on the descriptors we parse. Trying to divide the work into manageable chunks, so I'm putting this up for review now.

The newly introduced methods are almost identical right now, but that will change in the future.

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -1306,6 +1306,7 @@ int pmt_add(int adapter, int sid, int pmt_pid) {
     memset(pmt->name, 0, sizeof(pmt->name));
     memset(pmt->provider, 0, sizeof(pmt->provider));
     pmt->caids = 0;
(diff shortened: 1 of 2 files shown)


catalinii:
LGTM but what is the difference beyween the 2 methods in the implementation

Jalle19:
Nothing at the moment except which vector the descriptors are pushed to. All attempts I've made at refactoring this has required this splitting of the function as the first step, so I figured I'd push it first.

PR #1344, opened 2025-12-20
Use a vector to store stream PIDs

Jalle19:
Also changes `is_audio`/`is_video` to booleans.

The `test_create_capmt()` test case was broken, it overwrote `pmt->stream_pids[0]->type/pid` twice. The test should be expanded to actually test more stuff, mainly the structure of the CA PMT itself. Now it just calls `create_capmt()` and verifies that it returned a length, but the structure itself is untested. Especially going forward if/when we want to treat program-level CA descriptors and stream-level CA descriptors separately we need to have tests for this.

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -305,17 +305,17 @@ void disable_cws_for_all_pmts(ca_device_t *d) {
 
 int CAPMT_add_PMT(uint8_t *capmt, int len, SPMT *pmt, int cmd_id,
                   int added_only, int ca_id) {
-    int i = 0, pos = 0;
-    for (i = 0; i < pmt->stream_pids; i++) {
-        if (added_only && !find_pid(pmt->adapter, pmt->stream_pid[i]->pid)) {
+    int pos = 0;
+    for (const auto &stream_pid : pmt->stream_pids) {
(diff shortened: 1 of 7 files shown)


Jalle19:
Trying to move away from using `pmt->caid` and instead use `pmt->descriptors`, since we store all program-level descriptors now. It gets complicated by the fact that our tests use `pmt_add_caid` directly extensively instead of going through the same parsing logic we have for descriptors at the moment.

The CA PMT we create cheats and moves all CA descriptors to each stream PID. We can do away with this at some point now that we store descriptors in their right locations.

image

Another matter that complicates things is that we can't just copy PMT descriptors verbatim to the DDCI PMT in `ddci_create_pmt()`, since we need to do the PID mapping. This is I guess also why the helper function `pmt_add_ca_descriptor()` is not used there, even though the logic is essentially the same.

Jalle19:
@catalinii can you review?

Jalle19:
I suppose I could split the new additions out to a separate PR.

Jalle19:
Splitting this at some point, no need to review at the moment. Made a simple error that wasn't covered by the tests so I'm expanding them some more.

Jalle19:
@catalinii ready for review

PR #1349, opened 2025-12-27
Rework the stream locking to include also the sockets

catalinii:
In adapter thread, process dmx did not lock sockets which would cause the rsock_id to be polled at the same time in the main thread.
This locks also the rsock for the streams that are being locked and it also reduces the time the lock is held (excluding the stream processing).
Before 2.0.27 both flushing and buffering were locking the stream lock so it was ok.

https://github.com/catalinii/minisatip/blame/37b0ba34eadcd4d50118968dc299c5f78c4f2d1c/src/socketworks.cpp#L1388 was being hit while 1 thread (ADx was pushing data) and the main thread was flushing data at the same time.

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -896,6 +896,8 @@ void check_cc2(adapter *ad) {
     }
 }
 
+// Locks the used structs in order to avoid race conditions (sockets -> stream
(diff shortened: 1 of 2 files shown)


Jalle19 (approved):
If it works 👍

PR #1350, opened 2025-12-27
Minor CA PMT fixes, add more exhaustive tests for structure

Jalle19:
Need to have more tests if we're to confidently refactor the surrounding code

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -318,7 +318,8 @@ int CAPMT_add_PMT(uint8_t *capmt, int len, SPMT *pmt, int cmd_id,
         copy16(capmt, pos, stream_pid.pid);
         pos += 2;
         int pi_len_pos = pos, pi_len = 0;
-        pos += 2;
+        capmt[pos++] = 0; // pi_len
+        capmt[pos++] = 0;
 
         // append the stream descriptors
         if (pmt->caids) {


src/ca.cpp
```diff
@@ -325,8 +325,8 @@ int CAPMT_add_PMT(uint8_t *capmt, int len, SPMT *pmt, int cmd_id,
             capmt[pos++] = cmd_id;
             pi_len = pmt_add_ca_descriptor(pmt, capmt + pos, ca_id);
             pos += pi_len;
+            copy16(capmt, pi_len_pos, pi_len + 1);
```
catalinii, line 329: This is not correct, the descriptor length should always be there: https://github.com/openpli-arm/dvb-apps/blob/17a557bce03c9cf61d0d3b29c4c387b84f95196b/lib/libdvben50221/en50221_app_ca.c#L263

src/ca.cpp
```diff
@@ -336,15 +336,14 @@ int create_capmt(SCAPMT *ca, int listmgmt, uint8_t *capmt, int capmt_len,
...
-    int version = ca->version;
 
     if (!pmt)
         LOG_AND_RETURN(1, "%s: PMT %d not found", __FUNCTION__, ca->pmt_id);
 
     capmt[pos++] = listmgmt;
     copy16(capmt, pos, ca->sid);
     pos += 2;
-    capmt[pos++] = ((version & 0xF) << 1) | 0x1;
+    capmt[pos++] = ca->version;
```
catalinii: The version is also shifted to the left 1 bit: https://github.com/openpli-arm/dvb-apps/blob/17a557bce03c9cf61d0d3b29c4c387b84f95196b/lib/libdvben50221/en50221_app_ca.c#L240

src/ca.cpp
Jalle19: It is always there, the problem this fixes is that the field reports length of 1 when no descriptors are added.

src/ca.cpp
Jalle19: I need to revisit that and check, could be I was looking at the whole byte value only without realizing it containa current/next indicator too.

Jalle19: Fixed

Jalle19:
@catalinii fixed the version handling (changed the test to correctly check it)

src/ca.cpp
catalinii: Maybe we should explicitly set it to 0?

Jalle19: Yeah, we should. It's usually done but not here for some reason.

Jalle19: Explicitly setting length to zero now

Jalle19:
@catalinii is this okay to go in?

catalinii (approved):

PR #1353, opened 2025-12-28
Various tweaks

Jalle19:
Pushing some minor modifications I've racked up locally.

@catalinii what do you think about introducing a new `ddci` log "level" for use in `ddci.cpp`? This way people could make logs with `-l ca` or even `-v ca` without getting a bunch of relatively unrelated DDCI log spam.

### DIFF
--- a/src/ddci.cpp
+++ b/src/ddci.cpp
@@ -878,10 +878,11 @@ int push_ts_to_adapter(ddci_device_t *d, adapter *ad, uint16_t *mapping) {
             d->fifo.write_index - d->read_index[ad->id], d->read_index[ad->id]);
         dump_packets("DDCI -> AD", ad->buf + i, DVB_FRAME, i);
     }
-    LOGM("popped %d bytes from fifo up to index %d, rlen %d, lbuf %d, left in "
-         "fifo %jd",
-         popped, i, ad->rlen, ad->lbuf,
-         d->fifo.write_index - d->read_index[ad->id]);
+    DEBUGM(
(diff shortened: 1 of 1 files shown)


src/pmt.h
```diff
@@ -101,7 +101,15 @@ typedef struct descriptor {
...
-    bool is_ca_descriptor() { return this->type == 0x09; }
+    bool is_ca_descriptor() const { return this->type == 0x09; }
+
+    uint16_t get_ca_descriptor_caid() const {
+        return (data[0] * 256) + data[1];
+    }
+
+    uint16_t get_ca_descriptor_capid() const {
+        return (data[2] & 0x1F) * 256 + data[3];
+    }
```
Jalle19, line 112: Long term I'm thinking of changing these to classes, but for now let's just add the helpers we need here in the header

catalinii:
Yeah, adding a new log module sounds good for me.

tests/test_pmt.cpp
```diff
@@ -109,6 +109,16 @@ int test_descriptor_equality() {
     return 0;
 }
 
+int test_descriptor_caid_capid_getters() {
+    const uint8_t descr1_data[] = {0x09, 0x04, 0x0B, 0x00, 0x05, 0x73};
```
Jalle19, line 113: Added this test before I added the new helper, will change it to use the new one

Jalle19: Actually that helper is defined in another test file so let's not bother for now

PR #1354, opened 2025-12-29
add tests for update_pids and the expected behavior

catalinii:


Jalle19 (approved):
Would be helpful (at least to me) to have constants for the flag values.

catalinii:
@Jalle19 can you double check this diff?

src/adapter.cpp
```diff
@@ -1134,8 +1140,8 @@ int mark_pid_add(int sid, int aid, int _pid) {
     }
     // add the new pid in a new position
     for (i = 0; i < MAX_PIDS; i++)
-        if (ad->pids[i].flags <= 0) {
-            ad->pids[i].flags = 2;
+        if (ad->pids[i].flags == PIDS_FLAGS_DISABLED) {
```
Jalle19: Was <= not possible?

Jalle19:
Much clearer now with constants, thanks! `flags` is a bit of a bad name since it can only have a single value, usage seems to indicate we're talking about "state" instead. Constants could then be renamed accordingly (PID_STATE_DISABLED) etc.

src/adapter.cpp
catalinii: No, just 4 states

catalinii:
Updated

Jalle19:
~LGTM~ found a few more things

src/adapter.cpp
```diff
@@ -553,7 +553,7 @@ void dump_pids(int aid) {
     if (!p)
         return;
     for (i = 0; i < MAX_PIDS; i++)
-        if (p->pids[i].flags > 0) {
+        if (p->pids[i].flags > PID_STATE_INACTIVE) {
```
Jalle19, line 556: This now includes PID_STATE_DELETED too, is that the idea? Or should it only concern active and new PIDs?

src/adapter.cpp
```diff
@@ -937,14 +939,15 @@ void post_tune(adapter *ad) {
 #endif
 #ifndef DISABLE_PMT
     SPid *p_all = find_pid(aid, 8192);
-    if (!p_all || p_all->flags == 3) { // add pids if not explicitly added
+    if (!p_all ||
+        p_all->flags == PID_STATE_DELETED) { // add pids if not explicitly added
         int pid;
         uint32_t ppid = 0;
         int pids[] = {0, 1, 16, 18}; // pids not added by the PMT module
```
Jalle19, line 946: This array could be declared as a constant somewhere, similar to the PID list used when emulating pids=all. Additionally I think this should check if the adapter is type CI and skip the whole thing.

Jalle19: But this can be done in a later PR

src/adapter.cpp
```diff
@@ -1016,9 +1019,12 @@ SPid *find_pid(int aid, int p) {
     if (!ad)
         return NULL;
 
-    for (i = 0; i < MAX_PIDS; i++)
-        if ((ad->pids[i].flags > 0) && (ad->pids[i].pid == p))
+    for (i = 0; i < MAX_PIDS; i++) {
+        if ((ad->pids[i].flags > PID_STATE_INACTIVE) && (ad->pids[i].pid == p))
             return &ad->pids[i];
+        else if (ad->pids[i].flags == PID_STATE_INACTIVE)
```
Jalle19: Is this due to the new sorting? Maybe add a comment on top of the loop that explains that since the PIDs are sorted we don't need to look past the first inactive one found.

src/adapter.cpp
catalinii: Yeah this prints all pids and all states (if not 0)

src/adapter.cpp
catalinii: Added

Issue #1355, opened 2025-12-30
Crash when attempting to use more than 4 CA PMTs per adapter


Jalle19:
```
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.706 AD9]: returning new pmt 25 for adapter 9, pmt pid 100, sid 100 0064
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: new PMT 25 AD 9, pid: 0064 (100), filter 32, len 66, pi_len 6, ver 1, pcr 1001, sid 0064 (100)
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: PMT 25 PI pos 1 caid 4AFC => pid 03E8 (1000), index 0
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: PMT pid 100 - stream pid 03E9 (1001), type 27, es_len 11, pos 15, caids 1
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: PMT pid 100 - stream pid 03EA (1002), type 6 [AC3], es_len 21, pos 31, caids 1
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: PMT pid 100 - stream pid 03EB (1003), type 11, es_len 0, pos 57, caids 1
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: Sending PMT 25 to all CAs: ad_ca_mask 3, pmt_ca_mask 0, disabled_ca_mask 0
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.707 AD9]: CAID 4AFC and mask FFFF matched PMT 25
Dec 30 13:44:31 debian minisatip[1737]: RECEIVED SIGNAL 4 (2) - SP=0 IP=0
Dec 30 13:44:31 debian minisatip[1737]: [30/12 13:44:31.740 AD9]: Stack trace:
Dec 30 13:44:31 debian minisatip[1737]:    0# print_trace() at /__w/minisatip/minisatip/src/utils.cpp:230
Dec 30 13:44:31 debian minisatip[1737]:    1# posix_signal_handler(int, siginfo_t*, ucontext_t*) at /__w/minisatip/minisatip/src/utils.cpp:258
Dec 30 13:44:31 debian minisatip[1737]:    2#      at :0
[...]

PR #1357, opened 2025-12-31
Increase MAX_CA_PMT, cap value supplied by user

Jalle19:
This also fixes wrong default when no `-c` parameter is used.

Fixes https://github.com/catalinii/minisatip/issues/1355

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -3105,7 +3105,11 @@ void set_ca_channels(char *o) {
             multiple_pmt = 1;
         }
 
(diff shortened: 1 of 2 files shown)


Jalle19:
This seems to have broken something. If I use no `-c` parameter at all, the CAM can open just a single channel as expected. If I specify `2:2`, AD2 (the CAM) can open two channels. But if I use `2:*2`, the CAM can't open anything at all. It doesn't even try.

Issue #1318, opened 2025-09-23
Trouble: emulate_add_all_pids() overload


lars18th:
Hi @catalinii and @Jalle19 ,

When fixing the missing PMT pids in emulated pids all (#1317, and not related to this issue!) I have noticed that this function is generating a huge overload.

If you execute minisatip with "-k" inside a regular x64 server using the `satipc` module, you will see that this function is permanently executed in each loop. Something like:
```
[23/09 13:36:54.876 AD1]: adapter 1: Nothing to commit
[23/09 13:36:54.878 AD1]: emulate_add_all_pids: adding (enforced) pid 1 to emulate all pids
[23/09 13:36:54.878 AD1]: found already existing pid 1 flags 1
[23/09 13:36:54.878 AD1]: emulate_add_all_pids: adding (enforced) pid 16 to emulate all pids
[23/09 13:36:54.878 AD1]: found already existing pid 16 flags 1
[23/09 13:36:54.878 AD1]: emulate_add_all_pids: adding (enforced) pid 17 to emulate all pids
[23/09 13:36:54.878 AD1]: found already existing pid 17 flags 1
[23/09 13:36:54.878 AD1]: emulate_add_all_pids: adding (enforced) pid 18 to emulate all pids
[23/09 13:36:54.878 AD1]: found already existing pid 18 flags 1
[23/09 13:36:54.878 AD1]: emulate_add_all_pids: adding (enforced) pid 20 to emulate all pids
[23/09 13:36:54.878 AD1]: found already existing pid 20 flags 1
[23/09 13:36:54.878 AD1]: emulate_add_all_pids: adding (enforced) pid 21 to emulate all pids
[23/09 13:36:54.878 AD1]: found already existing pid 21 flags 1
[23/09 13:36:54.878 AD1]: satipc: commit on freq 10773, sys 6 for adapter 1 (expect reply 0)
```

The problems with standard enigma2 boxes are then:

1. Calling to this function in each loop consumes a lot of CPU.
2. Futhermore the LOG is generating a lot of verbosity and consumes CPU too.

Therfore, I suggest to change the `LOG` to `LOGM`. And starting to think for a solution to not call to this function when is not necessary.
Any ideas?

Regards.

Jalle19:
The function should exit early if there is no pids=all request active, isn't that enough?

lars18th:
> The function should exit early if there is no pids=all request active, isn't that enough?

This code is already on it:
```
[...]

PR #1358, opened 2025-12-31
Rework emulate pids all to be done when updating pids #1318

catalinii:


Jalle19 (approved):

lars18th:
Hi @catalinii ,

Thank you for rewrite it. However, I can't test it these days (out of office).

However, one point to address:
- Adding enforced pids (`EMU_PIDS_ALL_ENFORCED_PIDS_LIST`) requires to be done _only one time_. In the current code every time the loop `for (auto fpid : EMU_PIDS_ALL_ENFORCED_PIDS_LIST)` is executed. I suggest to optimize the current behaviour with this strategy: use a FLAG to know the `emulate_add_all_pids()` state: INIT, RUNNING, UPDATING, etc. Then the enforced pids requires to be added only in the INIT state.

I hope this helps. 😉

catalinii:
Yeah that would work if minisatip would know when all the PMTs have been processed, but it does not as each PMT is processed after each other.

the current implementation is pretty cheap O(number of added pids) so further optimization would not bring much advantage but it would complicate the code.

lars18th:
Hi @catalinii ,

> Yeah that would work if minisatip would know when all the PMTs have been processed, but it does not as each PMT is processed after each other.

This is true for "added" pids. However, for "enforced" pids you only need to add them _one time_. So why then not add a simple bool to add them the first time and not in further requests?

lars18th:
Hi @catalinii ,

Futhermore, I suggest to change: `uint16_t EMU_PIDS_ALL_ENFORCED_PIDS_LIST[] = {0, 1, 16, 17, 18, 20, 21};`
with: `constexpr uint16_t EMU_PIDS_ALL_ENFORCED_PIDS_LIST[] = {...};`
because the current code is C++. 😉

catalinii:
> This is true for "added" pids. However, for "enforced" pids you only need to add them _one time_. So why then not add a simple bool to add them the first time and not in further requests?

the goal is for minisatip to support multiple clients asking for pids=all and potentially other clients pods=0,x,y.

in this case tou need to add even pids 0,1,… fpr multiple streams (so multiple times).
Also if they start the streams at different time, the pids also need to be added at different times.

Thanks for the suggestion related to usng constexpr

PR #1360, opened 2026-01-02
WIP: Don't use fake SIDs in CA PMTs

Jalle19:
Testing a theory

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -556,22 +556,15 @@ int dvbca_process_pmt(adapter *ad, SPMT *spmt) {
     ver = capmt->version;
     sid = spmt->sid;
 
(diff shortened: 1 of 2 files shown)


Jalle19:
Tried making the code simpler, haven't noticed any CAM that actually requires the SID to be correct. @catalinii do you remember your reasoning behind the original logic?

Jalle19:
Trying another approach

Jalle19:
Current version seems to make Kabelio CAMs work better. If there's a SID overlap when trying to descramble two services from two separate transponders, they will both have the same SID, but nothing seems to indicate that this matters.

Jalle19:
Not sure whether to go forward with this or not, having some issues with a CAM and no matter what I do regarding the CA PMT I can't get it to behave the way I want it to...

PR #1364, opened 2026-01-04
WIP: CA PMT tweaks

Jalle19:


Jalle19:
`Jan 03 08:35:20 debian minisatip[7456]: [03/01 08:35:20.822 AD9]: PMT CA 9 pmt 3 pid 221 () ver 0 sid 221 (221), enabled_pmts 1, CLM_ONLY, PMTS to be send 3 -1, pos 0`

Jalle19:
Seems like version 0 is actually fine, it happens regularly in PMTs. The module should be 32 though, not 16.

Jalle19:
Here it jumps to 16 instead of to 0, then after 16 it jumps to 1 instead of 0. This works because we should actually be using module 32, but since we use a smaller, we end up not actually overflowing the value.

```
Jan 04 00:17:00 vladimir minisatip[312513]: [04/01 00:17:00.173 AD14]: PMT CA 14 pmt 643 pid 1283 () ver 15 sid 31000 (15200), enabled_pmts 2, CLM_UPDATE, PMTS to be send 524 643, pos 0
Jan 04 00:19:43 vladimir minisatip[312513]: [04/01 00:19:43.663 main]: PMT CA 14 DEL pmt 643 pid 1283, sid 31000 (15200), ver 16, CLM_UPDATE name: SVT Barn / SVT24 HD
Jan 04 00:19:44 vladimir minisatip[312513]: [04/01 00:19:44.270 AD14]: PMT CA 14 pmt 643 pid 1284 () ver 1 sid 23300 (15200), enabled_pmts 2, CLM_UPDATE, PMTS to be send 524 643, pos 0
```

Jalle19:
My ultimate goal is to replace `multiple_pmt` with a list of PMTs per CA PMT, so users can attempt to cram three PMTs into one CA PMT.

It's not that straight-forward since the current logic depends on all CA PMT "slots" being already instantiated, and "free" slots are indicated using `PMT_INVALID`. We also return pointers to CA PMTs and mutate them (mainly to set SID, which is why I've been trying to refactor that part a bit).

Jalle19:
Last commit still doesn't work, will be hard to make incremental changes. It currently ends up in a state where there's only a single PMT to be played, but it still uses CLM_UPDATE, causing descrambling failure.

Jalle19:
I finally got the code refactored to allow for more than two PMTs per CA PMT, only to realize that my CAMs don't support it :D Wondering why not.

Jalle19:
I also get a lot of continuity errors when I add a third channel, which doesn't make sense

catalinii:
Your CAM does not support 1 CAPMT and 2 channels?
That is rare

Jalle19:
It supports 2 CA PMT with two channels per, but I'm trying to go beyond that. I was sure I had it hacked together and working at some point but not anymore it seems. Need to do a bit more testing.

PR #1361, opened 2026-01-03
stores pid streams as a set

catalinii:
Storing streams was a bit of a mess with adding/removing/sorting and storing -1 as indicator that a stream is not set.

Replace the array with an unordered_set.

PR #1365, opened 2026-01-04
Fix inverted condition that caused stream PIDs to be missing from DDCI PMT

Jalle19:
Broke in https://github.com/catalinii/minisatip/pull/1361


Jalle19:
I had a commit lying around that removes this logic altogether. The only place we need to filter out unsubscribed PIDs from is the CA PMT. Since in normal operation the whole mux is sent through the CAM it seems unreasonable that we benefit from filtering out PIDs from the DDCI transport stream.

@catalinii let me know what you think and I'll update the PR

catalinii:
Thanks for fixing

PR #1366, opened 2026-01-05
Don't attempt to filter out PIDs from DDCI PMTs

Jalle19:
In normal operation, the whole mux is sent to the CAM, and all CAMs can handle that just fine, even though a mux can have upwards of 100 PIDs. When this filtering logic was added the main concern was that the CA PMT would balloon in size, however, that filtering happens elsewhere independent of this.

@catalinii continuation of https://github.com/catalinii/minisatip/pull/1365

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -271,12 +271,6 @@ int get_max_pmt_for_ca(int i) {
     return MAX_CA_PMT;
 }
 
-int get_ca_multiple_pmt(int i) {
-    if (i >= 0 && i < MAX_ADAPTERS && ca_devices[i] && ca_devices[i]->enabled)
-        return ca_devices[i]->multiple_pmt;
-    return 0;
-}
-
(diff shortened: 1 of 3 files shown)


catalinii:
@Jalle19 thr code here was added assuming there are 2 channels from different transponders decrypted by 1 CAPMT.

Most of the official CAMs can decrypt max 1 channel

Jalle19:
Yes, but this particular code has nothing to do with the CA PMT.

Jalle19:
This is actually not right after all - when `create_capmt` looks at which PIDs are subscribed, the adapter is the DDCI adapter, so if we don't leave them out from the PMT they'll never be omitted.

Jalle19:
Found another bug in this logic though, will make new PR

catalinii:
Maybe we should still rewrite it. The filtering should be just in the CAPMT. The actual pmt sent to the DDCI adapter (and CI) does not rewlly matter

PR #1369, opened 2026-01-06
Switch AES ECB implementation to openssl 3

catalinii:
- Moved back CI+ implementation to ca.cpp (copy paste)
- changes the aes.cpp implementation to be openssl 3 compatible
- this impacts AES ECB which is not used for CI+ (which uses CBC)

### DIFF
--- a/src/aes.cpp
+++ b/src/aes.cpp
@@ -17,7 +17,6 @@
  * USA
  *
  */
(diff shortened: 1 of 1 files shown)


Jalle19:
We really need to split ca.cpp into more files

catalinii:
Right, it was more files originally

PR #1370, opened 2026-01-06
Switch ca.cpp to openssl 3 and one shot methods

catalinii:
Switches CI+ routines to work on both openssl 3 and 1.x.
Removes also warnings and make the code easier to understand using one-shot encryption.

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -279,6 +279,38 @@ static int sha256_final(EVP_MD_CTX *ctx, uint8_t *out) {
     return ret;
 }
 
+int RSA_private_encrypt_wrapper(EVP_PKEY *pkey, int dlen, uint8_t *dbuf,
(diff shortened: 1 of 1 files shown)


Jalle19:
Have you checked whether the .bin file for a CAM created with a previous version of minisatip is still valid and working with this new code?

Jalle19:
And maybe vice versa too - create the .bin file with this patch and then try master again and see if it works

catalinii:
I’ve tested the scenarios I tjought are relvant:
1) start from an old .bin
2) remove the .bin and restart negotiation

all woked for me

catalinii:
> And maybe vice versa too - create the .bin file with this patch and then try master again and see if it works

yeah this works too.

Jalle19 (approved):

PR #1372, opened 2026-01-09
Simplifying dh_gen_exp to use random numbers

catalinii:
This is a continuation of #1370 but does actual behavior changes and wanted to keep it in a separate diff for bisection purposes.

This updates some methods to higher level openssl 3 APIs to simplify the old code.

### DIFF
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -5,20 +5,13 @@ alternative source
 
  */
 
-#include <arpa/inet.h>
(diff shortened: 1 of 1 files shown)


Jalle19:
Seems a bit too easy, why was the original implementation so complex?

catalinii:
Th original implementation is more generic however for DH seems random data is fine

PR #1375, opened 2026-01-18
Test bazel

catalinii:


Jalle19:
I would much rather switch to CMake instead.

catalinii:
Right, this is just to replace the build image and do cross compiling for mips and arm too

Jalle19:
Ah, right! This all seems a bit complex to me, does it bring any new benefits apart from not needing your old Docker image?

catalinii:
Feels a bit complex to me too, nothing new.

If I can build other additional deps with cmake then there is no need for this

PR #1376, opened 2026-01-19
Switch the build system to cmake

catalinii:


src/CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
...
+    utils/uuid.cpp
+    dvb.cpp  # Always included per Makefile
+)
+
+# Libraries to link
+set(MINISATIP_LIBS pthread)
+
+# Track feature dependencies
+set(NEEDS_PMT OFF)
+set(NEEDS_TABLES OFF)
```
Jalle19: IMO we should take this opportunity to remove these ifdefs. We almost never build without them anyway, and if a Digibit R1 can handle it then I think any other hardware can.

src/CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
...
+        list(APPEND STATIC_LIBS "-l:lib${LIB}.a")
+    endforeach()
+    target_link_libraries(minisatip PRIVATE ${STATIC_LIBS})
+    target_link_options(minisatip PRIVATE -static)
+else()
+    target_link_libraries(minisatip PRIVATE ${MINISATIP_LIBS})
+endif()
+
+# Install target
+install(TARGETS minisatip RUNTIME DESTINATION bin)
```
Jalle19: I'd skip this

CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+    set(VERSION_STRING "${GIT_TAG}+PR$ENV{PR}~${GIT_REVISION}")
+else()
+    set(VERSION_STRING "${GIT_TAG}~${GIT_REVISION}")
+endif()
+
+if(NOT VERSION_STRING)
+    set(VERSION_STRING "unknown")
+endif()
+
+project(minisatip VERSION 1.0 LANGUAGES CXX)
```
Jalle19: Can we omit version here? Or use the number determined from git?

.github/workflows/binaries.yml
```diff
@@ -35,31 +35,25 @@ jobs:
...
+        rm -rf build; CXX="zig c++" cmake -B build . -DCMAKE_CXX_COMPILER_TARGET=mipsel-linux-gnueabihf -DNETCVCLIENT=0 && (cd build && make)
         file minisatip
         zip -9 -r /minisatip_mips.zip minisatip html
 
     - name: Build MIPS (musl)
       run: |
-        make clean
-        ./configure --disable-netcv --enable-static --host=mipsel-linux-gnu CXX="zig c++ -target mipsel-linux-musl"
+        rm -rf build; CXX="zig c++" cmake -B build . -DCMAKE_CXX_COMPILER_TARGET=mipsel-linux-musl -DNETCVCLIENT=0 && (cd build && make)
         make
```
Jalle19: This should be removed now. I'd use one command per line in all of these to make it easier to grasp

.github/workflows/coverity.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     - uses: actions/checkout@v4
     - name: Install build tools and dependencies
       run: DEBIAN_FRONTEND=noninteractive sudo apt-get -y install build-essential libdvbcsa-dev libssl-dev zlib1g-dev libxml2-dev
-    - name: configure
-      run: ./configure
+    - name: cmake
+      run: mkdir build;cd build;cmake ..
```
Jalle19: Why not `&&` instead? Or one command per line

src/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+
+foreach(OPT ${OPTIONS})
+    if(NOT ${OPT})
+        string(APPEND CONFIG_CONTENT "#define DISABLE_${OPT}\n")
+    endif()
+endforeach()
+
+string(APPEND CONFIG_CONTENT "#define VERSION \"${VERSION_STRING}\"\n")
+string(APPEND CONFIG_CONTENT "\n#endif // CONFIG_H\n")
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/config.h "${CONFIG_CONTENT}")
```
Jalle19: Is this the standard way to inject defines? I'm not that familiar with cmake but it feels like there must be a less manual way?

Jalle19: In tests we use `add_compile_definitions`

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
+cmake_minimum_required(VERSION 3.16)
```
Jalle19, line 1: Would it make sense to use just the top-level CMakeLists.txt, i.e. no separate one inside src/?

src/CMakeLists.txt
catalinii: That is the other option like we do now (add to CXX_FLAGS), cmake allows also doing replaces in existing files, but autogenerating was the easy vibe implementation.

I am open to add to the cxxflags, but the issue is that cmake runs one time, but if the repo is updating without running CMAKE then the VERSION will be wrong.

Jalle19: I think it's okay if version only gets updated when cmake is run

.github/workflows/binaries.yml
catalinii: Fixed

CMakeLists.txt
catalinii: I am fine either way

Jalle19: Would be less to maintain to have just a single file

catalinii:
@Jalle19 ready for review

PR #1383, opened 2026-02-08
dvbapi: add descramble watchdog and extend CW lifetime

mrgs83:
This PR targets an intermittent DVBAPI descramble stall observed with OSCam: scrambled TS continues but successful decrypt stops until a zap/reconnect.

Changes:
- Extend DVBAPI CW lifetime to 180s (avoid parity-switch freezes when parity is unchanged > default CW timeout).
- Add PMT-level descramble watchdog: track last scrambled vs last decrypt-ok; on stall clear CWs + resend CAPMT once; hard fallback restarts PMT state with cooldown.
- Expose dvbapi_send_pmt() and implement dvbapi_enabled() for the watchdog.

Files: src/dvbapi.cpp, src/dvbapi.h, src/pmt.cpp, src/pmt.h.

Happy to adjust thresholds/logging/style if you prefer different defaults.

### DIFF
--- a/src/dvbapi.cpp
+++ b/src/dvbapi.cpp
@@ -56,6 +56,8 @@ int dvbapi_sock = -1;
 int sock;
 int dvbapi_is_enabled = 0;
 int enabledKeys = 0;
+
+int dvbapi_enabled() { return dvbapi_is_enabled; }
 int network_mode = 1;
 int dvbapi_protocol_version = DVBAPI_PROTOCOL_VERSION;
 int dvbapi_ca = -1;
@@ -318,7 +320,10 @@ int dvbapi_reply(sockets *s) {
                     k_id, parity, index, k->algo, correct ? "OK" : "NOK", cw[0],
                     cw[1], cw[2], cw[3], cw[4], cw[5], cw[6], cw[7]);
 
-                send_cw(k->pmt_id, k->algo, parity, cw, NULL, 0, &k->icam_ecm);
+                // Some systems may keep the same parity for >45s; keep CWs longer to
+                // avoid short freezes on parity switch when the "other" parity CW
+                // expired on our side.
+                send_cw(k->pmt_id, k->algo, parity, cw, NULL, 180, &k->icam_ecm);
             } else
                 LOG("dvbapi: invalid DVBAPI_CA_SET_DESCR, key %d parity %d, k "
                     "%p, "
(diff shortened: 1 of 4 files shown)


mrgs83:
FYI: GitHub shows workflows awaiting maintainer approval (fork PR). Please approve CI if you want checks to run.

lars18th:
Hi @mrgs83 ,

I think this PR is to address this use case: a _continuous_ streaming of the same channel without user interaction. In this case, this simulates a channel zapping that returns to the same channel. This is true?

In this case, this seems OK. Why not merge it, @catalinii ?

Jalle19:
Are we sure this isn't just a workaround for an oscam bug?

lars18th:
Hi @Jalle19 ,

> Are we sure this isn't just a workaround for an oscam bug?

I don’t think so. As I understand it, this is intended to resolve the issue when decoding fails _for whatever reason_. In some circumstances, the only way to resume decoding is to restart the entire process. This can be done easily by manually changing the channel. However, if you want continuous decoding 24/7 of the same channel, that’s when this solution makes sense. So it can be seen as a last resort in the event of a critical decoding failure. It might also be useful with CI.

So please @mrgs83 ... can you explain more?
EDIT: Full description at https://github.com/catalinii/minisatip/discussions/1381#discussion-9442822

lars18th:
Hi @catalinii ,

Why not merge this?

catalinii:
I still need to review it properly, more specifically around how to keep the existing boundaries in place. Basically what needs to be done in pmt to stay in pmt, everything else in dvbapi.cpp. Will reply once I have a clear idea how to do it, as I’ve been swamped lately.

catalinii:
hi @mrgs83 

I have a proposal for this change to keep clean the separation between pmt.cpp and dvbapi.cpp, please take a look: https://github.com/catalinii/minisatip/commit/e48b530935b161d4c3e53b89a07de7b87cd05c66

I did not test this, just asked an LLM to keep clean separation between the modules, feel free to test/let me know if it works.

if yes, update your diff with the change from the diff generated.

ci70:
Was it finally merged?

catalinii:
Hi @ci70 ,

can you test this build?

We can merge once we get confirmation this works.

thanks

lars18th:
Hi @mrgs83 and @ci70 ,

You need to check the branch `test`: https://github.com/catalinii/minisatip/tree/test
If all is working right, then please give us feedback. We need confirmation before merging it.

Regards.

PR #1384, opened 2026-02-10
Handle default PIDs differently depending on adapter type

Jalle19:
I'd like to get rid of `Pids: 0, 16, 18` that is always visible on CI adapters in the web interface. This approach here seems to work, but I have a feeling it won't work for non-DDCI CI adapters. Maybe we need to be able to distinguish those somehow? @catalinii what do you think?

### DIFF
--- a/src/adapter.cpp
+++ b/src/adapter.cpp
@@ -940,9 +940,11 @@ void post_tune(adapter *ad) {
     LOGM("adapter post_tune: aid %d", aid);
 #endif
 #ifndef DISABLE_PMT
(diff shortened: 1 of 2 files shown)


Jalle19:
@catalinii are you able to test this on Enigma?

catalinii:
image

Jalle19:
@catalinii so it works? :D

catalinii:
I installed it, what should I see?

Jalle19:
Playback of encrypted channels via a CAM should work.

catalinii:
I did not see anything wrong with that version. I used it for few daya

PR #1388, opened 2026-02-28
Add SRT (Secure Reliable Transport) support for TS streaming

catalinii:
Integrate the SRT library as an alternative transport between SAT>IP client and server, replacing RTP/UDP with reliable, low-latency delivery over lossy networks.

Server side (stream.cpp):
- New STREAM_RTSP_SRT stream type triggered by Transport: RTP/AVP/SRT
- decode_transport() parses SRT port, creates caller socket, connects
- flush_stream() sends raw TS data chunked to 1316 bytes (7x188)
- process_packets_for_stream() skips RTP headers and socket checks for SRT
- close_stream() and stream_timeout() handle SRT cleanup

Client side (satipc.cpp):
- ^ prefix on -s argument enables SRT mode (e.g. ^dvbc:host:port)
- SRT listener with dual-stack IPv4/IPv6 support
- Receiver thread: srt_accept() -> srt_recv() -> pipe -> event loop
- Transport header: RTP/AVP/SRT;port=<listen_port>

Build system (CMakeLists.txt):
- SRT option auto-detected, conditional compilation via DISABLE_SRT
- Non-SRT builds exclude all SRT code paths


Jalle19:
Very nice, will test this

src/satipc.cpp
```diff
@@ -1527,7 +1668,11 @@ std::string satipc_name(int aid, int fd) {
 void satipc_free(adapter *ad) {}
 
 int add_satip_server(char *host, int port, int fe, char delsys, char *source_ip,
-                     int use_tcp, int no_pids_all) {
+                     int use_tcp, int no_pids_all
+#ifndef DISABLE_SRT
+                     , int use_srt
+#endif
```
Jalle19: Maybe we could include `use_str` unconditionally and avoid having to hack the function definition?

src/satipc.h
```diff
@@ -61,6 +65,13 @@ typedef struct struct_satipc {
     unsigned int want_tune : 1;
     unsigned int force_pids : 1;
     unsigned int sent_transport : 1;
+#ifndef DISABLE_SRT
+    char use_srt;           // use SRT transport instead of UDP/TCP
+    SRTSOCKET srt_listener; // SRT listener socket (waiting for server connection)
+    SRTSOCKET srt_sock;     // accepted SRT connection (or SRT_INVALID_SOCK)
+    int srt_accepted;       // 1 when connection is accepted
```
Jalle19: Could this be bool instead?

src/satipc.h
```diff
@@ -61,6 +65,13 @@ typedef struct struct_satipc {
     unsigned int want_tune : 1;
     unsigned int force_pids : 1;
     unsigned int sent_transport : 1;
+#ifndef DISABLE_SRT
+    char use_srt;           // use SRT transport instead of UDP/TCP
```
Jalle19: Could this be bool instead?

src/stream.cpp
```diff
@@ -395,6 +406,72 @@ int decode_transport(sockets *s, char *arg, char *default_rtp, int start_rtp) {
...
+            get_sockaddr_host(s->sa, dest, sizeof(dest) - 1);
+
+            // Create SRT caller socket and connect to the client's SRT
+            // listener
+            SRTSOCKET srt_s = srt_create_socket();
+            if (srt_s == SRT_INVALID_SOCK)
+                LOG_AND_RETURN(-1, "srt_create_socket failed: %s",
+                               srt_getlasterror_str());
+
+            int latency = 120;
```
Jalle19: Should this be configurable or at least made into a constant?

Jalle19 (commented):
Web interface should be updated to print RTP/SRT or something when a satipc adapter uses this transport

Jalle19:
Are there any other SRT-related parameters that might be worth making configurable?

estimadarocha:
@catalinii do you think this can be pushed to https://github.com/oe-alliance/satip-client

catalinii:
@Jalle19 addressed

src/socketworks.h
```diff
@@ -65,7 +65,6 @@ typedef struct struct_sockets {
 #define TYPE_RTSP 4
 #define TYPE_DVR 5
 #define TYPE_RTCP 6
-#define TYPE_SRT 7
```
Jalle19: Why was this removed?

src/satipc.cpp
```diff
@@ -549,7 +550,7 @@ int satipc_open_device(adapter *ad) {
     sip->init_use_tcp = sip->use_tcp;
 #ifndef DISABLE_SRT
     if (sip->use_srt) {
-        sip->listen_srt = opts.start_rtp + 2000 + ad->id * 2;
+        sip->listen_udp = opts.start_rtp + 1000 + ad->id * 2;
```
Jalle19: So now it's not possible to use both RTP/UDP and SRT at the same time?

src/socketworks.h
catalinii: not used

src/satipc.cpp
catalinii: it is, just uses the same port range so you do not need to setup multiple port forwardings for the same adapter.

lars18th:
Nice!

lars18th (changes requested):
Hi @catalinii ,

Why not add the option to modify this value (`SRT_LATENCY_MS`)? Perhaps 1s is good as default for regular WAN streaming in common networks. But in some circunstances a lower or higher value could be useful.

catalinii:
> Hi @catalinii ,
> 
> Why not add the option to modify this value (`SRT_LATENCY_MS`)? Perhaps 1s is good as default for regular WAN streaming in common networks. But in some circunstances a lower or higher value could be useful.

sure, but will do that in a differnet PR when I see what other command line args are needed for srt

src/satipc.h
```diff
@@ -13,6 +17,10 @@
 #define SATIP_STATE_TEARDOWN 3
 #define SATIP_STATE_INACTIVE 0
 
+#define SIP_TRANSPORT_UDP 0
+#define SIP_TRANSPORT_TCP 1
+#define SIP_TRANSPORT_SRT 2
```
Jalle19: Maybe we could use an enum instead?

src/satipc.cpp
```diff
@@ -477,6 +485,97 @@ int satipc_rtcp_reply(sockets *s) {
...
+        sip->srt_sock = accepted;
+        sip->srt_accepted = true;
+        get_sockaddr_host(*(USockAddr *)&addr, dest, sizeof(dest));
+        LOG("SRT connection accepted for adapter %d, srt_sock=%d, from %s:%d",
+            ad->id, accepted, dest, get_sockaddr_port(*(USockAddr *)&addr));
+    }
+
+    // Read available SRT data into the adapter buffer
+    int total = 0;
+    while (total + 1316 <= len) {
```
Jalle19: 1316 appears a bit here and there, could be #defined

src/satipc.cpp
```diff
@@ -1006,7 +1126,12 @@ int satip_post_init(adapter *ad) {
     if (!sip)
         return 0;
 
-    if (sip->init_use_tcp)
+#ifndef DISABLE_SRT
+    if (sip->transport_type == SIP_TRANSPORT_SRT) {
+        sockets_setread(ad->sock, (void *)satipc_srt_read);
+    } else
+#endif
+        if (sip->transport_type == SIP_TRANSPORT_TCP)
```
Jalle19, line 1200: Bad indentation

src/satipc.cpp
```diff
@@ -1519,15 +1652,14 @@ std::string satipc_name(int aid, int fd) {
...
-
+    ss << sip->sip << " @ " << str_transport_type[sip->transport_type];
     return ss.str();
 }
 
 void satipc_free(adapter *ad) {}
 
 int add_satip_server(char *host, int port, int fe, char delsys, char *source_ip,
-                     int use_tcp, int no_pids_all) {
+                     char transport, int no_pids_all) {
```
Jalle19: Shouldn't this be `int`?

src/satipc.cpp
```diff
@@ -1623,11 +1754,22 @@ void find_satip_adapter(adapter **a) {
...
     for (i = 0; i < la; i++) {
-        use_tcp = opts.satip_rtsp_over_tcp;
+        char transport =
+            opts.satip_rtsp_over_tcp ? SIP_TRANSPORT_TCP : SIP_TRANSPORT_UDP;
         no_pids_all = 0;
+        if (arg[i][0] == '^') {
+            transport = SIP_TRANSPORT_SRT;
+            arg[i]++;
+#ifdef DISABLE_SRT
+            FAIL("SRT support is disabled, install libsrt-openssl-dev")
```
Jalle19, line 1836: Re-compilation is also required so maybe we should just warn here that SRT was requested but cannot be fulfilled and that it will fallback to UDP instead?

src/stream.cpp
```diff
@@ -45,6 +45,11 @@
 #include <unistd.h>
 #include <vector>
 
+#ifndef DISABLE_SRT
+#include <srt/srt.h>
+#define SRT_LATENCY_MS 120
```
Jalle19: This is now defined in two places

src/stream.cpp
```diff
@@ -380,6 +392,70 @@ int close_stream(int i) {
...
+        srt_close(sid->srt_sock);
+    }
+
+    SRTSOCKET srt_s = srt_create_socket();
+    if (srt_s == SRT_INVALID_SOCK)
+        LOG_AND_RETURN(-1, "srt_create_socket failed: %s",
+                       srt_getlasterror_str());
+
+    int latency = SRT_LATENCY_MS;
+    srt_setsockflag(srt_s, SRTO_LATENCY, &latency, sizeof(latency));
```
Jalle19: Should this be set on both sockets or just on the sender?

src/stream.cpp
```diff
@@ -709,19 +795,49 @@ int send_rtcp(int s_id, int64_t ctime) {
 int flush_stream(streams *sid, struct iovec *iov, int iiov, int64_t ctime) {
     int rv = 0;
     char ra[50];
-    rv = sockets_writev(sid->rsock_id, iov, iiov);
+
+#ifndef DISABLE_SRT
+    if (sid->type == STREAM_RTSP_SRT) {
+        // Send raw TS data over SRT, chunked to max 1316 bytes (7 * 188)
+        static const int SRT_TS_CHUNK = 1316;
```
Jalle19: Could be moved elsewhere so it can be reused

src/stream.cpp
```diff
@@ -380,6 +392,70 @@ int close_stream(int i) {
...
+    sid->srt_sock = srt_s;
+    memcpy(&sid->sa, &sa, sizeof(sa));
+    sid->rsock = -1;
+    sid->rsock_id = -1;
+
+    LOG("SRT connected to %s:%d, srt_sock=%d for sid %d", dest, srt_port, srt_s,
+        sid->sid);
+    return 0;
+#else
+    LOG0("SRT streams are not supported, install libsrt-openssl-dev");
```
Jalle19: Same thing here with the error message

src/stream.cpp
```diff
@@ -913,15 +1031,26 @@ int process_packets_for_stream(streams *sid, adapter *ad) {
...
-    sockets *s = get_sockets(sid->rsock_id);
-    if (!s)
-        LOG_AND_RETURN(0, "Stream %d rsock id not found %d", st_id,
-                       sid->rsock_id);
-    std::lock_guard<SMutex> lock(s->mutex);
-    if (!s->enabled)
+    sockets *s = nullptr;
+#ifndef DISABLE_SRT
+    if (sid->type != STREAM_RTSP_SRT)
+#endif
```
Jalle19: STREAM_RTSP_SRT does not need guarding

src/stream.cpp
```diff
@@ -944,6 +1073,13 @@ int process_packets_for_stream(streams *sid, adapter *ad) {
         //		max_iov = max_pack;
     }
 
+#ifndef DISABLE_SRT
+    if (sid->type == STREAM_RTSP_SRT) {
+        iiov = 0;     // no RTP header slot
+        max_pack = 0; // no RTP batching
+    }
+#endif
```
Jalle19: STREAM_RTSP_SRT does not need guarding

src/stream.cpp
```diff
@@ -1190,7 +1332,11 @@ int stream_timeout(sockets *s) {
     ctime = getTick();
     s->rtime = ctime;
 
-    if ((sid = get_sid(s->sid)) && sid->type != STREAM_HTTP) {
+    if ((sid = get_sid(s->sid)) && sid->type != STREAM_HTTP
+#ifndef DISABLE_SRT
+        && sid->type != STREAM_RTSP_SRT
+#endif
```
Jalle19: Same here

src/stream.h
```diff
@@ -50,6 +52,10 @@ typedef struct struct_streams {
     char useragent[40];
 } streams;
 
+#ifdef DISABLE_SRT
+#define SRT_INVALID_SOCK -1
+#endif
```
Jalle19, line 59: I think we should only guard code that uses `srt_` functions

Jalle19 (changes requested):

CMakeLists.txt
```diff
@@ -248,6 +248,7 @@ if(BUILD_TESTING)
         crypto
         dl
         pthread
+	srt
```
Jalle19: Bad indentation

src/satipc.cpp
catalinii: I think because the user expect SRT if SRT is not available just fail rather than silently switch and get a bad experience as a result of UDP not being unreliable.

src/stream.cpp
catalinii: I do not think we should downgrade here to UDP if the server does not support SRT, just fail.

PR #1395, opened 2026-04-03
Update compile instructions in README

localhosthack0r:


catalinii:
I see the instructions are already updated in readme_footer and README.md is just generated from readme_header and readme_footer.

Once you add more help in both files, feel free to use gen_readme script

localhosthack0r:
Ok, I changed now in readme_footer.

PR #1416, opened 2026-08-20
fix(socketworks): disassociate all stream sockets in sockets_del_for_sid

catalinii:
### Summary
This PR fixes a race condition where a stale socket close callback from a previously torn-down RTSP session prematurely closes a newly allocated stream sharing the same stream ID (`sid`).

### Cause
When a stream closes, `close_stream()` calls `sockets_del_for_sid(sid)` to disassociate sockets from the stream ID. Previously, `sockets_del_for_sid()` only checked for `TYPE_RTSP` sockets, leaving RTP (`TYPE_UDP`) and RTCP (`TYPE_RTCP`) sockets associated with `ss->sid`. When deferred socket deletion executed in the main loop (~200ms later), `close_stream_for_socket()` was invoked on a newly assigned stream sharing the same `sid`, setting `sid->timeout = 1` and force-closing it.

### Fix
- Updated `sockets_del_for_sid()` in `src/socketworks.cpp` to disassociate `TYPE_RTSP`, `TYPE_HTTP`, `TYPE_UDP`, and `TYPE_RTCP` sockets by resetting `ss->sid = -1` and clearing `ss->close = NULL`.
- Added a guard in `close_stream_for_socket()` in `src/stream.cpp` to ignore sockets with `s->sid < 0`.

### DIFF
--- a/src/socketworks.cpp
+++ b/src/socketworks.cpp
@@ -968,8 +968,8 @@ int sockets_del_for_sid(int sid) {
         return 0;
     for (i = 0; i < MAX_SOCKS; i++)
         if ((ss = get_sockets(i)) && ss->sid >= 0 && ss->sid == sid &&
-            (ss->type == TYPE_RTSP || ss->type == TYPE_HTTP ||
-             ss->type == TYPE_RTP || ss->type == TYPE_RTCP)) {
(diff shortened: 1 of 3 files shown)


catalinii:
@copilot review

PR #1421, opened 2026-08-22
fix(dvbapi): read the iCAM mode only if the section is long enough

dnehring7:
Refactored after @Jalle19's reviews: the length clamp is gone — he is right that
it could never do anything — and so is the comment. What is left is one line.

## Problem

`send_ecm()` takes the iCAM mode from a fixed offset:

```c
if (b[2] - b[4] == 4) {
    k->icam_ecm = b[0x15];
```

without checking that the section reaches 0x16 bytes.

`len` at that point is the section length, and the filter layer guarantees that
`b .. b+len-1` lies inside `f->data`, so testing `len` is enough to make both
reads legal.

Two callers, two effects:

* **ECM filter.** `b` is `f->data`, so a short section reads leftovers from the
  previous section in the same filter. The key schedule is then selected from
  stale bytes. In-bounds, wrong value.
* **EMM filter** (`send_ecm()` is registered with `FILTER_ADD_REMOVE | isEMM`).
  `b` is `f->data + i` for a sub-section inside the buffer. A three byte
  sub-section at the end of a full buffer puts `i` at 3997, so `b[0x15]` is
  `f->data[4018]`, while `f->data` is 4002 bytes and `sizeof(SFilter)` is 4096 —
  the last two bytes of that read are past the allocation, the sixteen before it
  are the `flags` and `next_filter` members.

## Fix

```c
-    if (b[2] - b[4] == 4) {
+    if (len > 0x15 && b[2] - b[4] == 4) {
```

One line. `b[4]` is covered by the same guard through short-circuit evaluation.

A well formed ECM is unaffected: on a live VideoGuard feed all 36 captured
sections were far longer than 0x16 and take the same branch as before.

## Testing

Fedora 44, gcc 16, `cmake -DDVBCSA=ON -DDVBCA=ON -DBUILD_TESTING=ON`.

* `ctest` 13/13, with `TEST_SANITIZERS=ON` (default) and `OFF`.
* Live: OSCam over dvbapi, Digital Devices DVB-S2, Sky Germany / VideoGuard, iCAM
  active. Descrambling unchanged — a 3 minute recording of an encrypted service,
  0 encrypted packets, `ad_decerr 0`, `ad_ccerr 0`.

## Not in this PR

The derivation of the iCAM mode itself, and the disagreement between `b[0x15]` and
OSCam's `get_ecm_mode()`. No evidence it costs anyone a packet; I keep the
measurements if it ever becomes interesting.

### DIFF
--- a/src/dvbapi.cpp
+++ b/src/dvbapi.cpp
@@ -654,7 +654,7 @@ int send_ecm(int filter_id, unsigned char *b, int len, void *opaque) {
     len = ((b[1] & 0xF) << 8) + b[2];
     len += 3;
     k->last_ecm = getTick();
-    if (b[2] - b[4] == 4) {
+    if (len > 0x15 && b[2] - b[4] == 4) {
         k->icam_ecm = b[0x15];
         k->is_icam = k->icam_ecm ? 1 : 0;
     } else {


catalinii:
@dnehring7 were you able to test this PR?

Jalle19:
Does this actually fix a bug?

dnehring7:
# How this was tested

Short answer: the out-of-bounds read is proven synthetically, the disagreement
with OSCam's `get_ecm_mode()` is measured on a live VideoGuard feed, and we
were **not** able to turn that disagreement into a visible descrambling failure
on our own provider. Details and the raw numbers below, including what the
tests do not show.

## Setup

| | |
|---|---|
| minisatip | 2.0.94, `cmake -DDVBCSA=ON -DDVBCA=ON`, gcc 16, Fedora 44 x86_64 |
| Tuner | Digital Devices DVB-S2, single frontend (`-e 0 -Y 0:dvbs2`) |
| Card server | OSCam over dvbapi (`-o <host>:30001,2`) |
| Provider | Sky Germany, VideoGuard, CAIDs 0x098C / 0x098D / 0x09F0, iCAM |
| Client | VDR 2.8.2 with vdr-plugin-satip over RTSP |

`nm -D minisatip | grep set_ecm` shows `w dvbcsa_bs_key_set_ecm` (unmangled),
so iCAM really is active in this build — that is the other patch in this
series, and without it none of this is reachable on a distribution build.

## 1. The out-of-bounds read — synthetic, conclusive

`icam_repro.c` models only the two lines under test and allocates each ECM
section to its exact length, so a read past the section is a real
heap-buffer-overflow rather than a read into slack:

```console
$ gcc -fsanitize=address -g -o icam_repro icam_repro.c && ./icam_repro
== case 1: short VideoGuard ECM, header satisfies b[2]-b[4]==4 ==
   section length 16, real mode (last byte nibble) = 4
   current code reads b[0x15] of a 16 byte section -> ERROR: AddressSanitizer:
   heap-buffer-overflow on address 0x7b4d9a9e0025
READ of size 1 at 0x7b4d9a9e0025 thread T0
    #0 mode_current icam_repro.c:16
    #1 main icam_repro.c:52
0x7b4d9a9e0025 is located 5 bytes after 16-byte region [...0010,...0020)
```

The correct mode for that section is 4, the low nibble of its last byte. The
current code reads five bytes past the end instead.

This is the part we can prove without qualification. It is a property of the
expression, not of any particular provider.

## 2. The mode derivation on real ECMs — measured

We built minisatip with one extra log line in `send_ecm()` that computes both
derivations for every ECM and prints them side by side:

```c
int hdr_len  = ((b[1] & 0xF) << 8) + b[2] + 3;
int old_cond = (b[2] - b[4] == 4);
int old_mode = old_cond ? (hdr_len > 0x15 ? b[0x15] : -1) : 0;
LOG("ECMCMP caid=%04X assembled=%d hdr_len=%d b4=%02X cond=%d "
    "old_mode=%d new_mode=%d last=%02X",
    k->caid, len, hdr_len, b[4], old_cond, old_mode, k->icam_ecm,
    b[ecm_len - 1]);
```

36 ECM sections on a live VideoGuard channel, roughly eight minutes:

| count | assembled | hdr_len | `b[2]-b[4]==4` | current code | `get_ecm_mode()` | last byte |
|------:|----------:|--------:|:--------------:|-------------:|-----------------:|----------:|
| 33 | 183 | 183 | yes | 4 (from `b[0x15]`) | 4 | 0xE4 |
| 2 | 382 | 382 | no | 0 — iCAM **off** | 4 | 0xB4 |
| 1 | 30 | 30 | no | 0 — iCAM **off** | 6 | 0x46 |

Two observations from this:

* The two expressions **disagree on 3 of 36 sections**. For the majority case
  the fixed offset happens to land on a byte that carries the same value as
  OSCam's rule, which is presumably why the current code works at all. For the
  382 byte and 30 byte sections the arithmetic fails and the current code turns
  iCAM off, while OSCam's rule selects mode 4 and mode 6.
* `hdr_len == assembled` in all 36 sections. On this provider the header
  derived length never exceeds what the filter layer assembled, so the clamp in
  the patch is purely defensive here — see below.

## 3. What these tests do **not** show

Being explicit, because it bears on how much weight the patch deserves:

* **The OOB path was never reached in the field.** It needs a section shorter
  than 0x16 bytes that also satisfies `b[2] - b[4] == 4`. The shortest section
  we saw was 30 bytes, and it did not satisfy the condition. The reproducer is
  the only evidence, and it is synthetic.
* **We could not produce a visible failure from the wrong mode.** We built
  minisatip with this patch reverted (keeping the C linkage fix, so iCAM stays
  active) and captured the same channel for 90 seconds:

  ```
  pid  1279:  181611 packets, encrypted: 0 (0.0 %)
  pid  1283:   23611 packets, encrypted: 0 (0.0 %)
  ad_ccerr 0   ad_decerr 0
  ```

  No decoding errors, no encrypted packets. So on this provider the three
  mismatching sections did not cost us a single packet, and we cannot claim the
  current code breaks reception here. Whether they matter for a provider that
  actually varies the mode, we have no way to test.

## 4. Regression testing of the patched build

With the patch applied, the same setup ran for several hours of live viewing
and recording without a decryption failure:

* `pmt: decrypted pmt 0, packets 225, CW 0, parity 0` and CW log lines carrying
  `ICAM` — the iCAM path is the one being exercised, not a fallback.
* ECM round trip to the card server 214–282 ms.
* A three minute recording of an encrypted service: 313.6 MB, **0 encrypted
  packets**, 0 continuity errors across all 8 pids.
* A two minute recording of a second encrypted service: 67.2 MB, 0 encrypted
  packets, 0 continuity errors.
* `ad_ccerr 0`, `ad_decerr 0` throughout.

## Summary, and what I would suggest

What this establishes: the read past the end of the section is real and
provable, and the mode derivation demonstrably differs from the one OSCam used
to produce the control words — on 3 of 36 sections of a live feed. What it does
not establish: that any of it costs a user a single packet. On our provider it
does not.

Three things worth raising, and then a suggestion:

**1. Where do `b[2] - b[4] == 4` and `b[0x15]` come from?** This is the one
thing I cannot determine from the outside. If that arithmetic encodes a real
observation about some provider's ECM layout, I would rather know before you
merge; if it was a guess that happened to work, the patch is straightforwardly
right. All I can report is that on VideoGuard 0x098D it agrees for the 183 byte
sections and fails for the other two shapes.

**2. There is a cleaner way to get the CAID, if you want to go further.** The
patch treats the `k->caid == 0` startup window as possibly VideoGuard, because
`k->caid` is only filled in when OSCam answers with `DVBAPI_ECM_INFO`. But
minisatip already knows the answer: it parses the CA descriptors into
`pmt->ca[i]->id` (caid) and `pmt->ca[i]->pid` (the ECM pid), and OSCam's
`SET_FILTER` names exactly that pid. Seeding `k->caid` there would make the
VideoGuard test exact from the first ECM and drop the startup case entirely.
Happy to send that on top if you like the direction.

**3. A build assertion for the other patch in this series.** The C linkage bug
is invisible at build time and produces a working binary with iCAM silently
absent. One line catches it:

```console
nm -D minisatip | grep -q ' w dvbcsa_bs_key_set_ecm$'
```

So my suggestion is to take it whole rather than split it: the offset is
unexplained magic, but I ran both derivations against the same live VideoGuard
feed and neither dropped a single packet, so the change is measurably not a
regression here — and I do not think splitting it buys enough additional
confidence to be worth the round trip.

Jalle19:
I still don't see what this fixes. Sky Deutschland descrambling works fine with the current code.

dnehring7:
> I still don't see what this fixes. Sky Deutschland descrambling works fine with the current code.

@Jalle19 , read the issue. (1) is a programming bug -> fixed by the patch. (2) is a guard for a possible programming bug. It does not hurt any functionality, but make the code more correct.

Jalle19:
The out of bounds check can be added, sure, but I'm not convinced the check for iCAM is wrong here and OScam is correct.

dnehring7:
@Jalle19 — fair point. Let us split it: the bounds check and the length clamp go
into their own PR, the mode derivation stays separate.

I should be clear about what I can and cannot show.

I cannot prove the current derivation is wrong. I built minisatip with the mode
patch reverted, kept the C linkage fix so iCAM stays active, and captured a
VideoGuard service for 90 seconds: 0 encrypted packets, `ad_decerr 0`. Both
derivations descramble my provider without losing a packet. No regression, but no
visible benefit either.

What I did measure: 36 ECM sections on a live feed, both expressions computed side
by side in `send_ecm()`.

| count | assembled | `b[2]-b[4]==4` | current code | `get_ecm_mode()` | last byte |
|------:|----------:|:--------------:|-------------:|-----------------:|----------:|
| 33 | 183 | yes | 4 (from `b[0x15]`) | 4 | 0xE4 |
| 2 | 382 | no | 0 — iCAM off | 4 | 0xB4 |
| 1 | 30 | no | 0 — iCAM off | 6 | 0x46 |

They disagree on 3 of 36. For the common 183 byte section the fixed offset lands
on a byte that carries the same value as OSCam's rule. That is probably why the
current code works at all.

My argument is not that OSCam is always right. It is this: OSCam computes the mode
in `get_ecm_mode()`, and OSCam produces the control words we key the descrambler
with. If both sides derive the mode differently, they disagree about the key
schedule. On my provider that does no harm, because they agree for the section
shape that matters.

The one thing I cannot answer from outside: where do `b[2] - b[4] == 4` and
`b[0x15]` come from? If that encodes a real observation about some provider's ECM
layout, then my patch is wrong and the code should carry a comment saying so. If
it was a guess that happened to work, then following the rule OSCam uses is the
safer default. You or @catalinii probably know which it is.

So: bounds check now, mode change separately and without hurry.

dnehring7:
@Jalle19 — done. The PR now contains only the bounds checks, rebased on current
master. `b[2] - b[4] == 4` and `b[0x15]` are untouched, the mode derivation is out.

One correction to my own claim while I was rewriting it. `f->data` is
`FILTER_PACKET_SIZE + 2` = 4002 bytes. So for an ECM filter, reading `b[0x15]` of
a short section reads a leftover byte from the previous section in the same
filter — stale data, not a read off the end. My ASan reproducer allocated each
section to its exact length, which made it look worse than it is here.

Where it can really leave the buffer: `send_ecm()` is also registered for EMM
filters (`FILTER_ADD_REMOVE | isEMM`). There `b` is `f->data + i` for a sub-section
inside the buffer, so for a sub-section near the end the same arithmetic reads past
`f->data`.

A well formed ECM is not affected. On my live VideoGuard feed the header derived
length equalled the assembled length in all 36 captured sections, so `len` stays
the same and the same branch is taken as before.

On the mode itself: I have no evidence that the current derivation costs anyone a
packet, so it stays out. I keep the measurements if it ever becomes interesting.

src/dvbapi.cpp
```diff
@@ -651,10 +652,16 @@ int send_ecm(int filter_id, unsigned char *b, int len, void *opaque) {
...
 
-    len = ((b[1] & 0xF) << 8) + b[2];
-    len += 3;
+    // The length from the section header must not be taken past what the
+    // filter layer assembled: b is only len bytes long, and len bounds the
+    // memcpy() into buf below that forwards the ECM to OSCam.
+    section_len = ((b[1] & 0xF) << 8) + b[2] + 3;
+    if (section_len < len)
+        len = section_len;
+
```
Jalle19: This part seems completely unnecessary, the value of `len` was always overwritten before.

dnehring7: @Jalle19 — you are right, and I should have traced the callers before writing
that part. Both paths pass in exactly the length the header says:

* `assemble_normal()` returns `f->len`, which it computes from `f->data[1]` and
  `f->data[2]` — the same two bytes `send_ecm()` reads. Identical by construction.
* the EMM loop in `process_filter()` passes `cl + 3` with `cl` from
  `data[i+1]`, `data[i+2]`, and `b` is `data + i`. Again the same bytes.

So `section_len` could never be smaller than `len`, and the clamp was dead code.
Removed, and the sentence about the over-long `memcpy()` is out of the
description too — that cannot happen either.

Jalle19 (approved):

src/dvbapi.cpp
```diff
@@ -654,7 +654,8 @@ int send_ecm(int filter_id, unsigned char *b, int len, void *opaque) {
     len = ((b[1] & 0xF) << 8) + b[2];
     len += 3;
     k->last_ecm = getTick();
-    if (b[2] - b[4] == 4) {
+    // b[4] and b[0x15] are only inside the section if it is that long
```
Jalle19: IMO drop the comment

dnehring7: Comment dropped; the guard is self-explanatory.

Jalle19 (approved):

PR #1424, opened 2026-08-23
fix(adapter): resize the adapter buffers when -b is parsed

dnehring7:
Updated after @Jalle19's and @lars18th's review: the resize is now called from the
`-b` handler instead of from `main()`, and the three places that worked on
`ad->buf` through the global option use `ad->lbuf`.

## Problem

Options such as `--delsys`, `--diseqc`, `--lnb`, `--unicable` or `--slave`
allocate an adapter while the command line is still being parsed. That adapter
gets the buffer size that was current at that moment:

```c
ad->lbuf = opts.adapter_buffer;
ad->buf = malloc(ad->lbuf + 10);
```

A `-b` that comes *after* it only raises `opts.adapter_buffer`. The adapter keeps
the smaller allocation, while the code that works on it uses the global option:

```c
memset(ad->buf, 0, opts.adapter_buffer + 1);               // adapter_set_dvr()
set_socket_buffer(ad->sock, ad->buf, opts.adapter_buffer); //      "
memset(ad->buf, 0, opts.adapter_buffer + 1);               // init_hw()
```

The second one is worse than the memsets: `set_socket_buffer()` stores the length
in `ss->lbuf`, and the DVR read path uses that as its limit
(`pos_len = master->lbuf - master->rlen`), so the reads themselves write past the
allocation.

```
minisatip --delsys 0:dvbs2 -b 1000000:1000000
```

segfaults at startup, or under ASan reports a heap-buffer-overflow WRITE in
`init_hw()`.

## Fix

Resize the adapters that already exist in the `-b` handler; the ones allocated
after it get the size from `adapter_alloc()` as before.

With that in place `ad->lbuf` is the buffer length in every case, so the three
sites above now use it and there is one source of truth for the size of `ad->buf`.

## Testing

Fedora 44, gcc 16, ASan build, `ctest` 13/13.

| command line | result |
|---|---|
| `--delsys 0:dvbs2 -b 1000000:1000000` | `adapter 0: buffer resized from 376000 to 999972 bytes` |
| `--delsys 0:dvbs2 -b 10000:10000` | `adapter 0: buffer resized from 376000 to 9964 bytes` |
| `-b 1000000:1000000 --delsys 0:dvbs2` | no resize — allocated at the right size |
| `--delsys 0:dvbs2` | no resize |

Live: OSCam over dvbapi, Digital Devices DVB-S2, VDR over SAT>IP — unchanged.

### DIFF
--- a/src/adapter.cpp
+++ b/src/adapter.cpp
@@ -119,6 +119,29 @@ adapter *adapter_alloc() {
     return ad;
 }
 
+// Adapters allocated by an option that came before -b still carry the
+// previous buffer size.
+void adapter_resize_buffers() {
+    int i;
+    for (i = 0; i < MAX_ADAPTERS; i++) {
+        adapter *ad = a[i];
+        unsigned char *nb;
+        if (!ad || !ad->buf || ad->lbuf == opts.adapter_buffer)
+            continue;
+        nb = (unsigned char *)realloc(ad->buf, opts.adapter_buffer + 10);
+        if (!nb) {
+            LOG("could not resize the buffer of adapter %d from %d to %d "
+                "bytes",
+                i, ad->lbuf, opts.adapter_buffer);
+            continue;
+        }
+        LOG("adapter %d: buffer resized from %d to %d bytes", i, ad->lbuf,
+            opts.adapter_buffer);
+        ad->buf = nb;
+        ad->lbuf = opts.adapter_buffer;
+    }
+}
+
 void find_adapters() {
     static int init_find_adapter;
     if (init_find_adapter == 1)
@@ -212,8 +235,8 @@ void adapter_set_dvr(adapter *ad) {
         sockets_add(ad->dvr, NULL, ad->id, TYPE_DVR, (socket_action)read_dmx,
                     (socket_action)close_adapter_for_socket,
                     (socket_action)adapter_timeout);
-    memset(ad->buf, 0, opts.adapter_buffer + 1);
-    set_socket_buffer(ad->sock, (unsigned char *)ad->buf, opts.adapter_buffer);
+    memset(ad->buf, 0, ad->lbuf + 1);
+    set_socket_buffer(ad->sock, (unsigned char *)ad->buf, ad->lbuf);
     if (ad->adapter_timeout > 0)
         sockets_timeout(ad->sock, ad->adapter_timeout);
     if (opts.th_priority > 0)
@@ -286,7 +309,7 @@ int init_hw(int i) {
                        "buffer was not initialized",
                        ad->id);
     }
-    memset(ad->buf, 0, opts.adapter_buffer + 1);
+    memset(ad->buf, 0, ad->lbuf + 1);
     ad->tp.clear();
     mark_pids_deleted(i, PID_STREAM_ID_UNDEFINED, NULL);
     update_pids(i);
--- a/src/minisatip.cpp
+++ b/src/minisatip.cpp
@@ -802,6 +802,7 @@ void set_options(int argc, char *argv[]) {
                 opts.adapter_buffer = MAX_UDP_PACKET_SIZE;
             if (opts.dvr_buffer == 0)
                 opts.dvr_buffer = DVR_BUFFER;
+            adapter_resize_buffers();
 
             break;
         }
--- a/src/adapter.h
+++ b/src/adapter.h
@@ -169,6 +169,7 @@ int init_hw(int i);
 int init_all_hw();
 int getAdaptersCount();
 adapter *adapter_alloc();
+void adapter_resize_buffers();
 int close_adapter(int na);
 int get_free_adapter(transponder *tp);
 int set_adapter_for_stream(int sid, int aid);


Jalle19:
Seems like a better way would be to somehow ensure the buffer size is parsed before we call adapter_alloc(). That way we don't need to resize anything.

dnehring7:
@Jalle19 — you are right in principle. If the buffer size were known before the
first `adapter_alloc()`, there would be nothing to resize.

But the order is only the trigger. `ad->buf` is allocated with `ad->lbuf + 10`,
and three places use the global `opts.adapter_buffer` on that buffer instead of
`ad->lbuf`:

```c
memset(ad->buf, 0, opts.adapter_buffer + 1);                                /* adapter.cpp:215 */
set_socket_buffer(ad->sock, (unsigned char *)ad->buf, opts.adapter_buffer); /* adapter.cpp:216 */
memset(ad->buf, 0, opts.adapter_buffer + 1);                                /* adapter.cpp:289 */
```

The second one is the real problem. It sets `ss->buf = ad->buf` and
`ss->lbuf = opts.adapter_buffer`, and the read path uses `ss->lbuf` as the limit
(`pos_len = master->lbuf - master->rlen`). That is the DVR read path writing past
the allocation, not only a memset at startup.

The correct parse order makes both values equal, so nothing happens. But two
sources of truth for one buffer remain.

One more point for the decision: `adapter_alloc()` copies 19 `opts.*` fields, and
eleven option handlers allocate adapters — `--delsys`, `--lnb`, `--diseqc`,
`--unicable`, `--slave`, `--adapter-timeout` and others. A first pass only for
`-b` fixes this crash. A first pass that really removes the ordering problem must
cover everything `adapter_alloc()` reads.

Four possibilities:

1. This PR as it is.
2. This PR plus `ad->lbuf` at the three places. My preference, because it fixes
   the cause and not the trigger.
3. Only the `ad->lbuf` fix. Smallest diff, no crash, but an adapter allocated
   during parsing keeps the default buffer instead of the one from `-b`.
4. Two-pass parsing instead of the resize. Also fine, but bigger than it looks.

Tell me which one and I will push it.

lars18th:
Hi @dnehring7 ,

> Tell me which one and I will push it.

@catalinii and @Jalle19 are the maintainers. However, my recommendation is that:
- The best option is 2 with a small modification: change the name of the function from `adapter_resize_buffers()` to `check_adapters_buffers()`. This is more clear because you're checking the correct values at every loop of the parameters processing, and only doing resizes if it's necessary.

You agree?

Jalle19:
Maybe the existing resize function is fine, we could call it when parsing the -b option.

dnehring7:
@Jalle19 @lars18th — done, pushed.

@Jalle19: the `-b` handler is the right place. `opts.adapter_buffer` is written in
exactly two spots, the default in `set_options()` and that handler, so calling the
resize there covers everything: adapters that already exist get corrected,
adapters allocated afterwards get the size from `adapter_alloc()` anyway. The call
in `main()` is gone.

@lars18th: I kept the name `adapter_resize_buffers()`. `check_adapters_buffers()`
fits a function called on every pass of the parse loop — with the call sitting in
the `-b` handler it runs once, at the one moment the size can change, and then
"resize" is what it actually does. The `ad->lbuf == opts.adapter_buffer` early-out
stays, so the no-op case costs nothing.

The second half of your option 2 is in: `adapter_set_dvr()` and `init_hw()` worked
on `ad->buf` through the global option, and `set_socket_buffer()` even hands that
length to the read path as `ss->lbuf`. All three use `ad->lbuf` now, so there is
one source of truth for the size of that buffer.

Tested in all four orders:

```
--delsys 0:dvbs2 -b 1000000:1000000   -> adapter 0: buffer resized from 376000 to 999972 bytes
--delsys 0:dvbs2 -b 10000:10000       -> adapter 0: buffer resized from 376000 to 9964 bytes
-b 1000000:1000000 --delsys 0:dvbs2   -> no resize, allocated at the right size
--delsys 0:dvbs2                      -> no resize
```

Jalle19 (approved):

lars18th:
Ready to merge it?

src/minisatip.cpp
Jalle19, line 801: This should probably be ">" BTW?

src/adapter.cpp
```diff
@@ -119,6 +119,32 @@ adapter *adapter_alloc() {
     return ad;
 }
 
+// Options such as --delsys, --diseqc or --virtual-diseqc allocate an adapter
```
Jalle19: IMO drop the whole comment

Jalle19 (changes requested):

src/adapter.cpp
lars18th: I recommend maintaining relevant comment inside the code. This will make easier in the future to do modifications.

Jalle19: It's way too long and not super relevant

src/minisatip.cpp
dnehring7: `<` is right — it is a floor, not a cap.

`MAX_UDP_PACKET_SIZE` is 1316 (7 * 188), the default `ADAPTER_BUFFER` is
2000 * 188 = 376000. With `>` every buffer would be clamped down to 1316, the
default included.

The floor is needed because the buffer is filled in whole datagrams:

```c
while (total + MAX_UDP_PACKET_SIZE <= len)   // satipc.cpp, len = adapter buffer
```

Below 1316 that loop never runs and the adapter reads nothing. 1316 is also a
multiple of 188, so it survives the `(x / 188) * 188` rounding above it.

`-b 100:100` ends up at 1316 here, which is the intended behaviour.

src/adapter.cpp
dnehring7: Shortened to two lines, at the function:

```c
// Adapters allocated by an option that came before -b still carry the
// previous buffer size.
void adapter_resize_buffers() {
```

The comment at the call site in the `-b` handler is gone.

@Jalle19 @lars18th — I hope that is the middle ground: the *why* is one sentence
and stays, the rest was the commit message in the wrong place.

Jalle19 (approved):

lars18th:
Thank you @dnehring7 for providing this patch! 👍

PR #1425, opened 2026-08-23
feat(pmt): add --clean-psi to strip CA descriptors from descrambled PMTs

dnehring7:
A service minisatip descrambles reaches the client in the clear, but its PMT
still carries the CA descriptors of the broadcast, so a client that decides from
the PMT keeps treating it as encrypted. VDR refuses such a channel on a SAT>IP
device and drops a running recording when the CA descriptors change:

```
ZDF HD (FTA)         -> 250 2 ZDF HD
RTL HD (descrambled) -> 554 Error switching to channel "3"
```

`--clean-psi` hands the client the broadcast PMT section minus its CA
descriptors. Every stream and every other descriptor stays as sent; only the
lengths, the version and the CRC change. It is built once per PMT version from
the section `process_pmt()` already has, so the parser is not touched and this
no longer depends on #1428.

The rewrite waits until the service is known to be in the clear: a control word
that passed `test_decrypt_packet()`, or the PCR stream arriving unscrambled
because another SAT>IP server, a CAM or a hardware descrambler already did the
work. Until then the PMT pid is withheld for up to `--clean-psi=seconds` (3 by
default, 0 never), so a client never sees two PMTs for one service. The adapter
buffer itself is left alone; the replacement packets are picked per client.
Off by default.

Tested live on RTL HD (Nagra, 11 CAIDs, 7 streams), with a second minisatip as a
SAT>IP client of the first:

| | PMT the client gets |
|---|---|
| upstream descrambles, downstream `--clean-psi` | v1, 111 bytes, no CAIDs |
| tuner + OSCam, `--clean-psi` | v1, 111 bytes, no CAIDs |
| no card server | v0, 655 bytes, 11 CAIDs |
| upstream can't descramble, downstream relays | v0, 655 bytes, 11 CAIDs |

In both rewritten cases the section is byte for byte the broadcast one without
its CA descriptors, checked with a separate script. `ctest` 13/13 with and
without sanitizers, plus a minimal build.

### DIFF
--- a/tests/test_pmt.cpp
+++ b/tests/test_pmt.cpp
@@ -23,6 +23,7 @@
 #include "socketworks.h"
 #include "utils.h"
 #include "utils/testing.h"
+#include "utils/ticks.h"
 #include <arpa/inet.h>
 #include <ctype.h>
 #include <errno.h>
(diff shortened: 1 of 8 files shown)


Jalle19:
Tvheadend has an option to "clear scrambled bits", we should check how it does things in case there's a simpler solution than rewriting the PMT.

catalinii:
We just removed a similar feature few commits ago as it was unmaintained and not working.

The ultimate question is, who benefits from the clean pmt ?
Also what happens if a service becomes encrypted and unencrypted for example due to oscam issues. 
Also what happens before clearing the channel: do we send the PMT with the encrypted streams or wait for it to clear ? if we wait, what is the timeout.

Pretty much everyone assumes that the PMTs are pretty static and changing version rarely, this could cause a lot of issues downstream.

lars18th:
Hi @catalinii ,

IMHO the answer to these questions are quite simple:
- The parameter `--clean-pmt` requires _three_ modes: [on, off, disabled]. Where `disabled` is the current behavior and it will be the default. Therefore, this _new_ functionality doesn't hurt current users that don't want it.
- When `--clean-pmt on` the PMT will be untouched until the first descrambling will be done. After that it will be patched _updating_ the version value. And it will be untouched until a new version will arrive (and then it will be repatched). That implies that any decrypt error will not retouch the PMT. And this is consistent with the common behaviour: decoding errors are like receiving errors.

I hope @dnehring7 will want to consider these suggestions.
You agree?

dnehring7:
@catalinii @lars18th — thanks, and fair point about `clean_psi`; I should have
found it before opening this.

**Why this one is different.** `clean_psi` wrote its own continuity counter into
the TS header (`b[3] = (b[3] & 0xF0) | cpmt->clean_cc;`), which cannot match what
the client already received on that pid. This patch never touches a TS header at
all: removing descriptors only makes the section shorter, so it is written back
into the packets the original occupied and every header, continuity counter
included, stays exactly as broadcast.

**@catalinii, on PMT versions:** the client never sees a change. The pid is
withheld until we know whether we can descramble, so the first PMT it ever gets
is the final one, and it stays constant for as long as the broadcast version
does. Off by default.

**@lars18th:** your second point is exactly what the patch does — after the first
control word the PMT stays put, and later decrypt errors do not touch it again.
Your first point I built and had to drop: passing the original through and
swapping it on the first key makes VDR abort a running recording —

```
changing caids of channel 5 from 098C,098D,09F0 to
stopping recording due to modification of channel 5
```

Holding the pid avoids it, because the client never sees the first PMT.

**What I can push — pick any:**

1. Never change the version at all: emit the broadcast version, and bump only in the one case where the original already went out to a client.
2. Split the PR: the section builder plus its unit test, separately from the hold logic.
3. Rename to `--clean-psi`.
4. A no-hold mode as its own flag, if that is what @lars18th wants available.
5. Configurable grace window instead of a fixed 3 s.

If the answer is still no, that is fine — I will carry it downstream.

lars18th:
I agree with all, @dnehring7 !

> 1. Never change the version at all: emit the broadcast version, and bump only in the one case where the original already went out to a client.
> 2. Split the PR: the section builder plus its unit test, separately from the hold logic.
> 3. Rename to `--clean-psi`.
> 4. A no-hold mode as its own flag, if that is what @lars18th wants available.
> 5. Configurable grace window instead of a fixed 3 s.

However, one question: What happens inside VDR and TVHE when a broadcaster that uses encryption and FTA inside the same service changes from encrypted to clean mode? Several channels use this technique, and I feel the best way is to mimic this, right?

catalinii:
Can u guys answer the question around:
- who benefits from the clean pmt ?
- what happens if we do not do it ? (what is the downside, what does not work).

dnehring7:
@catalinii — short answers first.

**Who benefits?** Anyone running minisatip with `-o` and a client that decides
from the PMT whether a service is encrypted. Off by default, so nobody else is
affected.

**What does not work without it?** With VDR this is not about picture quality, it
is about the service being unusable. VDR takes the CAIDs from the PMT
(`cChannel::SetCaIds()`), so the channel counts as encrypted
(`Ca() >= CA_ENCRYPTED_MIN`). In `cDevice::GetDevice()`, if no CAM slot provides
those CAIDs, `InternalCamNeeded` is set (device.c:270) and every device without
`HasInternalCam()` is skipped (device.c:285). `cSatipDevice::HasInternalCam()` is
only true with the CI extension and a CI slot (satip device.c:529), which nobody
has in a minisatip + OSCam setup.

So VDR refuses the channel completely — no live, no recording — while minisatip is
delivering that same service fully in the clear. (VDR 2.8.2, satip plugin 2.5.2.)

Two workarounds, both bad:

* load a softcam plugin in VDR — works, but every packet is pushed through the CAM
  slot, and OSCam gets a second ECM session for a service it is already
  descrambling for minisatip;
* leave the PMT pid out of the SAT>IP request — live TV works, recording does not,
  VDR needs the PMT.

With `--clean-pmt` the service is simply an FTA channel, which is what it actually
is at that point.

**A correction to my own earlier comment in this thread:** I wrote that VDR hands
the clear stream to its softcam "which corrupts it". I have read the code since,
and that is not what happens — vdr-plugin-dvbapi checks the scrambling bits and
passes clear packets through untouched. It does route them through
`DeCSAFillControl`, which sleeps the receiver thread to batch packets for the
descrambler, so you pay latency for nothing. Not corruption. The device selection
above is the real problem.

---

@lars18th — good question, and the answer argues against mimicking it.

VDR does not handle that flip gracefully either. A changed CA descriptor list sets
`CHANNELMOD_CA` (channels.c:477), which is part of `CHANNELMOD_RETUNE`
(channels.h:29). The main loop (vdr.c:1070) then stops and restarts a running
recording ("stopping recording due to modification of channel", menu.c:5881) and
retunes the live channel. That is exactly what I measured when an earlier version
of this patch forwarded the original PMT and swapped it after the first key.

And as far as I have seen, broadcasters that mix FTA and encrypted inside one
service do not touch the PMT at all: the CA descriptors stay, only
`transport_scrambling_control` in the TS header goes to 00. That is why it works
for them — the PMT stays static. Mimicking them therefore means "never change the
PMT", which is what the hold achieves: the client sees one version and it stays
that way.

I cannot say the same about Tvheadend with any confidence, I have not read that
part.

---

On your five points: I will do 1 (emit the broadcast version, bump only if the
original already went out), 5 (configurable grace window) and the split from 2 as
soon as @catalinii says the feature is wanted at all.

lars18th:
Hi @dnehring7 ,

If VDR can't handle services with CA and FTA at the same time, when the broadcaster doesn't changes the PMT, then the trouble is inside VDR.

Anyway, IMHO this function of `--clean-psi` is a must-have. Some IPTV players that could use minisatip over HTTP requests refuses to play when the PMT describes encrypted streams. So as an optional functionality it will be useful. And regarding the implementation, I feel it's clear: configurable timeout window, sending the patched PMT only after starting the decryption and with the same version number; and if the version number updates, then repatch the new PMT. And if the timeout window is set to "0" then "wait forever until decrypted", and if "-1" don't wait at all and patch it even is not already decrypted. I feel this is the most powerful solution. You agree?

dnehring7:
Agreed. A default timeout of 3 seconds seems sensible:

- 3 (default): wait up to 3 seconds for decryption, then send the patched PMT.
- 0: wait indefinitely until decryption is active.
- -1: patch and send the PMT immediately, regardless of decryption status.

The patched PMT should retain the original version number, and any updated PMT version should be patched again accordingly.

Any objections? If not, I would apply these changes to the PR.

lars18th:
> Any objections? If not, I would apply these changes to the PR.

Only one comment: The paramater will be `--clean-psi XXX` where the XXX is an integer value with the timeout value? And when it's missing be 3 (default)?

Jalle19:
Can you check how tvheadend's "remove scrambled bits" option works? Does it also modify the PMT or does it do something simpler?

catalinii:
Should we attempt to fix VDR first and then figure out if something is left to do in minisatip?

To me is very clear that the thing that hurts the most is how vdr handles it.

I am not fan in general on hacking minisatip to go around vdr bugs.

I think VDR bug can be challenged on Encrypted channels going FTA temporarily.

dnehring7:
> Should we attempt to fix VDR first and then figure out if something is left to do in minisatip?
> 
> To me is very clear that the thing that hurts the most is how vdr handles it.
> 
> I am not fan in general on hacking minisatip to go around vdr bugs.
> 
> I think VDR bug can be challenged on Encrypted channels going FTA temporarily.

@catalinii — fair question, and I went and measured it instead of arguing. Short
version: **you are partly right. There is a client-side fix, it works today, and
I will say so plainly.** But it is not a VDR bug, and it is not free.

## What VDR does, exactly

VDR takes the CAIDs from the PMT (`cChannel::SetCaIds()`, channels.c:471), so the
channel counts as encrypted (`Ca() >= CA_ENCRYPTED_MIN`). In
`cDevice::GetDevice()`:

* no CAM slot provides those CAIDs → `InternalCamNeeded = true` (device.c:270)
* every device without `HasInternalCam()` is then skipped (device.c:285)
* `cSatipDevice::HasInternalCam()` is `GetCIExtension() && ciSlot > 0` (satip
  device.c:529), and `ciExtensionM` defaults to 0

Measured on VDR 2.8.2 with satip 2.5.2, SAT>IP as the only device, no softcam
loaded, minisatip descrambling the service:

```
ZDF HD  (FTA)          -> 250 2 ZDF HD
RTL HD  (descrambled)  -> 554 Error switching to channel "3"
                          info: Kanal nicht verfügbar!
```

This is not a bug. VDR is being told "this service is encrypted" and correctly
answers "then I need something that can decrypt it". The information it is given
is simply wrong: the stream is clear.

## Can it be fixed on the client side? Yes — I tested it

Give the satip plugin `-c 0x098C,0x098D,0x09F0`, set
`satip.EnableCIExtension = 1` and `satip.CIAssignedDevice = 1`. That makes
`HasInternalCam()` true, VDR stops looking for a CAM slot, and the channel plays
with `--clean-psi` **off**:

```
250 3 RTL HD
vaapivideo: video codec h264 opened
vaapivideo: audio codec ac3 opened
```

So the honest answer to "should we fix VDR first" is: nothing needs fixing in VDR
core. What exists is a satip-plugin configuration that means telling VDR the
server has an OctopusNet CI slot it does not have. The plugin then appends
`&x_pmt=...&x_ci=1` to the RTSP SETUP (satip device.c:432), which minisatip
ignores. The CAID list is maintained by hand per provider, and it fixes exactly
one client.

## On "encrypted channels going FTA temporarily"

I checked that too, because @lars18th asked the same thing. VDR does not handle it
gracefully: a changed CA descriptor list sets `CHANNELMOD_CA` (channels.c:477),
which is part of `CHANNELMOD_RETUNE` (channels.h:29), and the main loop
(vdr.c:1070) then stops and restarts a running recording ("stopping recording due
to modification of channel", menu.c:5881) and retunes live viewing.

But that is not what those broadcasters do. As far as I have seen they leave the
PMT alone and only clear `transport_scrambling_control` in the TS header. The PMT
stays static, so VDR never sees a channel modification. Challenging VDR on that
would be challenging it on something broadcasters do not actually do.

## Where that leaves the PR

`--clean-psi` says the true thing — this service is not encrypted any more — in
the one place every client already looks, instead of each client being told
separately to ignore what the PMT says. Off by default, one comparison per
adapter buffer when off.

If you would still rather this lived in the clients, I accept that; the
measurements are in the PR for whoever needs them, and I will carry it
downstream. But I did not want to argue for it on a claim I had not tested.

dnehring7:
> Can you check how tvheadend's "remove scrambled bits" option works? Does it also modify the PMT or does it do something simpler?

@Jalle19 — I read it. It is the simpler thing, and it does not touch the PMT.

"Remove scrambled bits" is an input-level option
(`mpegts_input.c:325`, default on, `PO_EXPERT`). All it does is clear the
`transport_scrambling_control` bits in the **TS header** of every packet
(`mpegts_input.c:1275`):

```c
if (mi->mi_remove_scrambled_bits || (flags & MPEGTS_DATA_REMOVE_SCRAMBLED) != 0) {
  uint8_t *tmp, *end;
  for (tmp = mp->mp_data, end = mp->mp_data + len2; tmp < end; tmp += 188)
    tmp[3] &= ~0xc0;
}
```

Its own description says it: "a workaround for the special streams which are
descrambled, but these bits are not touched."

minisatip already does the equivalent unconditionally in `decrypt_batch()`:

```c
for (i = 0; i < pmt->blen; i++)
    pmt->batch[i].data[3] &= 0x3F;
```

So that half is not missing, and this PR is not a duplicate of it — the two act on
different things. Clearing the scrambling bits fixes clients that decide from the
TS header. It does nothing for a client that decides from the CA descriptors in
the PMT, which is what VDR does.

Tvheadend itself is tolerant here, which is worth saying: in
`descrambler_descramble()` an unscrambled packet with an empty queue goes
straight to `ts_recv_packet0()` (descrambler.c:1077), so a CA-carrying PMT on an
already-clear stream does not stop it. I have not run that as a live test, only
read it — the live measurements in this PR are all VDR.

dnehring7:
> > Any objections? If not, I would apply these changes to the PR.
> 
> Only one comment: The paramater will be `--clean-psi XXX` where the XXX is an integer value with the timeout value? And when it's missing be 3 (default)?

@lars18th — yes to all of it, and it is pushed.

**The parameter.** `--clean-psi[=seconds]`, so:

* `--clean-psi` — 3 s, the previous fixed value
* `--clean-psi=10` — 10 s
* `--clean-psi=0` — never withhold the pid

It is a GNU `optional_argument`, so the value needs the `=`; `--clean-psi 5`
would be read as `--clean-psi` followed by a stray argument.

`--clean-psi=0` is your point 1 no-hold mode without a second flag: the broadcast
PMT goes out first and is replaced when the first key arrives. Measured:

```
--clean-psi     -> one section, the rewritten one, version 0
--clean-psi=0   -> broadcast PMT version 0, then rewritten version 1
```

**The version.** Emitting the broadcast version turned out to be the more
demanding of your suggestions, and it found a real bug. The window used to start
when the client asked for the pid — but a client asks *before* the adapter has
locked, and a cold tune here took 3064 ms. The window was spent on tuning, so on
a fresh tune the broadcast PMT went out before anything had been classified, and
the client got two PMTs. With the old always-bump behaviour that was survivable;
with the broadcast version kept, both would have carried version 0 and a client
caching by version would have kept the wrong one.

The window now starts at the first buffer after the request. Three consecutive
fresh tunes now deliver exactly one section. So thank you — that suggestion is
what exposed it.

**Your question about a service that goes FTA temporarily.** I looked at VDR for
this. It does not handle it gracefully: a changed CA descriptor list sets
`CHANNELMOD_CA` (channels.c:477), part of `CHANNELMOD_RETUNE` (channels.h:29),
and the main loop (vdr.c:1070) stops and restarts a running recording and
retunes live viewing.

But I do not think those broadcasters actually change the PMT. As far as I have
seen the CA descriptors stay and only `transport_scrambling_control` in the TS
header goes to 00 — which is why it works for them, the PMT stays static. So
mimicking them means "never change the PMT", which is what the hold gives us: the
client sees one version and it stays.

Still open from your list: splitting the PR. I would rather wait for @catalinii
to say whether he wants the feature at all before restructuring it.

lars18th:
Hi @dnehring7 ,

My preference is to include this functionality. Perhaps not all users will use it, but in _some_ cases it will be useful.

After, that opinion I suggest this in reference to the version control:
- Why not add the option `--clean-psi=~` too? This say _patch all PMT in all channels_. Obviously this could generate troubles because not all channels will be decrypted. However, if the user needs it (for example with a central minisatip serving HTTP clients) then this guaranties that all received PMT will describe FTA services. This will solve the problem entirely with the only problem that if you tune an encrypted service you will receive the encoded stream if you can't open it.

What you think about this?

Jalle19:
VDR should be changed to use the same mechanism as tvheadend does (check for scramblong bit). Then no changes to minisatip would be needed.

Not sure what is really best practice here.

lars18th:
Hi @Jalle19 ,

> VDR should be changed to use the same mechanism as tvheadend does (check for scramblong bit). Then no changes to minisatip would be needed.

This seems correct to me: VDR needs some improvement.

However, not only VDR or TVHE are the clients to serve from minisatip. I accept that a lot of users have an environment like this:  one or more STBs with minisatip as the SAT>IP server; and VDR/TVHE as TV clients or central managers. But this is not true for all users. One example of another common environment is: STB+minisatip (without decryption) serving to one central minisatip (using satipc and with physical CI or oscam) serving to multiple HTTP clients (different Apps in AndroidTV and smartphones). So, in this specific environment for the _central_ minisatip instance the use of the `--clean-psi=~` will be _very_ useful.

Therefore, IMHO as the proposed PR doesn't change the current behavior and it's off by default, it provides an optional functionality that is valuable for some users.

dnehring7:
> Hi @dnehring7 ,
> 
> My preference is to include this functionality. Perhaps not all users will use it, but in _some_ cases it will be useful.
> 
> After, that opinion I suggest this in reference to the version control:
> 
> * Why not add the option `--clean-psi=~` too? This say _patch all PMT in all channels_. Obviously this could generate troubles because not all channels will be decrypted. However, if the user needs it (for example with a central minisatip serving HTTP clients) then this guaranties that all received PMT will describe FTA services. This will solve the problem entirely with the only problem that if you tune an encrypted service you will receive the encoded stream if you can't open it.
> 
> What you think about this?

@lars18th — thank you, and it is implemented and pushed.

I called it `--clean-psi=all` rather than `=~`, because the argument is otherwise
a number of seconds and a word says what it does. Same meaning as your `~`:

```
--clean-psi        rewrite the PMT of services a control word really decrypted
--clean-psi=10     the same, with a 10 s window
--clean-psi=0      the same, but never withhold the pid
--clean-psi=all    rewrite the PMT of every encrypted service
```

`all` drops the control word condition, so a client that cannot do anything
sensible with a CA descriptor never sees one. Measured on the same service, with
and without a card server configured, so both halves of the trade are visible:

| card server | mode | PMT the client gets | video packets scrambled |
|---|---|---|---|
| yes | `--clean-psi` | rewritten, 108 B, version 0 | 0 / 14740 |
| yes | `--clean-psi=all` | rewritten, 108 B, version 0 | 0 / 14740 |
| no | `--clean-psi` | broadcast, 652 B, 11 CAIDs | 38717 / 38717 |
| no | `--clean-psi=all` | rewritten, 108 B, version 0 | 14783 / 14783 |

The last row is exactly what you described: the service is announced free to air
and arrives scrambled. That is the mode doing its job, and it is why it stays
opt-in and why the help text says it in as many words.

One detail worth knowing: `all` does not skip the hold. The PMT still cannot be
rebuilt before it has been parsed, so the pid is withheld until it is — which is
one section, not an ECM round trip. The client still sees exactly one PMT, with
the broadcast version. Verified: one section per fresh tune in that mode too.

Your environment argument is the one I would put in front of @catalinii, not
mine. My VDR case has a workaround — configuring the satip plugin's CI extension
makes VDR accept the service with the option off, I tested it. A central
minisatip feeding player apps over HTTP has no client to configure.

dnehring7:
> VDR should be changed to use the same mechanism as tvheadend does (check for scramblong bit). Then no changes to minisatip would be needed.
> 
> Not sure what is really best practice here.

@Jalle19 — you are right about the mechanism, and I want to be precise about why
it does not transfer to VDR, because I checked both.

Tvheadend's gate really is the scrambling bit, in `ts_recv_packet1()`:

```c
if(!t->s_scrambled_pass && ((tsb[3] & 0xc0) || scrambled)) {   /* tsdemux.c:192 */
```

`service_is_encrypted()` — which is the one that reads the CA descriptors out of
the PMT — is only used inside that branch to latch `s_scrambled_seen` and to
raise `TSS_NO_DESCRAMBLER`. It never stops the stream. A clear packet takes the
`else` branch and is delivered, whatever the PMT says.

The difference is not which bit is checked, it is **when**. Tvheadend's check
runs on packets that are already arriving. VDR's runs in
`cDevice::GetDevice()` — before a device is chosen, before tuning, before a
single packet exists. At that moment there is no scrambling bit to look at. That
is why:

```
ZDF HD  (FTA)          -> 250 2 ZDF HD
RTL HD  (descrambled)  -> 554 Error switching to channel "3"
                          channel not available!
```

with SAT>IP as the only device and no softcam, while minisatip is delivering RTL
HD in the clear.

And tvheadend has no equivalent of that stage at all: it does not refuse to start
a service because no CA client can handle it — it starts it, streams it, and sets
a status flag if descrambling fails. So "make VDR do what tvheadend does" is not a
matter of swapping a check; it means removing a device selection rule that exists
for a good reason. With a real CAM, VDR *has* to pick a device that has the right
one.

What VDR already has for exactly this case is `HasInternalCam()` — "this device
delivers decrypted data by itself". I tested it: give the satip plugin
`-c 0x098C,0x098D,0x09F0` with `satip.EnableCIExtension = 1` and
`satip.CIAssignedDevice = 1`, and RTL HD plays with `--clean-psi` off, h264 and
ac3 decoding. So the VDR-side fix exists today — it just means telling VDR the
server has an OctopusNet CI slot it does not have, and maintaining the CAID list
by hand.

The clean version would be a small option in vdr-satip, something like "the
server descrambles", that sets `HasInternalCam()` without pretending to be CI
hardware. That is a handful of lines and I am happy to write it and send it to
the satip plugin.

But it fixes VDR. @lars18th's second environment has no client to fix: a central
minisatip feeding player apps over HTTP. That is the case I would judge this PR
on, not mine — mine has a workaround, his does not.

Hopefully now all are happy with the proposal, no default behaviour changes, and the changes do not depend on special client side handling. It should work for all situations we can construct.

lars18th:
Hi @dnehring7 ,

Thank you for the effort with this PR. Regarding the VDR patch, I agree that it will be necessary despite this new `--clean-psi` parameter. So before committing this patch (I hope @Jalle19 and @catalinii will accept it) I have another comment:

> One detail worth knowing: all does not skip the hold. The PMT still cannot be
> rebuilt before it has been parsed, so the pid is withheld until it is — which is
> one section, not an ECM round trip. The client still sees exactly one PMT, with
> the broadcast version. Verified: one section per fresh tune in that mode too.

To overcome this problem I suggest to also add the option `--clean-psi=all-secure` that means: block all pids until the first PMT is read and patched. This will filter out all pids (or all except 0-21) until you are sure that the PMT is received, validated and patched. So after that, all is ready to send to the client that will receive consistent data.

You agree with that? In that case, please don't hurt the case `-E -E -k` that sends all packets. The idea is only to filter out the _initial_ packets until you read the PMT.

Jalle19:
> VDR's runs in
cDevice::GetDevice() — before a device is chosen, before tuning, before a
single packet exists. 

I don't buy this.

Jalle19:
We had something similar to this and it was removed in https://github.com/catalinii/minisatip/commit/b0463d28eaea3b3893449246097ad6fa1a858a16

lars18th:
> We had something similar to this and it was removed in [b0463d2](https://github.com/catalinii/minisatip/commit/b0463d28eaea3b3893449246097ad6fa1a858a16)

I remember it (in 2023)! And it was introduced in 2015 with: 7121e766dbda182eb8bf98d63c0d51a2489a0292

However, this was removed because the implementation fails. But the implementation from @dnehring7 seems robust.
IMHO this PR could be approved when finished (I hope the option `--clean-psi=all-secure` will be implemented too).

dnehring7:
> We had something similar to this and it was removed in [b0463d2](https://github.com/catalinii/minisatip/commit/b0463d28eaea3b3893449246097ad6fa1a858a16)



> > VDR's runs in
> > cDevice::GetDevice() — before a device is chosen, before tuning, before a
> > single packet exists.
> 
> I don't buy this.

@Jalle19 — fair, "I don't buy this" deserves evidence rather than a description.
So here it is from both ends.

## The call chain

`cDevice::SwitchChannel()` → `cDevice::SetChannel()` (device.c:879):

```c
cDevice *Device = (LiveView && IsPrimaryDevice(false)) ? GetDevice(Channel, LIVEPRIORITY, true) : this;   // :897
...
if (NeedsTransferMode) {
   if (Device && PrimaryDevice()->CanReplay()) { ... }
   else
      Result = scrNotAvailable;                                                                           // :917
   }
else {
   ...
   if (SetChannelDevice(Channel, LiveView)) {                                                             // :930  <- the tune
```

`GetDevice()` is line 897, the tune is line 930, and line 917 returns
`scrNotAvailable` in between. Inside `GetDevice()`:

```c
if (Channel->Ca() >= CA_ENCRYPTED_MIN) { ...                     // device.c:257
   if (!NumUsableSlots) InternalCamNeeded = true;                // device.c:270
...
if (InternalCamNeeded && !HasInternalCam) continue;              // device.c:285
```

If nothing qualifies, `GetDevice()` returns NULL and `SetChannelDevice()` is
never called. No tuner is programmed, so there is no transport stream, so there
is no scrambling bit to look at.

## On the wire

Better than reading it — I put tcpdump on the RTSP link between VDR and minisatip.
SAT>IP as the only device, no softcam, VDR sitting on ZDF HD (11361), then told to
switch to RTL HD (**10832**), the service minisatip is descrambling:

```
15:48:18.538  PLAY     rtsp://.../stream=1?delpids=6110,6120,6121,6122,6123,6131
15:48:18.551  TEARDOWN rtsp://.../stream=1
15:48:19.483  OPTIONS  rtsp://.../
15:48:19.483  SETUP    rtsp://.../?src=1&freq=11361&pol=h&ro=0.35&msys=dvbs2&mtype=8psk&sr=22000&fec=23
15:48:19.484  PLAY     rtsp://.../stream=1?pids=none
15:48:19.485  DESCRIBE rtsp://.../stream=1
15:48:19.783  PLAY     rtsp://.../stream=1?addpids=0,16,17,18,20,6110,6120,6121,6122,6123,6131
```

SVDRP answered `554 Error switching to channel "3"`.

Every request in that window is **11361** — the transponder it was already on,
which it drops and re-establishes as it falls back. The string `10832` does not
appear anywhere in the capture: `grep -c 10832` → **0**. The server was never
asked to tune the encrypted transponder.

That is the whole claim. VDR did not tune, receive a packet, look at a scrambling
bit and give up. It decided from `Channel->Ca()` and never tuned.

For contrast, the FTA channel on the same setup produces real
`PLAY ...?addpids=...` traffic and `250 2 ZDF HD`.

## On b0463d2

Yes, and I should have found `clean_psi` myself before opening this — I went
through it in an earlier comment here. The short version: it wrote its own
continuity counter into the TS header, used fixed-size copies that ignored the
adaptation field, corrupted the version field so `current_next_indicator` ended up
0, dropped every program-level descriptor rather than only the CA ones, and had
no bounds checking. This patch writes the rewritten section into the packets the
original occupied, so no TS header is touched at all — 0 continuity errors on the
PMT pid across the rewritten packets here.

Same goal, and I am happy to be judged on whether this implementation is
maintainable. But "the last one did not work" is a statement about that code.

dnehring7:
> > We had something similar to this and it was removed in [b0463d2](https://github.com/catalinii/minisatip/commit/b0463d28eaea3b3893449246097ad6fa1a858a16)
> 
> I remember it (in 2023)! And it was introduced in 2015 with: [7121e76](https://github.com/catalinii/minisatip/commit/7121e766dbda182eb8bf98d63c0d51a2489a0292)
> 
> However, this was removed because the implementation fails. But the implementation from @dnehring7 seems robust. IMHO this PR could be approved when finished (I hope the option `--clean-psi=all-secure` will be implemented too).

@lars18th — thank you for the support, and I owe you a straight answer on
`all-secure`: **I built it, measured it, and it does not do what you want at this
layer. So I took it back out.** Here is the reasoning, because I think the idea
is right and only the place is wrong.

## First, you are right that there is something to fix

I measured how much of a service reaches the client before its first complete
PMT, on three fresh tunes:

```
run 1: first PMT section at packet 1443   video+audio before it: 1439
run 2: first PMT section at packet  741   video+audio before it:  740
run 3: first PMT section at packet 1362   video+audio before it: 1361
```

So roughly 700–1400 packets, a good second of stream. Not a couple of packets.

## What I implemented

Nulling, while a PMT is withheld, exactly the elementary pids **that PMT names** —
not "all pids". Deliberately per service: the adapter buffer is shared by every
client on that adapter, so blanking all pids there would blank a second client
watching a different service on the same transponder. Your `-E -E -k` case was
safe with it, since only the held service's own pids were touched.

## Why it did not work

The pids only become known once the PMT has been parsed — and almost the whole
window you want to cover is *before* that. Measured, same service, same setup:

```
--clean-psi          video+audio before the first PMT: 1101
--clean-psi=secure   video+audio before the first PMT: 1742
```

No improvement, within run-to-run noise. And the distribution of video packets
across the capture is flat — there is no withheld window anywhere:

```
video packets per 1/20 of the stream:
258 266 262 267 268 259 265 267 256 262 259 234 255 255 207 244 252 252 256 239
```

What it can do is create a gap in the *middle* of the stream, after the PMT is
parsed but before the key arrives, which is worse than the problem.

## Where it belongs

"Give this client nothing until its PMT is ready" is a property of one client, and
the adapter buffer is shared by all of them. Doing it correctly means gating in
the per-client send path, not in the buffer — and then "all pids except 0-21" is
safe, because it only affects the client that has not had its PMT yet.

That is a change to the streaming path rather than to this patch, and I would
rather not smuggle it into a PR that is still being argued over. If @catalinii or
@Jalle19 want it, I will do it as its own PR with its own measurements.

So this PR keeps three values, all measured and all doing what they say:
`--clean-psi[=seconds]`, `--clean-psi=0`, `--clean-psi=all`.

Jalle19:
Getting tired of talking to an AI. VDR obviously knows which channel is encrypted or not since it knows to assign it to a CI etc. Tvheadend has no problem playing or recording channels descrambled by minisatip.

I will close this issue unless someone can point to a clear improvement here. Fixing VDR issues is not good enough. The fact that this PMT rewriting is off by default is not enough grounds to merge a complex feature like this. It was removed for a reason (both broken and probably thus unused).

lars18th:
Hi,

From my point of view the functionality of `---clean-psi` is useful in some different scenarios. So I suggest to not withdraw this PR. Anyway, after thinking more on the right implementation, it seems that the place to do the filtering is not at the adapter level, but at the served stream. This implies that the PMT requires to be patched when it's sent to the user. This will solve the case when different users are showing services from the same transponder. Furthermore, to minimize the impact of this some caching of the active patched PMT is required. With this approach any secondary request of the same service will not require to repatch the same PMT. But in any case, the procedure implemented on this PR patching inplace the PMT reducing the size and recalculating the stuffing in the TS packet header is the correct way to do the task without troubles.

So @dnehring7 , what you think about this proposal?

dnehring7:
> Getting tired of talking to an AI. VDR obviously knows which channel is encrypted or not since it knows to assign it to a CI etc. Tvheadend has no problem playing or recording channels descrambled by minisatip.
> 
> I will close this issue unless someone can point to a clear improvement here. Fixing VDR issues is not good enough. The fact that this PMT rewriting is off by default is not enough grounds to merge a complex feature like this. It was removed for a reason (both broken and probably thus unused).

@Jalle19 you have a point, and I'm sorry about the walls of text. Yes, I use AI
tooling to help me testing all edge cases and English is not my first language - but the complicated setup, the
testing and the persistence are mine. I'll try to keep it short from now on.

The one clear improvement, in plain words: when minisatip descrambles with
oscam, the stream is FTA but the PMT still announces the CAIDs. Tvheadend
doesn't care, it looks at the scrambling bits when the data arrives. VDR decides
before tuning, from the CAIDs in the PMT - and a SAT>IP device has no CI, so
it refuses the channel completely, live and recording. That is my daily setup,
that's why I keep pushing. The client has no way to know the server already
descrambled the stream. The PMT is the only place the server can say so.

About the old clean_psi: I read it before writing mine. The code was broken at all.
I understand the scepticism, once burned...

If you still don't see the value, close it - I keep it in my tree, no hard
feelings, really. But if the concern is the design rather than the feature, I'd
rather rework it than push to merge it as it is.

@lars18th I think you're right that per-stream is the better place in the long
run. The adapter buffer is shared, so at that level every client of a service
must get the same PMT. A per-client version is more flexibly, it's a real rework though.

We could refactor it, but only if @Jalle19 or @catalinii say the feature has a chance with that design.
I don't want to build anything that gets closed anyway.

lars18th:
Hi @dnehring7 ,

I appreciate your effort with this PR. IMHO is a must have. However, @catalinii and @Jalle19 needs to approve it. And for my environment it will be very useful. I hope they agree (almost @catalinii) with the addition of the functionality. However, I feel it will be necessary to move the PMT patching to the right place: the client stream. And for this, we need to wait until the agreement.

Please, @catalinii what's your point of view?

Jalle19:
> The client has no way to know the server already
descrambled the stream. The PMT is the only place the server can say so.

This is demonstrably not so since tvheadend correctly looks at transport_scrambling_control instead of trusting the PMT. I suggest you try to patch VDR to look at this field instead of only looking at the PMT.

lars18th:
Hi @Jalle19 ,

> > The client has no way to know the server already
> > descrambled the stream. The PMT is the only place the server can say so.
> 
> This is demonstrably not so since tvheadend correctly looks at transport_scrambling_control instead of trusting the PMT. I suggest you try to patch VDR to look at this field instead of only looking at the PMT.

This is not my responsibility. But @dnehring7 has responded previously that the VDR code doesn't start playing any channel before reading the PMT, and this _by design_. So it's a no-go patching VDR.

However, the problem in my use case is not VDR. The problem is with players that always trust the PMT (which is what the DVB and MPEG-TS standards specify). At beginning (in 2015) I've started to use the `--clean-psi` introduced by @catalinii . And then I was disappointed when the feature was removed. IMHO it's difficult to understand why an _optional_ functionality correctly implemented is not accepted. Anyway, thank you @dnehring7 for your effort.

Please, @catalinii can you add your point of view?

sam-stenvall-iceye:
I wouldn't trust anything @dnehring7 's AI says. Obviously VDR can be patched.

> IMHO it's difficult to understand why an optional functionality correctly implemented is not accepted

The code, and thus the complexity, is still there even if it's optional.

sam-stenvall-iceye:
It's still unclear to me what other clients have issues with this.

lars18th:
Hi @Jalle19 ,

> I wouldn't trust anything @dnehring7 's AI says. Obviously VDR can be patched.
> 
> > IMHO it's difficult to understand why an optional functionality correctly implemented is not accepted
> 
> The code, and thus the complexity, is still there even if it's optional.

Your point of view is clear. I suggest to wait what @catalinii thinks about this. I don't want to create friction. But for my use case the option `--clean-psi` is valuable.

> It's still unclear to me what other clients have issues with this.

Because the standards indicate this! The strategy of discarding the PMT info and use the encrypted flag of the TS packet _is a patch_ done by specialized software/hardware. When using standard players the info of the PMT requires to be consistent. Why then you think that in 2015 @catalinii has introduced this functionality? Because is _very_ useful. However, the implementation at this time was inefficient. And hi has decided to remove it only because hi doesn't have time to fix it. But now @dnehring7 has presented a good alternative. IMHO the complexity of this PR is acceptable and it includes the corresponding unit tests. So from my point of view, it's good to merge it (when completed the change from "patching PMT at adapter level to patching at user level").

So, please wait until @catalinii expressed his point of view. You agree? 😉

dnehring7:
I regret that @Jalle19 does not have enough confidence in this PR, and I fully respect his concerns.

I have tried to explain the motivation from several perspectives to show the usefulness of the patch. VDR is only one example, not the main reason — and @Jalle19 is absolutely right that I could also change the behavior in the VDR plugins themselves.

@lars18th understood the motivation behind the PR and supports it. I also think @lars18th’s suggestion is a very good way forward.

catalinii:
@dnehring7 thank you for working on this feature so diligently. I really appreciate the time you've put into it.

To me the bottom line is that is hardly used. The second thing is that there are workarounds (as you said, marking that the satip plugin can descramble encrypted channels, or HasInternalCam()). 

Thirdly and probably most importantly, is a complex feature that will need a lot of tests to be maintained, otherwise we will do the same thing: add it now, remove it few year later. 

To me the VDR logic seems to be the problem. With DDCI/TBS6900 where the stream comes from has no relevance as those CI cards can decrypt even streams coming from a minisatip instance not decrypting anything. Also the fact that there are channels that have the PMT send CA descriptors but the stream is clean makes this logic a bit outdated.

Jalle19:
> Because the standards indicate this! The strategy of discarding the PMT info and use the encrypted flag of the TS packet is a patch done by specialized software/hardware.

The DVB CI standard makes no mention of PMT rewriting, it does however say that the CAM should clear the transport_scrambling_control bit. If you can point me at some standard that says otherwise I might be convinced.

lars18th:
> > Because the standards indicate this! The strategy of discarding the PMT info and use the encrypted flag of the TS packet is a patch done by specialized software/hardware.
> 
> The DVB CI standard makes no mention of PMT rewriting, it does however say that the CAM should clear the transport_scrambling_control bit. If you can point me at some standard that says otherwise I might be convinced.

Sure! But this is for _receivers_ with CA support. And players for FTA only streams can't manage CA info. And several of them refuse to play when some encrypted info is present in the PMT. I feel you're mixing TV receivers with Video Players. Today exist a lot of devices that can play MPEG-TS streams that don't know anything about TV standards.

dnehring7:
> > Because the standards indicate this! The strategy of discarding the PMT info and use the encrypted flag of the TS packet is a patch done by specialized software/hardware.
> 
> The DVB CI standard makes no mention of PMT rewriting, it does however say that the CAM should clear the transport_scrambling_control bit. If you can point me at some standard that says otherwise I might be convinced.

My sentence was not technically correct: EN 50221 does not require anyone to rewrite the PMT! What I want to say: a CA descriptor says which CA system belongs to a stream and where its ECMs are. It does not say the packets
are scrambled at this moment - that is what the scrambling bit says.

The difference matters because a recorder does not decide per packet. VDR
decides once per service: does this channel need a CAM? It cannot answer that
from a packet header - packets can be clear for a while anyway - so it reads the
PMT, and on an already descrambled service it finds CA descriptors there.
tvheadend does not run into this because it clears the bit and judges per packet.
VDR does not work that way. So I cannot rewrite the full VDR logic here. And I think
other clients handle it in a similar way.

So the patch is normalizing the PSI of a stream minisatip has remultiplexed anyway (look at  ddci_create_pmt()).
I only uses the same idea in the other direction, drops the CA
descriptors instead of remapping them, and only after a control word really
decrypted the service. Off by default.

But I understand all your arguments here: narrow use case, patch is too big. I'll think about it, will come back to this question.

lars18th:
@dnehring7 ,

> But I understand all your arguments here: narrow use case, patch is too big. I'll think about it, will come back to this question.

My point of view is: _not_ narrow use case, and _not_ too big.
However, I recommend reorganizing the workflow to patch the PMT at user stream level. And please, remove references to VDR inside the code because this functionality is generic and not targeting only it.

I hope finally @catalinii will accept to reestablish this useful functionality. 🤞

dnehring7:
Thank you very much, @Jalle19, @catalinii and @lars18th for critically  reviewing my proposed changes.

I have read your comments again and I think you are both right about the feature. It is narrow, it needs tests to stay alive, and I cannot argue it from the standards - that was my mistake.

But while testing it I found a few things in the PMT parser that have nothing to do with the option:

- the two descriptor loops trust the length byte and can read past the end of the section
- a CA descriptor shorter than 4 bytes gives pmt_add_caid() a negative length
- the program level descriptors are not cleared on a PMT version change, so descriptors of the old version stay in the parsed PMT
- streams that are neither audio nor video never get their descriptors parsed, so teletext and subtitles lose them

The first two are memory safety, the other two make the parsed PMT incorrect. All four are worth having with or without --clean-psi.

So my proposal: I close this PR and open small separate ones for those fixes, each with tests. The feature itself I keep downstream. If more people ask for it later, it can come back on top of a clean base.

Is that ok for you?

Jalle19:
There are som WIP pull requests that attempt to refactor the PMT handling, you can look at those if there is something you want to pick up.

dnehring7:
I have splitted this PR in 5 parts, hopefully PR #1427 (3 patches) and PR #1428 (1 patch) find acceptance.

lars18th:
Hi @dnehring7 ,

I feel you need to consider this edge case:

- minisatip is receiving the stream from another SAT>IP or tuner that has already decrypted the stream, but without cleaning the PMT. In this case all TS packets will be in clear with the corresponding 00 bits set in the header.

I feel that in this case the detection of the decryting will fail in your current code. Please check it.

dnehring7:
> Hi @dnehring7 ,
> 
> I feel you need to consider this edge case:
> 
> * minisatip is receiving the stream from another SAT>IP or tuner that has already decrypted the stream, but without cleaning the PMT. In this case all TS packets will be in clear with the corresponding 00 bits set in the header.
> 
> I feel that in this case the detection of the decryting will fail in your current code. Please check it.

You were right, it failed exactly there. It only trusted a control word it had checked itself, so a stream arriving already in the clear was never recognized as descrambled and the PMT went out with all its CAIDs.

Reproduced with two minisatip instances, the second fed by the first. It now looks at the stream itself as well, which also covers a CAM or a hardware descrambler. Tested on RTL HD, all four cases correct. Thanks!

tests/test_pmt.cpp
```diff
@@ -335,10 +336,568 @@ int test_emulate_add_all_pids() {
...
+// A broadcast PMT: CA descriptors on the program level and on the audio
+// stream when ca is set, a private descriptor, a video, an audio and a teletext
+// stream with their descriptors.
+static int build_broadcast_pmt(uint8_t *s, int sid, int version, int cni,
+                               int ca) {
+    static const uint8_t pca[] = {0x09, 0x04, 0x09, 0x8C, 0xE0, 0xCC};
+    static const uint8_t priv[] = {0x5F, 0x04, 0x00, 0x00, 0x00, 0x2A};
+    static const uint8_t aca[] = {0x09, 0x04, 0x09, 0x8D, 0xE0, 0xCD};
+    static const uint8_t lang[] = {0x0A, 0x04, 'd', 'e', 'u', 0x00};
+    static const uint8_t ttx[] = {0x56, 0x05, 'd',  'e', 'u',
```
lars18th: Hi @dnehring7 ,

Remember to normalize these values: _all_ values in HEX `0x??` format, please.

lars18th:
> Reproduced with two minisatip instances, the second fed by the first. It now looks at the stream itself as well, which also covers a CAM or a hardware descrambler. Tested on RTL HD, all four cases correct. Thanks!

Nice! I feel this PR is improving. And I hope soon the required #1428  be accepted to concentrate on this. 😉

lars18th:
Hi @dnehring7 ,

I feel all other PRs as requirement are now merged. That's right?
If yes, then you'll wait until the RFC #1445 will be closed? Or your idea is to continue?

Please, don't think that I want to enforce you to finish this. My objective is to know if we can do anything to help you. In my use cases the `--clean-psi` option correctly implemented is a must-have.

Thank you.

PR #1427, opened 2026-08-29
fix(pmt): clear the program level descriptors on a PMT version change

dnehring7:
`process_pmt()` clears `stream_pids` when a new version arrives but not
`descriptors`, so the program level descriptors of the previous version stay in
the parsed PMT. Anything that reads them back afterwards - `ddci_create_pmt()`,
or `--clean-psi` in #1425 - then sees a table that was never broadcast.

Earlier revisions of this PR also bounds checked the two descriptor loops,
`is_ac3_es()` and the section lengths, and refused a CA descriptor too short to
hold a CAID. @Jalle19 pointed out that the PMT filter is registered with
`FILTER_CRC` and that `assemble_packet()` verifies the CRC before
`process_pmt()` is ever called, and that a CA descriptor of abnormal length is
not something that occurs, the field being standardized. Both are dropped, so
what is left is the one line above.

### DIFF
--- a/src/pmt.cpp
+++ b/src/pmt.cpp
@@ -1903,6 +1903,7 @@ int process_pmt(int filter, unsigned char *b, int len, void *opaque) {
     pi = b + 12;
     pmt_b = b + 3;
 
+    pmt->descriptors.clear();
     pmt->stream_pids.clear();
 
     // Add PMT level desciptors from program info


lars18th:
Hi @dnehring7 ,

Nice. However, please change to `LOGM()` instead when the PMT descriptor is truncated. For a regular use with some services from providers with wrong hardware will create a large unnecessary logging. You agree?

Jalle19:
Correct me if I'm wrong, but when the packet is parsed and the CRC matches, doesn't that mean we can trust the various length indicators?

dnehring7:
> Correct me if I'm wrong, but when the packet is parsed and the CRC matches, doesn't that mean we can trust the various length indicators?

You're right, so I've dropped both bounds checking commits.

Two left. The CA descriptor check and the process_pmt() check, they are still necessary!

src/pmt.cpp
```diff
@@ -1770,8 +1770,10 @@ void pmt_add_descriptors(SPMT *pmt, unsigned char *pi, int len) {
 
         pmt->descriptors.push_back(d);
 
-        // Handle CA descriptors separately
-        if (d.is_ca_descriptor()) {
+        // Handle CA descriptors separately. Below four bytes there is no
+        // CAID and no CA pid to read, and the private data length would be
+        // negative.
+        if (d.is_ca_descriptor() && pi_len >= 4) {
```
Jalle19, line 1776: You can add the length check to is_ca_descriptor instead

src/pmt.cpp
```diff
@@ -1903,6 +1907,9 @@ int process_pmt(int filter, unsigned char *b, int len, void *opaque) {
     pi = b + 12;
     pmt_b = b + 3;
 
+    // Both describe the version being parsed now, so neither may keep what the
+    // previous one had
```
Jalle19, line 1911: Comment unnecessary IMO

src/pmt.cpp
Jalle19: In general I don't think it's possible to encounter a CA descriptor with abnormal length. These things are standardized.

dnehring7: The length check moved into is_ca_descriptor(). The whole commit is five lines in pmt.h plus a test.

src/pmt.cpp
dnehring7: Comment removed.

Jalle19 (approved):

Jalle19:
@catalinii please sanity check

catalinii (approved):

Issue #1415, opened 2026-08-15
The -9 flag does not work


iandebris:
I guess this is a continuation of #1129.

When using `-9`, when I first load up a channel with a shared video PID, everything works fine. When I change to a different channel using the same video PID but different audio, PCR and PMT PID, minisatip still tries to use the PMT of the first loaded channel.

For example, load "NATURE ESCAPE" on 12188 V @ 42 E (VPID: `108`, APIDs: `208` and `308`, PCR PID: `308`, PMT PID (also SID): `1908`) with:

```
http://172.30.0.10:8080/?freq=12188&sr=27500&pol=V&msys=DVBS2&mtype=8psk&pids=0,1,16,17,18,20,108,208,308,1908
```

Video and audio works alright, then change to "2000'ler" on same TP (VPID: `108`, APID: `220`, PCR PID: `308`, PMT PID (also SID): `1920`) with:

```
http://172.30.0.10:8080/?freq=12188&sr=27500&pol=V&msys=DVBS2&mtype=8psk&pids=0,1,16,17,18,20,108,220,308,1920
```

Video is decoded correctly, audio isn't.

(FWIW, the video is actually from a different channel called "iTVGAMETEST" with VPID: `108`, PCR PID: `8191`, PMT PID: `4250`, SID: `275`)

Logs (taken with cmdline `/usr/local/minisatip/minisatip -f -i 99 -D 0 -o '~/tmp/camd.socket' -9 -v pmt > pmt.log`): [pmt.log.gz](https://github.com/user-attachments/files/31109230/pmt.log.gz)

MPEG-TS sample of the entire TP (with `pids=all`, not logged): [12188.ts](https://fileditchfiles.st/alpha25/c54dcad72ed0d1715549/12188.ts)

catalinii:
Hi,

This is by design the behavior, where if there are 2 channels with overlapping pids but one of them has different audio pid (in this case), depending on which pmt arrives first, that PMT will be the master pmt and send for the decryption of both channels (so your audio channel is not sent as part of the CAPMT to be decrypted).

How many channels have the same behavior on 42e?

PR #1428, opened 2026-08-29
fix(pmt): parse every elementary stream the same way

dnehring7:
`process_pmt()` decided per stream whether it was audio or video and treated
everything else as second class: `pmt_add_stream_pid_descriptors()` was skipped
for it, so a teletext or subtitle stream ended up in the parsed PMT with no
descriptors at all. `ddci_create_pmt()` rebuilds the PMT from exactly that and
emits those streams bare.

What that costs, on RTL HD, measured through the PMT rebuild in #1425 because
it is the same parsed data:

```
without this PR                      with it
section_len=69                       section_len=111
  0x1B 255: 28/4                       0x1B 255: 28/4
  0x06 259: 0A/4 6A/4                  0x06 259: 0A/4 6A/4
  0x06  32: -                          0x06  32: 52/1 56/5        teletext
  0x05 261: -                          0x05 261: 6F/3
  0x06  48: -                          0x06  48: 59/8             subtitling
  0x0B 262: -                          0x0B 262: 52/1 13/5 66/2
  0x0C 263: -                          0x0C 263: 52/1
```

The client is told those streams exist but not that one is subtitles, nor in
which language.

## is_audio and is_video

@Jalle19 they are gone rather than carried further by a new flag. The three
places that asked for them want the same thing, and it is not a stream type:
skip the one entry in `stream_pids` that is not an elementary stream, the PCR
pid of a service that has one of its own. It already carries `type == 0`, which
no broadcast uses, so that is what they test now.

What changes for a stream that is neither audio nor video:

| | |
|---|---|
| its descriptors | recorded, so a rebuilt PMT still describes it |
| a CA descriptor on it | registers its CAID |
| a client requesting only that pid | starts the PMT that carries it |
| the CAPMT | lists it, as the DDCI PMT already did |
| `get_master_pmt_for_pid()` | matches on it, so two services sharing one can end up master and slave |

The last row is the one worth an opinion. Two services that share a data pid
but not a control word would now be linked, where before only shared audio or
video linked them. I have not seen it on 19.2E, and the alternative is keeping
a stream classifier alive for that one call, but say the word and it can test
something narrower there.

## DDCI

`ddci_create_pmt()` maps every pid it writes into the PMT it hands the CAM: the
stream pids, and the ECM pids of the program level CA descriptors it rebuilds
from `pmt->ca[]`. A CA descriptor inside a stream's descriptor list was copied
out byte for byte, ECM pid included, so it named a pid of the source adapter -
one the CAM never sees, because nothing is routed to it under that number. That
is true on master already for audio and video; this PR would extend it to every
other stream, so it is fixed here. `test_create_pmt()` asserted the unmapped
pid, so it described the bug; it now maps that ECM pid to a number already
taken on the DDCI, which makes an unmapped one visible.

The CAIDs are the other half. `pmt_add_caid()` already skips a CAID and pid it
has, and in practice an ES level CA descriptor repeats what the program level
carries, so the list usually does not grow at all. `CAPMT_add_PMT()` writes into
a caller supplied buffer whose length it was passed and never looked at; now
that more streams reach it, it stops at the end of that buffer instead of
writing past it, and the CA descriptors - identical for every stream of a PMT -
are built once rather than per stream.

## Bounds

The first commit is a prerequisite, not a drive by. The parser follows the
descriptor loops with the lengths the section itself carries and never checks
them, and `is_ac3_es()` is already called with an `es_len` that is validated
only afterwards. Parsing the streams that used to be skipped means following
more of those lengths, so they are checked first: the section against the header
fields read from it, `program_info_length` against the room left for the
streams, each `ES_info_length` against the end of the loop, and each descriptor
against the end of the loop it sits in. `pmt_len` and `pi_len` are read before
`stop_pmt()`, so a section that fails the check leaves the running PMT alone.

## Testing

Fedora 44, gcc 16, OSCam over dvbapi, Digital Devices DVB-S2, RTL/Nagra on
Astra 19.2E. `ctest` 13/13 with `TEST_SANITIZERS=ON` and `OFF`. The section and
descriptor loop tests are driven over buffers sized to their exact content, so
a read past one is an ASan error rather than slack.

Live, RTL HD: descrambling unchanged, and a client asking for the subtitle pid
and nothing else now starts the PMT and gets it sent to the card server, which
on master it did not. No DDCI hardware here, so that path is covered by
`test_create_pmt()` only.

### DIFF
--- a/tests/test_pmt.cpp
+++ b/tests/test_pmt.cpp
@@ -80,6 +80,11 @@ uint8_t cw_invalid[] = {0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77};
 extern adapter *a[MAX_ADAPTERS];
 extern SCW *cws[MAX_CW];
 
+int is_ac3_es(unsigned char *es, int len);
+void pmt_add_descriptors(SPMT *pmt, unsigned char *pi, int len);
+void pmt_add_stream_pid_descriptors(SPMT *pmt, SStreamPid &sp,
+                                    unsigned char *es, int len);
+
 int test_descriptor_equality() {
     const uint8_t descr1_data[] = {0x09, 0x04, 0x0B, 0x00, 0x05, 0x73};
     descriptor_t descr1 = create_descriptor(descr1_data);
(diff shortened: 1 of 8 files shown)


lars18th:
No updates over this PR?

tests/test_pmt.cpp
```diff
@@ -335,10 +337,36 @@ int test_emulate_add_all_pids() {
     return 0;
 }
 
+int test_stream_pid_descriptors() {
+    SPMT pmt = {};
+    SStreamPid sp;
+    unsigned char es[] = {0x0A, 0x04, 'd',  'e',  'u',  0x00, 0x56,
+                          0x05, 'd',  'e',  'u',  0x10, 0x00, 0x09,
+                          0x04, 0x09, 0x8D, 0xE0, 0xCC};
```
lars18th: I recommend normalizing to hex numbers for all values

Jalle19 (commented):
Not really sure anymore what we're trying to fix here. Ideally all "is audio is video" stuff should go away. And everything needs to be compatible with all the PMT and CAID stuff that happens for DDCI.

dnehring7:
> Not really sure anymore what we're trying to fix here. Ideally all "is audio is video" stuff should go away. And everything needs to be compatible with all the PMT and CAID stuff that happens for DDCI.

Fair, I retitled it.

Non-A/V streams get no descriptors at all, so DDCI hands the CAM a subtitle stream with no subtitling descriptor and no language. 5 of 7 streams on RTL HD here in Germany.

I dropped is_audio/is_video. Are you fine with it?

First commit contains bounds checks, needed because we now follow more section lengths. Can split it out if you want.

Jalle19:
Sorry but I don't have any confidence in these changes, I don't think we should pursue this.

dnehring7:
Fair enough, closing. I've reworked #1425 to use the broadcast PMT section directly, so it doesn't need this anymore. Thanks for looking at it.

lars18th:
Hi @Jalle19 and @dnehring7,

Why this PR is discarded? I understand why @Jalle19 says that is not _necessary_. However, the example at top from @dnehring7 about the PMT rewrite is relevant... The internal code requires to do _correct_ actions. And any pid inside a PMT is a valid pid (whether it’s video or audio) so it's needs to continue inside it _every time_.

Please clarify this further.
Thank you.

Jalle19:
I said I don't have confidence in the changes, not that we don't want to improve this.

catalinii:
I looked also at the diff and the main issue that is not solved is related to linking PMTs.

The second problem contributing to this is the fact that the PMTs are linked as they are read (eg first PMT owning pid X will be master, following PMTs hl having the same pid X will be slave).

There are 3 places where that matters:
- sending the pmt to oscam/CAM will send the master PMT
- getting a CW for a PMT will happen on the master PMT
- Decrypting the stream (using the CW) happens also on the master PMT.

The issue with this diff is that it relaxes the linking criteria and if 2 PMTs having different audio and video streams can be linked, so one of the PMt will not work.

The approach with audio and video linking is also not fully correct, see #1415 .

dnehring7:
Thanks, that's a fair point. The linking change came from pulling out is_audio/is_video everywhere, it wasn't what I was after, and #1415 shows the audio/video rule is fragile enough without loosening it. Shouldn't touch it now.

What @lars18th means only needs every stream to keep its descriptors, and that doesn't have to touch linking. The earlier version of this PR did just that: linking, CAIDs and the CAPMT unchanged. If you want it, I'll reopen with only that. Otherwise I'll leave it to the PMT refactor.

catalinii:
Maybe the other way around this is to scrape the master/slave concept and generate a CAPMT based on the list of pids that are asked for:
- if pmt pid is in the list use that PMT pid + sid
- for every pid asked, generate the ES record from the PMT containining all CAIDs from all PMTs

Then store pid-> CW mapping.

PR #1429, opened 2026-08-30
pmt: keep PMT/ECM if ES still derived (mirror satip-client psi fix for 225/2250)

catalinii:
Hypothesis for 10.0.0.5 vuuno4kse: force PMT pid itself as active and linger PMT 225 + ECM 2250 while ES pids still in derived/requested, avoiding PLAY delpids gap. Mirrors satip-client psi.cpp fix (fix-keep-pmt-if-es-still-added).

- Verified without reboot via ssh: enigma2 pid 5051, satipclient 1931, logs show PMT 225/ECM 2250 lifecycle
- Exercised rebuilt satip-client 1.6M in build dir: model shows old drops {0} vs new keeps {0,225,2250} for force-PMT and linger paths

### DIFF
--- a/src/pmt.c
+++ b/src/pmt.c
@@ -1110,9 +1110,11 @@ void start_active_pmts(adapter *ad) {
         int pmt_started = 0;
         for (j = 0; j < pmt->stream_pids; j++)
             // for all audio and video streams start the PMT containing them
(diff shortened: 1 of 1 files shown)


Jalle19:
What?

catalinii:
Seems like --yolo is not a good idea :)

Issue #1129, opened 2023-10-14
Minisatip tries wrong PMT


surfcu:
I can watch all Digital Platform channels on my enigma2 (Zgemma h9 twin se) box 42e.
When running Minisatip can provide all but beIN SPORTS 2 (PMT=48).
Instead when pmt=48 is requested minisatip tries Ticari beIN Sports 2 (PMT=52) and fails since it's not in the subscription.

[minisatip.log](https://github.com/catalinii/minisatip/files/12907415/minisatip.log)

Any ideas?

surfcu:
I think problem is pmt 48 is misinterpreted as audio. But I don't know much about coding.

iandebris:
i think this happens when the VPID (and PCR) is the same on multiple channels and only APID is different (or reverse, same APID but different VPID on different PMTs like beIN SPORTS ticari/commercial and non-commercial) as it also happens on d\*\*\*t 42e as well with stingray music channels (12188 V 27500, 8PSK DVB-S2 5/6), all music channels like `DEEP BLUES` (PMT 1935, APID 235), `RIDDIM` (PMT 1937, APID 237) etc use the VPID of `iTVGAMETEST` (PMT 4250, VPID 108) and its a roll of dice whether or not the correct PMT is utilized.

for instance if you load up `NATURE ESCAPE` (PMT 1908, PCR 308, VPID 108, APID 208), minisatip may use the correct thing; or just go pick up random other PMT that has the same VPID, for instance `'80'ler` (PID 1922, PCR 308, VPID 108, APID 222)

i've included a sample log from a raspberry pi 4 with TurboSight TBS 5930 DVB-S/S2/S2x

[msip.log](https://github.com/user-attachments/files/15860397/msip.log)

Jalle19:
Thanks for the additional clues!

Jalle19:
So it seems that the problem is that `get_master_pmt_for_pid()` returns any PMT that happens to contain some particular PID. This leads to that PMT (PID 1922 in the log above) being set as the master PMT for all the other PMTs in the same PAT, which is obviously incorrect.

Jalle19:
Same issue in @surfcu 's log (PMT 1 (52) gets incorrectly marked as master for PMT 10 (48)):

```
$ cat minisatip\(8\).log | grep -e 'new PMT 1 \|new PMT 10 \|master pmt set to'
[...]

PR #1439, opened 2026-09-16
Fix duplicate-SID PMT master election without disabling PMT scan (#1129)

surfcu:
get_master_pmt_for_pid() elected whichever sibling PMT was parsed first as master, regardless of which SID the client actually requested. This now prefers whichever sibling the client's own pid subscription (or an explicit x_pmt=) indicates, so full PMT scanning can stay enabled instead of requiring --disable-pmt-scan.


catalinii:
This solves the issue only on the first tune. If you have a 2 PMTs pointing at the same list of pids, and first time is being tuned to the first PMT, second time to the second PMT, only the first pMT will be sent in both scenarios.

A similar discussion is in https://github.com/catalinii/minisatip/pull/1428

surfcu:
I assume if the fix lives in start_active_pmts(), instead of process_pmt(), it would work

Jalle19:
I think we should just forego the whole "master PMT" concept and just require clients to specify the PMT PID. All this complexity isn't doing anyone any good.

surfcu:
@Jalle19 I agree the master PMT election is the actual source of this bug class — both #1129 and
the second-tune issue catalinii found are symptoms of the same thing: order-dependent state
computed once at discovery time, consumed later at tune time.

One case that needs a caveat first, though: `pids=all` (`emulate_add_all_pids()`). With it, every
PMT on the transponder gets marked with the client's real sid, not just the one the client
actually wants — so "the client specified this PMT pid" becomes true for all of them at once. If
we require clients to specify the PMT pid and start whatever they specify, that's every duplicate
sibling starting simultaneously, which blows through `max_channels` on the CAM fast.

So `master_pmt` isn't only doing selection — it's also deduplication (one CA-PMT per unique
elementary-pid set, not one per SID). That part still needs to exist, just not as stored,
order-dependent state.

Proposal: drop the `master_pmt` field entirely, and instead of "the client's own request wins," do
this fresh on every check:

1. Group active PMTs by identical elementary-pid set (computed on the fly, nothing stored).
2. Within a group, start whichever PMT the client's subscription actually points at (PMT pid in
   `pids=`, or `x_pmt=`).
3. If more than one in a group qualifies (`pids=all`, or two clients on different siblings), start
   exactly one — pick deterministically, e.g. lowest PMT pid.
4. If none qualify (client never specified any PMT pid), fall back to the same deterministic pick
   rather than not starting anything.

This is basically what my current patch does already (it's already computing a fresh decision each
check instead of trusting a stored field) — it would just go further and drop `master_pmt` as a
concept entirely instead of keeping it as a fallback, and add 3/4 so `pids=all` and
no-PMT-pid clients don't break.

Happy to update the PR this way if that direction sounds right to both of you.

Jalle19:
Emulate PIDs all is also a hack, I wouldn't worry too much if something breaks with it enabled.

lars18th:
> Emulate PIDs all is also a hack, I wouldn't worry too much if something breaks with it enabled.

Sorry @Jalle19 , but "emulate pids all" is not a hack, it's a functionality. We need to maintain it, because in some scenarios the method to pass all pids is not available (at system level, at hardware level, or at software level).

PR #1442, opened 2026-09-17
Fix duplicate-SID PMT selection (#1129) without requiring --disable-pmt-scan

surfcu:
Supersedes the [earlier revision](https://github.com/catalinii/minisatip/pull/1439) of this PR, addressing catalinii's review comment ;

> This solves the issue only on the first tune. If you have a 2 PMTs pointing at the same list of  pids, and first time is being tuned to the first PMT, second time to the second PMT, only the first PMT will be sent in both scenarios.
> 
> A similar discussion is in #1428

The fix now lives in start_active_pmts() rather than process_pmt(), so it is re-evaluated on every tune instead of being frozen at PMT-discovery time. Two further fixes turned out to be hard dependencies rather than optional follow-ups — see "Why these three changes ship together".

When two services share the same elementary stream pids (a duplicate-SID pair — e.g. a public service and a commercial simulcast with identical vpid/apid), get_master_pmt_for_pid() elects whichever PMT was parsed first as the group's master, and start_active_pmts() only ever starts the master:
```
if ((stream_pid.is_audio || stream_pid.is_video) && pids[stream_pid.pid] &&
    pmt->id == pmt->master_pmt) {
```
Which PMT is parsed first is scan-order luck, unrelated to which SID the client asked for, so the wrong sibling is silently streamed and decrypted. --disable-pmt-scan avoids this only by ensuring the sibling is never discovered — it doesn't resolve the ambiguity.

## 1. The original bug

When two services share the same elementary stream pids (a duplicate-SID pair — e.g. a public service and a commercial simulcast with identical vpid/apid), get_master_pmt_for_pid() elects whichever PMT was parsed first as the group's master, and start_active_pmts() only ever starts the master:```

```
if ((stream_pid.is_audio || stream_pid.is_video) && pids[stream_pid.pid] &&
    pmt->id == pmt->master_pmt) {
```

Which PMT is parsed first is scan-order luck, unrelated to which SID the client asked for, so the wrong sibling is silently streamed and decrypted. --disable-pmt-scan avoids this only by ensuring the sibling is never discovered — it doesn't resolve the ambiguity.

## 2. Why the first attempt was wrong

The earlier revision re-elected the master inside process_pmt()'s stream-pid loop. That path is gated by:

```
if (pmt->version == ver) {
    return 0;   // the re-election never runs again
}
```
A client retuning to a different sibling doesn't change the broadcaster's PMT content or version, so process_pmt() short-circuits and the stale master from the first tune is never reconsidered — exactly as catalinii described. This is the same structural point raised on #1428: PMT processing is "once per content version, shared globally", and a per-client, per-tune decision doesn't belong there.

## 3. The fix

start_active_pmts() already re-runs independently of PMT versioning and already has current pid subscription state. It now derives, fresh on every call, which sibling the current subscription actually points at:

```
std::unordered_map<int, int> preferred_owner; // stream pid -> pmt id
for (i = 0; i < ad->active_pmts; i++) {
    SPMT *pmt = get_pmt(ad->active_pmt[i]);
    if (!pmt) continue;
    bool client_wants_this_pmt = false;
    if (ad->tp.x_pmt.has_value() && ad->tp.x_pmt.value() == pmt->pid) {
        client_wants_this_pmt = true;
    } else if (SPid *client_pid = find_pid(ad->id, pmt->pid)) {
        for (int s : client_pid->sid)
            if (s != PID_STREAM_ID_UNDEFINED) { client_wants_this_pmt = true; break; }
    }
    if (!client_wants_this_pmt) continue;
    for (const auto &stream_pid : pmt->stream_pids)
        if (stream_pid.is_audio || stream_pid.is_video)
            preferred_owner[stream_pid.pid] = pmt->id;
}
```
and the activation check consults it, falling back to the existing static field when no preference exists:
```
auto pref = preferred_owner.find(stream_pid.pid);
bool is_master_here = (pref != preferred_owner.end())
                           ? (pref->second == pmt->id)
                           : (pmt->id == pmt->master_pmt);   // unchanged fallback
```
Two signals count as "the client wants this one":

1. x_pmt=<pid> in the tune request. This parameter already existed (parsed in dvb.cpp into ad->tp.x_pmt) but was never read anywhere — this is its first use.
2. The PMT's own table pid present in the client's pids= subscription, under a real SID. Per the SAT>IP spec the client must fully define the media stream object, so it has to request the PMT pid it intends to parse. This works with clients that don't know about x_pmt (Tvheadend included). The sentinel PID_STREAM_ID_UNDEFINED is excluded, because the background PAT scanner marks every PMT pid on the transponder active regardless of client interest — "active" alone does not mean "wanted".

process_pmt() is unchanged; the static master_pmt field is no longer written by this fix, only read as a fallback.

## Why these three changes ship together

Enabling full PMT scanning for duplicate-SID transponders exposes two pre-existing bugs that are latent while --disable-pmt-scan hides the sibling PMTs. Without them, this PR makes things worse, not better — tuning three channels fails outright.

### 3a. DDCI channel slots leaked on re-registration (ddci.cpp, 33 lines)

send_pmt_to_cas() is called again for an already-registered PMT whenever a new CA descriptor is discovered for it, because pmt_add_caid() clears pmt->ca_mask to force a re-send.
ddci_process_pmt() already detected the existing slot but still incremented d->channels unconditionally. Observed with scanning enabled:

```
found DDCI 8 for pmt 3, running channels 1
found DDCI 8 for pmt 3, running channels 2     <- same PMT
found DDCI 8 for pmt 3, running channels 3     <- same PMT
...
find_ddci_for_pmt: DDCI 8 cannot be used for PMT 57 (used channels 6 max 6), skipping
```
Two real channels consumed all six slots and the third was refused. The increment is now conditional on a genuinely new registration, and the decrement in ddci_del_pmt() is made symmetric (it previously fired before confirming the PMT held a slot).

### 3b. pmt->ca_mask overloaded, breaking CA cleanup (pmt.h, tables.cpp, 28 lines)

ca_mask carried two different meanings:

- "already sent to this CA" — cleared by pmt_add_caid() to force a re-send
- "registered with this CA, so cleanup is required on stop" — read by close_pmt_for_ca()

When a CA-descriptor update landed in the same processing burst as a PMT stop, the first meaning clobbered the second, ca_del_pmt was silently skipped, and the CA kept the PMT registered permanently. On DDCI that leaked a channel slot per duplicate-SID channel tuned — three such channels held four slots, so roughly three zaps exhausted the CI.

A separate ca_registered_mask now tracks actual registration. pmt_add_caid() still clears ca_mask (the re-send signal is legitimate) but no longer destroys the cleanup guard. close_pmt_for_ca() and close_pmt_for_cas() guard on the new field and clear it on cleanup.

## Testing

Unit tests (tests/test_pmt.cpp), two added:

- test_duplicate_sid_master_election — the #1129 case: client tunes the sibling discovered second. Asserts on PMT state via start_active_pmts() rather than the now-unused master_pmt.
- test_duplicate_sid_master_election_second_tune — catalinii's case: tune sibling A, then retune to sibling B with no new PMT section arriving (only pid subscriptions change). Asserts B runs and A does not.

Both fail on unpatched source and on the earlier process_pmt()-based revision; both pass here.
All 8 pre-existing test_pmt tests continue to pass.

Live testing — Türksat 42°E, duplicate-SID triples (public / commercial variants of be Sports 1–3, identical vpid/apid within each triple), real Tvheadend SAT>IP client (which does not send x_pmt=, so signal 2 is exercised exclusively), Digital Devices DuoFlex CI:

- Correct sibling resolved, verified by actual content playback, across single-channel, two-channel same-adapter, two-channel different-adapter, and mixed configurations.
- Retune between siblings with no restart resolves correctly (the case that failed before).
- Three channels from three transponders start correctly. Slot accounting verified:
```
found DDCI 8 for pmt 3,  running channels 1      (x3 re-registrations, count stays 1)
found DDCI 8 for pmt 19, running channels 2      <- unrequested sibling
ddci_del_pmt: deleting pmt id 19 ... (registered 1, running channels now 1)
found DDCI 8 for pmt 24, running channels 2      <- the requested PMT
found DDCI 8 for pmt 56, running channels 3
```
The unrequested sibling is now released in the same millisecond the correct PMT takes over; three channels occupy three slots. Before 3a/3b that was four slots, or a hard failure.

- --disable-pmt-scan is no longer needed in any tested configuration.

## Backward compatibility

- Channels with no duplicate-pid sibling are unaffected — preferred_owner is only populated when a pid-sharing group with an active client signal exists.
- Clients sending x_pmt= behave as before.
- With neither signal present (client subscribed only to shared video/audio pids), behaviour falls back to the original master_pmt decision, unchanged.
- 3a and 3b correct counting and cleanup only; neither changes when a PMT is sent to a CA. 3b touches the shared CA path, so it also applies to dvbapi/oscam setups — cleanup now happens in cases where it was previously skipped.

## Not included

Separately, I have a change that moves the DDCI write to a dedicated thread (decoupling CI write cadence from the tuner read cadence) plus short-write retry and EOVERFLOW checking. It is unrelated to #1129 and did not affect the continuity-error rate it was written to investigate, so I've kept it out of this PR and can raise it separately if useful.


catalinii:
Let's not do this, this is adding another hack on top of an already existing hack. I am making a proposal for scraping master->slave pmt and we can all comment on that proposal.

https://github.com/catalinii/minisatip/issues/1445

surfcu:
Fair enough, I will not argue. You are right that it is a second election on top of the first one, and it still falls back to master_pmt when there is no signal from the client. Two places deciding the same thing. Jalle19 said the same in #1439. I am happy to see #1445 replace all of it and I will comment there.

Before I close this, one question. This PR is really three separate changes and only the first one is about master/slave PMT:

1. The election change in start_active_pmts(). This is the one #1445 replaces. Dropping it.
2. ddci_process_pmt() counts a channel again on re-registration. This is a plain counting bug, nothing to do with master or slave PMTs. pmt_add_caid() clears pmt->ca_mask to force a re-send when a new CA descriptor shows up, so send_pmt_to_cas() runs again for a PMT that already has a slot. ddci_process_pmt() finds the existing slot but increments d->channels anyway:

```
   found DDCI 8 for pmt 3, running channels 1
   found DDCI 8 for pmt 3, running channels 2     <- same PMT
   found DDCI 8 for pmt 3, running channels 3     <- same PMT
   ...
   find_ddci_for_pmt: DDCI 8 cannot be used for PMT 57 (used channels 6 max 6), skipping
```

Two real channels used all six slots and the third was refused. The decrement in ddci_del_pmt() has the mirror problem, it fires before checking that the PMT held a slot.

3. pmt->ca_mask carries two different meanings. Also not about master/slave PMTs. The field means both "already sent to this CA" (cleared by pmt_add_caid()) and "registered with this CA, so cleanup is needed on stop" (read by close_pmt_for_ca()). When a CA descriptor update lands in the same burst as a PMT stop, the first meaning overwrites the second, ca_del_pmt is skipped and the CA keeps the PMT registered for ever. This one is in the shared CA path, so it is not only DDCI, dvbapi and hw_descrambler go through the same code.

As far as I can see #1445 does not touch either of those. P5 changes the CAPMT descriptors but not this field, and nothing in the RFC touches the DDCI channel counter. So they will still be there after the refactor.

Would you take 2 and 3 as two small separate PRs, with the logs? They are a few lines each and neither one mentions PMT selection. If you would rather wait until #1445 lands and look at them in that context, that is fine too.

catalinii:
yeah anything not related to master/slave and the proposal in #1445 still needs fixing. Feel free to submit them separately.

Issue #1437, opened 2026-09-14
PID Continuity errors with -9 flag in DDCI


surfcu:
-9 flag solved https://github.com/catalinii/minisatip/issues/1129 issue. Everything works flawlessly in Enigma2, but without MTD so I switched to DDCI.

After DDCI, there are PID Continuity errors with -9 flag only for duplicate vpid/apid channels.  (minisat2.log) [minisat2.log](https://github.com/user-attachments/files/32188239/minisat2.log)
There are no errors for non duplicate vpid/apid  channels. (minisat3.log) [minisat3.log](https://github.com/user-attachments/files/32188276/minisat3.log)

Interestingly, If you first open a non duplicate vpid/apid channel and then simultaneously open a duplicate vpid/apid channel (minisat4.log) [minisat4.log](https://github.com/user-attachments/files/32188327/minisat4.log) there are no errors for any of them.

surfcu:
According to Claude ;

"**Summary**

On a Digital Devices DuoFlex CI (DDCI) setup, enabling --disable-pmt-scan causes intermittent PID Continuity error log lines on the DDCI's decrypted readback adapter for specific channels, even though the underlying vpid/apid are correct and the channel plays fine in Enigma2/SAT>IP mode with the same flag. The errors do not occur when full PMT scanning is enabled, and do not occur on Enigma2 mode regardless of the flag.

This is not reported as a definite minisatip bug — the root trigger appears to be broadcaster-side (a non-compliant DVB-CSA parity switch on certain channels), but --disable-pmt-scan measurably changes how often the DDCI's decrypted-output FIFO gets serviced, which determines whether that broadcaster quirk becomes a visible continuity error. Documenting it here in case others hit the same symptom, and in case a small mitigation on the DDCI side is worth considering.

**Working theory**

ddci_process_ts() (in ddci.cpp) is the function that writes remapped TS into the CI bridge and then drains its decrypted output via push_ts_to_adapter(). It's guarded by a single d->mutex shared across every source adapter feeding that DDCI device, and it only runs when a source adapter's read loop has data ready to process.

With --disable-pmt-scan and only one channel active, minisatip does the minimum background work (only the requested PMT, only one adapter), so the adapter's read/processing loop — and therefore this drain call — runs at its lowest natural cadence.
Without the flag, background PMT scanning across the whole transponder keeps the same adapter's processing loop cycling more often, incidentally increasing the FIFO drain frequency.
[...]

PR #1444, opened 2026-09-18
[DO NOT MERGE] feat(ddci): add --ddci-instrument diagnostics to localize DDCI packet loss (driver vs minisatip)

catalinii:
Follow-up to #1437 / #1440: the fifo fix did not change the PID Continuity error rate on the DDCI readback (`minisatip-fifo-fix.log`: 27 errors in ~4 min, same as before). The remaining question is where exactly the packets are lost: in the kernel driver, in the CAM, or still somewhere in minisatip.

This PR adds an opt-in diagnostic mode that answers that from a single log run.

## Usage

```
./minisatip -9 -c 8:*3 -R ... -l ddci --ddci-instrument -F log.txt
```

Reproduce the failing channel for 4+ minutes, then stop and share the log.

## What it does

TS continuity is tracked per **CI-link pid** (the remapped pids, same semantics as the existing `check_cc`) on **both sides of the bridge**:

| tag | where | meaning |
|---|---|---|
| `[W]` | `ddci_process_ts()`, on the exact iovec passed to `writev()` | what minisatip feeds the CAM |
| `[R]` | `ddci_read_sec_data()`, on the raw bytes read back from the CI device, **before** any minisatip buffering (FIFO) | what the kernel returns to minisatip |

Both sides log lines like:

```
DDCI-INSTR [R] CC error on CI-link pid 4942: expected 5, got 7 (gap 2), 37 ms since previous read (15504 bytes), 8 ms since previous writev
DDCI-INSTR [W] CC error on CI-link pid 4942: expected 5, got 7 (gap 2)
```

Every `[R]` error also carries the cadence context (time since previous read + its size, time since previous `writev`), because the error rate is known to depend on the processing cadence (`-9`, number of active source adapters).

A stats line is logged every 10s:

```
DDCI-INSTATS ddci 8 [10005 ms]: writev 214 (5.12 Mbit/s, max gap 43 ms), reads 812 (5.09 Mbit/s, max gap 8 ms), pushed 6412 KB, drained 6409 KB, fifo push-fails 0, fifo max-level 210 KB, [W] CC errs 0, [R] CC errs 3
DDCI-INSTATS ddci 8 pids (wcnt/rcnt/werr+rerr): 0:w437/r437/e0+0 1:w2297/r2297/e0+0 67:w485/r485/e0+1 4942:w670411/r670395/e0+16 ...
```

## Decision matrix

| observation | conclusion |
|---|---|
| `[W]` errs = 0 **and** `[R]` errs match the final `PID Continuity error` count on the readback adapter | loss is in the **kernel driver / CAM**, before minisatip buffers anything |
| `[W]` errs = 0, `[R]` errs = 0, but final adapter errors > 0 | loss is **inside minisatip** (FIFO/drain path) — `pushed` vs `drained` and `fifo push-fails` will show where |
| `[W]` errs > 0 | minisatip corrupts its own input (mapping/rewrite bug) |
| written vs read packet counts per pid drift apart steadily | quantifies the loss rate inside the kernel path |

The `ms since previous read` / `ms since previous writev` distribution around the `[R]` errors identifies which cadence actually triggers the loss (read servicing vs feed pattern), which is what the -9/adapter-count correlation in #1437 suggests but could not prove.

## Notes

- Zero overhead when the flag is not used: the state is only allocated with `--ddci-instrument` and all hooks check `d->instr` first.
- Pids 0/17/18 (PAT/SDT/EIT regenerated by minisatip) may show `[R]` errors if the CAM regenerates those tables with its own CC; they are tracked per pid so they can be separated from the media pids.
- Also fixes the misleading `dropping 0 bytes` log (it now prints the actual byte count).
- Unit tests added in `tests/test_ddci.cpp` (clean stream, read/write gaps, null pid / no-payload handling).

@surfcu could you run the failing channel with `-l ddci --ddci-instrument` for ~4 minutes and attach the log? The `[W]`/`[R]`/`DDCI-INSTATS` lines will directly show whether the gap is already present in the data returned by the kernel (driver/CAM side) or introduced by minisatip.


surfcu:
[diagnostics-log.txt](https://github.com/user-attachments/files/32391962/diagnostics-log.txt)
Thanks, this is very useful. Ran it on v2.0.105~293d8f2 with:

./minisatip -9 -c 8:*3 -R /usr/src/minisatip/html -l ddci --ddci-instrument -F diagnostics-log.txt

Single channel, beIN Sports 1 (PMT pid 67, video 4942, audio 5042/5142, ECM 1541), DuoFlex CI on adapter 8, tuner on adapter 2. About 170 s of stats, 17 intervals. Full log attached.

Short version: 14 CC errors in the run, all of them on the read side, none on the write side.

1. What we write into the CI is clean

[W] CC errs is 0 in all 17 intervals, and per-pid werr is 0 on every pid for the whole run. [R] CC errs goes 0 -> 14. All 14 PID Continuity error lines are adapter 8.

So the discontinuity appears between our writev() into the CI and our read() back from it. Nothing is wrong with the stream we hand over.

2. The fifo never backed up

Every interval:

fifo push-fails 0, fifo max-level 0 KB

So this is not a buffering or backpressure problem on our side. That also means it is not in the area #1438 / #1440 touched.

3. No packets are actually lost

I checked whether written-minus-read grows over the run. It does not, it just oscillates:

20:54:10  gap  82
20:54:30  gap 787
20:54:50  gap 404
20:55:00  gap 742
20:55:30  gap 220
20:56:20  gap 629

That is just how much is in flight at the moment the stats are printed. If packets were really being dropped this number would only go up. Totals at the end: 576416 written, 575787 read.

4. The interesting part: every error is right after a write

This is the thing I would like your opinion on. All 14 errors say the same thing:

[R] CC error on CI-link pid 4942: expected 7, got A (gap 3), 98 ms since previous read (126336 bytes), 1 ms since previous writev
[R] CC error on CI-link pid 4942: expected B, got E (gap 3), 166 ms since previous read (126336 bytes), 0 ms since previous writev
[R] CC error on CI-link pid 5042: expected A, got B (gap 1), 263 ms since previous read (126336 bytes), 1 ms since previous writev
[R] CC error on CI-link pid 4942: expected F, got 2 (gap 3), 130 ms since previous read (126336 bytes), 1 ms since previous writev

All 14 are in the 0-4 ms bucket for "since previous writev".

writev runs 303 times per 10 s, so one every 33 ms. If the read landed at a random point between two writes, it would fall in the first 4 ms about 12% of the time. Getting 14 out of 14 that way is 0.12^14.

At the same time "ms since previous read" on those same reads is 97-265 ms, which is normal against the ~170 ms average. So the reads that fail are not unusual in their own timing, only in how close they are to a write.

But I cannot prove this from my log. The instrumentation only prints the failing reads, so I have no distribution for the good ones. If the CI produces output as a reaction to being fed, then reads follow writes closely anyway and this correlation means nothing.

Could you make --ddci-instrument also sample "ms since previous writev" for normal reads, not only for errors? Even one line per interval with a small histogram would be enough. If good reads are spread over 0-33 ms and the failing ones sit at 0-4 ms, then it is read/write arbitration on the CXD2099AR and the next place to look is calc_con() and ci_bitrate in the dddvb driver. If both distributions look the same, then I am chasing a ghost and we can drop this idea.

Smaller details
Error positions inside the read buffer are 139 to 628 out of 672 packets. They are spread out, not at the start or the end, so it does not look like a DMA buffer boundary problem.
When two errors land in the same read they are 54 and 74 packets apart, not next to each other. So it is several small gaps of 1-3 packets, not one long burst.
Gap sizes over the whole run: 1, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 3. Never more than 3.
Rate was 14 in 170 s, about 4.9/min. On this hardware I have measured between 3.2 and 13.7/min across earlier runs, so this one is on the low side but normal.
It comes in bursts. 5 errors in one 10 s interval, at the highest bitrate of the run (9.56 Mbit/s), then the last 40 s had none at all.

catalinii:
Analyzed `diagnostics-log.txt` (v2.0.105~293d8f2, 170 s, 17 intervals). Great run — this settles the localization, with one correction and one open discriminator.

## Confirmed

- `[W]` CC errs = 0 in all intervals and per-pid `werr` = 0 → what minisatip feeds the CAM is continuous.
- `[R]` CC errs 0→14 = the 14 `PID Continuity error` lines, all on the readback → **the 28 missing packets (gap sum 1×5, 2×4, 3×5) are dropped between `writev()` and `read()`, in the kernel driver or the CAM.** minisatip buffering is exonerated (`fifo push-fails 0`, `max-level 0 KB`, `pushed == drained` in every interval).

## Correction to point 3 ("no packets are actually lost")

They are permanently lost. There are **zero compensating backward-CC/duplicate events** in ~857 reads — if the packets were merely delayed we would see `got < expected` corrections later in the run. The written−read delta (82…787, end 629) is in-flight buffering in the CI output ring plus our FIFO; a permanent +28 drift cannot be resolved against that ±700 oscillation. The CC accounting is the evidence, not the delta.

## On the 0–4 ms correlation

The 14/14 statistic (p ≈ 0.12¹⁴ if random) is only valid **if reads are write-independent — and they probably are not**:

- Every read in the log is **exactly 126336 B = 672 packets** — reads are "DMA buffer full" events.
- The CAM output follows the input bursts (~112 packets per writev every ~33 ms), so the 672-packet threshold is expected to be crossed right after a write burst — all reads may be write-locked at 0–4 ms **by construction**, which collapses 0.12¹⁴ to ≈ 1.

To settle it, `--ddci-instrument` now also samples `ms since previous writev` for **all** reads and logs a per-interval histogram with the errors attributed per bucket (commit e1bcd7f):

```
DDCI-INSTATS ddci 8 read-phase histogram (ms after writev): 0-4:812 (errs 14), 5-16:0 (errs 0), 17-33:3 (errs 0), 34-66:45 (errs 0), 67-132:0 (errs 0), >132:0 (errs 0) | pkts-since-prev-read min/avg/max: 0/671/672
```

- **Good reads spread over 0–33 ms + errors concentrated at 0–4 ms** → real read/write arbitration race → next stop `calc_con()`/`ci_bitrate`/turnaround handling in dddvb.
- **Good reads also ~all 0–4 ms** → write proximity is trivial, and the primary lead becomes the bitrate dependence already visible in this run: **5.3% of reads failed at 9.46–9.56 Mbit/s vs ~0% at 4.4 Mbit/s**, with read cadence compressing at high rate (max read gap 137 ms vs 230–298 ms). That points at CI-link saturation (arbitration contention scaling with throughput) rather than a phase race.

The `pkts-since-prev-read min/avg/max` figure additionally shows whether reads are output-threshold-driven (≈672 every time) or otherwise.

Could you run the same scenario once more with the updated branch (4+ min, `-l ddci --ddci-instrument`)? One run with the histogram is enough to pick the branch.

surfcu:
[diagnostics-log2.txt](https://github.com/user-attachments/files/32412382/diagnostics-log2.txt)

<html>
<body>
<p dir="ltr" style="">Ran <code>e1bcd7f</code> with the same scenario, about 6 minutes this time (35 stats intervals):</p>

<div role="group" aria-label="Code" tabindex="0" style=""><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">./minisatip -9 -c 8:*3 -R /usr/src/minisatip/html -l ddci --ddci-instrument -F diagnostics-log2.txt</code></pre></div></div>

<p dir="ltr" style="">Same channel as before, be Sports 1 on the DuoFlex CI. Log attached.</p>

<p dir="ltr" style="">Short version: the timing idea I brought up is wrong, and the bitrate dependence does not show up
either. The part that still stands is that the loss is between <code>writev()</code> and <code>read()</code>.</p>

<h3 dir="ltr" style="">1. Read timing: good reads are not spread over 0-33 ms</h3>

<p dir="ltr" style="">Summed over all 35 intervals, 2147 reads:</p>

ms after writev | reads | share | errs | errs per read
-- | -- | -- | -- | --
0-4 | 2035 | 94.8% | 24 | 1.18%
5-16 | 112 | 5.2% | 2 | 1.79%
17-33 | 0 | 0% | 0 | -
34-66 | 0 | 0% | 0 | -
67-132 | 0 | 0% | 0 | -
>132 | 0 | 0% | 0 | -

<p dir="ltr" style="">Almost every read happens within 4 ms of a write, and not a single one happened later than 16 ms.
So the reads just follow the writes, probably because the CI only gives data back after it has been
fed. The errors were at 0-4 ms only because nearly everything is at 0-4 ms. The error rate per read
is about the same in both buckets.</p>

<p dir="ltr" style="">So the "good reads spread over 0-33 ms" case does not apply and this is not evidence of a
read/write arbitration race. That was my idea and it did not hold up, sorry for the detour. Thanks
for adding the control, it answered the question clearly.</p>

<h3 dir="ltr" style="">2. Bitrate: no dependence that I can see</h3>

<p dir="ltr" style="">The 5.3% at 9.46-9.56 Mbit/s came from one interval in the first run (5 errors in 95 reads at
9.56). The interval right before it, at 9.46, was 1 in 94. And one of the 4.4 Mbit/s intervals had
1 in 44, so not really ~0%:</p>

<div role="group" aria-label="Code" tabindex="0" style=""><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">20:54:20  4.35 Mbit/s  reads 43  errs 0  -&gt;  0.0%
20:54:40  9.46 Mbit/s  reads 94  errs 1  -&gt;  1.1%
20:54:50  9.56 Mbit/s  reads 95  errs 5  -&gt;  5.3%
20:55:20  4.41 Mbit/s  reads 44  errs 1  -&gt;  2.3%
20:55:40  4.41 Mbit/s  reads 43  errs 0  -&gt;  0.0%</code></pre></div></div>

<p dir="ltr" style="">This run went from 4.27 to 10.74 Mbit/s, so I pooled both runs, 52 intervals, 3150 reads, 40
errors:</p>

<div role="group" aria-label="Code" tabindex="0" style=""><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">below 5 Mbit/s:     8 / 583 = 1.37%   (13 intervals)
8 Mbit/s and above: 7 / 465 = 1.51%   (5 intervals)
correlation between bitrate and errors per read: r = 0.02</code></pre></div></div>

<p dir="ltr" style="">The worst interval of this run per read was actually at 4.27 Mbit/s, the lowest bitrate of the run.</p>

<h3 dir="ltr" style="">3. What still holds</h3>

<p dir="ltr" style="">Same as the first run:</p>

<div role="group" aria-label="Code" tabindex="0" style=""><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">[W] CC errs 0, [R] CC errs 26
fifo push-fails 0, fifo max-level 0 KB   (all 35 intervals)</code></pre></div></div>

<p dir="ltr" style="">So by your table the loss is in the kernel driver or the CAM, and minisatip is out of it. 26
errors in about 350 s, 4.5/min. Two more errors landed after the last stats line, so the log has 28
on adapter 8 in total. The one error on adapter 2 is pid 18 at position 0, 33 ms after the stream
started, the usual start-up one on a table pid.</p>

<p dir="ltr" style="">About 1.2-1.8% of reads contain a CC gap, and with these counters it looks the same regardless of
timing or bitrate.</p>

<h3 dir="ltr" style="">4. A correction</h3>

<p dir="ltr" style="">In my first comment I wrote "no packets are actually lost". That was too strong. What I meant is
that written-minus-read does not keep growing, so nothing is lost in bulk. But the CC gaps are real
missing packets, and your count of 28 in that run is right. There are just too few of them to see
against the few hundred packets that are in flight at any moment.</p>
</body>
</html>

catalinii:
@surfcu I suggest if you have an agent to try to figure out what happens in the driver:
1) instrument the driver
2) recompile
3) reload the driver
4) start minisatip
5) play analyze results
6) stop minisatip
7) unload driver

Some other people may have reported similar issues IIRC.

PR #1446, opened 2026-09-20
fix(satipc): separate addpids and delpids with '&' in upstream PLAY requests

catalinii:
When minisatip acts as a SAT>IP client and a single commit has both PIDs to add and PIDs to delete, satipc_get_pids() concatenated the two query parameters without a separator:

    PLAY rtsp://server/stream=1?addpids=202delpids=201,213

The SAT>IP server splits the query string on '&', so it sees a single addpids parameter with the garbage value "202delpids=201,213": the leading numbers get added while the delpids half is never parsed and those PIDs are never removed. Since minisatip believes the deletes were sent, the stale PIDs linger on the server for the rest of the session.

Symptom observed in production: after the client finished a service scan (stepping through PMT PIDs 201-231) and settled on PIDs 0,18,221,2211,2212, the upstream server still held 20 PIDs including the previous channel's ES PIDs (2130,2132), i.e. two channels' worth of PIDs on a single stream.

Emit '&' before delpids= whenever an addpids= parameter was already written. Add-only and del-only requests are unchanged.

Add test_satipc_get_pids_add_and_del regression test covering the combined, add-only and del-only cases.


lars18th:
> When minisatip acts as a SAT>IP client and a single commit has both PIDs to add and PIDs to delete, satipc_get_pids() concatenated the two query parameters without a separator:

🤭 Serious bug for a long time!

Thank you @catalinii for continuing to maintain this great project.👍

PR #1449, opened 2026-09-22
fix(adapter): floor max_pids at 16 and report pid-add failures

catalinii:
A demux failure (e.g. DMX_ADD_PID errno 22) permanently lowered the per-adapter max_pids cap, eventually toward 0, while the RTSP PLAY reply still returned 200 OK even though no data (not even PAT) could flow.

- Keep max_pids at MIN_ADAPTER_PIDS (16) or above.
- update_pids returns an error when a pid fails to be added while below that count, so tune() fails and the RTSP reply reflects the error status.
- Fix the adjacent log line to print ad->max_pids instead of opts.max_pids.
- Add test_update_pids_max_pids_floor regression test.
- Reset the state on adapter open

Verified: test_adapter passes, full ctest 13/13 green.


Jalle19:
Why does "DMX_ADD_PID errno 22" happen? Is this something that only happens on some hardware and not others?

catalinii:
DMX_ADD_PID before DMX_SET_PES_FILTER leads to -22.

src/dvb.cpp
```diff
@@ -904,6 +914,9 @@ int dvb_open_device(adapter *ad) {
     }
     ad->type = ADAPTER_DVB;
     ad->dmx = -1;
+    // Fresh demux fd carries no PES filter: drop any counter left over from
+    // a previous session, so the first demux pid uses DMX_SET_PES_FILTER.
+    ad->active_demux_pids = 0;
```
Jalle19, line 919: So is this actually the only fix needed?

catalinii: no, this resets the counter in case of leak, also the leak itself was fixed in https://github.com/catalinii/minisatip/pull/1450

The other change is returning failure if the filter for the pid failed to be added. On failure to delete, we do not report failure, because there is no point (as most likely the stream is shut down).

